PR #1, opened 2016-04-19
Bunch of small fixes

mithro:
The first two changes are from the timvideos fork of misoc at https://github.com/timvideos/misoc/commits/legacy

I'm guessing these changes should also be sent to upstream MiSoC too?


enjoy-digital:
Merged thanks. Yes we can experiment things here and merge that later in MiSoC.

PR #2, opened 2016-04-19
More fixes.

mithro:
* Make it easier to understand the cause when verilator build fails.
 * Make it easier to see what arch the BIOS is running on.


enjoy-digital:
Thanks, merged. I just changed where the cpu information was printed and put it on the first line.

PR #7, opened 2016-10-27
soc/software/Makefile: Fix Makefile depend generation.

mithro:
Previously the flags were not actually set and the *.d files were never
actually generated.


enjoy-digital:
Thanks.

PR #9, opened 2016-10-30
libbase: Adding missing vprintf function.

### LINKED ISSUE #8: libbase in litex is missing vprintf
vprintf is needed to be able to wrap printf calls.

mithro:
Fixes #8.

```
int vprintf(const char *format, va_list ap);

The functions vprintf(), vfprintf(), vsprintf(), vsnprintf() are equivalent to
the functions printf(), fprintf(), sprintf(), snprintf(), respectively, except
that  they  are  called  with  a  va_list instead of a variable number of
arguments.
```


enjoy-digital:
Thanks.

mithro:
We should send this upstream to misoc

PR #10, opened 2016-12-17
Patches to make it easier to work with Etherbone

mithro:
* Provide the CSR data bus width in a constant.
 * Allow the Etherbone objects to autocompleted in things like `IPython` and `bpython`.


enjoy-digital:
thanks.

PR #11, opened 2016-12-17
Only require rebuild on actual changes

mithro:
* Files are only modified when their contents change.
 * Fix Makefile so bios is correctly rebuilt when dependencies change (but isn't otherwise).


enjoy-digital:
thanks.

PR #12, opened 2016-12-19
Fix Makefile dependency inclusion for other software.

mithro:
I previously fixed it for the bios, but not these libraries.


enjoy-digital:
thanks.

PR #16, opened 2017-01-13
Fixing accidental revert in merge commit.

mithro:
You accidentally reverted this feature in the merge which meant the firmware was completely rebuilding every time (as csr.h was modified).


enjoy-digital:
thanks.

PR #17, opened 2017-01-14
Fixing missing csr_constant/config support.

mithro:
Missed as part of misoc merge at ff31959aea1e3815d95798996c3999f122314ae6.


enjoy-digital:
I didn't merged that because I think it requires others changes. Will the other way we define constant will still work with this? If so, OK to merge. If not then we need to modify everything at the same time.

mithro:
All existing constants seem to work. This just enables us to use CSRConstants too....

enjoy-digital:
ok thanks.

PR #20, opened 2017-02-21
Add Mercury development board (port from MiSoC)

cr1901:
I've been sitting on this commit for quite a while. I've simply been forgetting to send it! Self-explanatory commit message.


enjoy-digital:
thanks.

PR #21, opened 2017-03-05
Allow using gcc for or1k.

mithro:
* Using CLANG can set by using CLANG=1 or CLANG=0 in the environment.
 * or1k continues to default to CLANG if environment is not net.

This change was clobbered by your merge in ff31959aea1e3815d95798996c3999f122314ae6


enjoy-digital:
merged, thanks.

PR #22, opened 2017-03-12
soc_core: Add CPU_RESET_ADDR as a constant.

mithro:
So we can do a "soft reset" by jumping to this address.

Nothing currently provides this information. Discovered this when trying to use `reboot` on the MimasV2 which has a non-zero CPU_RESET_ADDR.


enjoy-digital:
merged, thanks.

PR #25, opened 2017-07-05
Add Versa ECP5-5G Platform.

q3k:
This PR implements support for the [Lattice ECP5-5G Versa Development Kit](http://www.latticesemi.com/Products/DevelopmentBoardsAndKits/ECP55GVersaDevKit.aspx) platform.

I've not tested the code (especially pin mapping) apart from blinking one of the user_leds in a test module and running that on the board via .create_programmer(). DDR support and a proper SoC target might come in a future PR.

### DIFF
--- a/litex/boards/platforms/versaecp55g.py
+++ b/litex/boards/platforms/versaecp55g.py
@@ -1,0 +1,138 @@
+# This file is Copyright (c) 2017 Serge 'q3k' Bazanski <serge@bazanski.pl>
+# License: BSD
+
+from litex.build.generic_platform import *
+from litex.build.lattice import LatticePlatform
+from litex.build.lattice.programmer import LatticeProgrammer
+
+
+_io = [
+    ("clk100", 0, Pins("P3"), IOStandard("LVDS")),
+    ("rst_n", 0, Pins("T1"), IOStandard("LVCMOS33")),
+
+    ("user_led", 0, Pins("E16"), IOStandard("LVCMOS25")),
+    ("user_led", 1, Pins("D17"), IOStandard("LVCMOS25")),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, merged! (timing is fine, just received one a week ago :) )

mithro:
This is super cool! I'm always interested in seeing more non-Xilinx FPGA options :-)

@enjoy-digital Since you have one of these boards now what are your plans around it? @q3k - I would love to know your plans too!

Do you think you'll add support to LiteDRAM for the ECP5 and any idea of the possible performance? The above page claims "DDR3-1866" which has a clock speed of 933.33MHz -- that seems much faster than the Artix-7 / Spartan-6 speeds but should be pretty comparable to the Kintex-7 or Ultrascale stuff?  It also then later goes on to say "64Mb/x16, 667 MHz" so maybe that is only how fast it can go?

I'm also assuming that LiteEth support will be pretty trivial?

What about LitePCI?

Lattice has some kits like the [Embedded Vision Development Kit
](http://www.latticesemi.com/en/Products/DevelopmentBoardsAndKits/EmbeddedVisionDevelopmentKit.aspx) & the [Lattice USB3 Video Bridge Development Kit](http://www.latticesemi.com/en/Products/DevelopmentBoardsAndKits/LatticeUSB3VideoBridgeDevelopmentKit.aspx) (which is actually an ECP3).

Sadly, last time I checked you can't do HDMI 1080p60 into the IO pins of the ECP5/ECP3 boards, they only go up to ~1Gbit/s LVDS.

enjoy-digital:
@mithro: I need a CPRI core for a customer project and wants to use ECP5 + CPRI core from Lattice (99$ each) as a low cost reference.

PR #26, opened 2017-07-20
Add Diamond toolchain support for Linux.

q3k:
This tries to replicate the same setup as in the Windows buildsystem. We
also remove the Jedecgen step, as it doesn't seem to be supported nor
necessary in newer versions of Diamond.


enjoy-digital:
Thanks.

PR #29, opened 2017-09-26
Re-enable SPI flash bitbanging

cr1901:
This pull request reimplements the CSR interface for writing to SPI flash from a CPU using bitbanging/CSR egister writes. Additionally, _reading_ spiflash via bitbanging is also functional, and I added a `flash_read_byte` function to libbase for convenience. The register interface is identical regardless of whether a board has a naive SPI flash or supports dual/quad mode.

Reading/writing via the bitbang interface has been confirmed to work with both single and quad SPI flash memories.

### DIFF
--- a/litex/soc/cores/spi_flash.py
+++ b/litex/soc/cores/spi_flash.py
@@ -26,7 +26,7 @@ def _format_cmd(cmd, spi_width):
 
 
 class SpiFlashDualQuad(Module, AutoCSR):
-    def __init__(self, pads, dummy=15, div=2):
+    def __init__(self, pads, dummy=15, div=2, with_bitbang=True):
         """
         Simple SPI flash.
         Supports multi-bit pseudo-parallel reads (aka Dual or Quad I/O Fast
@@ -35,6 +35,11 @@ class SpiFlashDualQuad(Module, AutoCSR):
         self.bus = bus = wishbone.Interface()
         spi_width = len(pads.dq)
         assert spi_width >= 2
+
+        if with_bitbang:
+            self.bitbang = CSRStorage(4)
+            self.miso = CSRStatus()
+            self.bitbang_en = CSRStorage()
 
         # # #
 
@@ -47,7 +52,6 @@ class SpiFlashDualQuad(Module, AutoCSR):
         read_cmd_params = {
             4: (_format_cmd(_QIOFR, 4), 4*8),
             2: (_format_cmd(_DIOFR, 2), 2*8),
-            1: (_format_cmd(_FAST_READ, 1), 1*8)
         }
         read_cmd, cmd_width = read_cmd_params[spi_width]
         addr_width = 24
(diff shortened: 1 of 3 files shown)


enjoy-digital:
I removed it because code was already complicated to maintain and a better (and faster way) to access SPI flash for write would be to just reuse SPI module (https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/spi.py) and multiplex the pads in your top module.
What do you think about that?

cr1901:
>What do you think about that?

This isn't possible currently:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/spi.py#L326

I would have to multiplex tristates internally, which only exist on I/O pads of modern FPGAs.

enjoy-digital:
We don't need tristate for the case `len(pads.cs_n) == 1`.
Would you be ok to keep this code in hdmi2usb-litex-firmware repository until i have a closer look. (put your modified spi_flash.py in gateware).

enjoy-digital:
This hardware is now integrated in hdmi2usb-litex-firmware:
https://github.com/timvideos/HDMI2USB-litex-firmware/commit/8f52555e6c06a5a5861091636dd042851b3a56a6

You just need to integrate the software.

I'll test spi hardware read/write with pads muxing soon and you will be able to use it if wanted.

cr1901:
>You just need to integrate the software.
>I'll test spi hardware read/write with pads muxing soon and you will be able to use it if wanted.

That's fair; for now HDMI2USB still has code/gateware that is "known to work" and we can eventually upgrade to a cleaner solution.

mithro:
I think a cleaner solution would be the following;

 * Have a "mux" module which lets you select between different "spi controllers".
 * Have the spiflash memory object be one such "spi controller" module.
 * Have the bitbanging stuff be another such "spi controller" module.
 * Have Florent's other spi stuff also be such a "spi controller" module.
 * Have a module which has the spiflash memory + bitbanging mode + mux inside it.

Kinda like this -> 
image

https://docs.google.com/drawings/d/13vIy5G7S6xrlvVx2DAzWNhmfWYhZzoDXfxlM-_9nXi8/edit

cr1901:
> * Have Florent's other spi stuff also be such a "spi controller" module.

I don't think this is appropriate for the other spi stuff, which right now is only a `SPIMaster`. `spiflash` is probably one of the only SPI peripherals where it's desirable to access the device in two different methods (abstract SPI bus away, or interact w/ SPI bus directly).

enjoy-digital:
@mithro, @cr1901: we can create a SPIMux module that takes pads as parameter and has a get_pads() method. Others controllers will use this method to get their pads and inside SPIMux we will have a CSR to select SPI controller according to the number of pads requests we had.

mithro:
Only thing which might be a little dicy is that you don't want to switch the mux in the middle of the spiflash module doing things...

xobs:
Are there any recommendations on how to accomplish this?  My current approach has been to forego the Litex SPI modules in favor of the PicoRV32 SPI controller, but I gather that's not the best approach.

enjoy-digital:
I re-integrated recently the bit-bang write support, maybe we don't need to do the multiplexing i was suggesting in https://github.com/enjoy-digital/litex/pull/29#issuecomment-332158629 (i would need to evaluate the real speed up using a hardware SPI core for writes).

So if you want to have memory-mapped reads and bitbanging write, you can use SpiFlashDualQuad or SpiFlashSingle (depending if you flash support Single/Dual/Quad SPI reads) and you can this to see how to do write accesses:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/main.c#L312
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/spiflash.c

PR #34, opened 2017-10-29
Change the default IRQs.

mithro:
* Reserve IRQ 0 to be used as a "non-maskable interrupt" (NMI) in the future.

 * Make it harder to define duplicate IRQs and provide better error messages when it happens.

 * Use IRQ 2 for the LiteX. This matches the standard mor1k config which connects the UART to IRQ 2.

This change is needed for Linux running on LiteX as it gets grumpy with using IRQ 0 for anything other other than an NMI.


mithro:
@stffrdhrn - Can you give this change a test?

mithro:
Adding a bunch of fixes.

 * Put uart back to 'uart' rather than 'uart0'.
 * Made it harder to accidently define duplicate IRQs.

mithro:
And a datapoint why something like this is important - I just discovered we are using the same IRQ value for LiteEth and HDMI0 in the HDMI2USB-litex-firmware (before we changing the value). That now provides the helpful error message;
`AssertionError: Interrupt vector conflict for IRQ 3, user defined hdmi_in0 conflicts with user defined ethmac`

mithro:
Merging!

PR #40, opened 2017-12-30
cpu: Adding "variant" support.

mithro:
It is useful to support slightly different variants of the CPU
configurations. This adds a "cpu_variant" option.

For the mor1k we now have the default mor1k configuration and the
"linux" variant which enables the features needed for Linux support on
the mor1k.

Currently there are no variants for the lm32, but we will likely add a
"tiny" variant for usage on the iCE40.


enjoy-digital:
Thanks, merged.

PR #41, opened 2017-12-30
fhdl/tracer: Import Python 3.5/3.6 version guards from Migen.

cr1901:
Self-explanatory, this commit ensures Python 3.6 works. I have not tested this, but it's a direct `cp` from my Migen and thus should work fine.  I don't know if/when Migen/MiSoC and LiteX will reunite, but I'll keep making PRs to LiteX for the important contributions.

Minor contributions of mine can be imported whenever you do a bulk merge.

### DIFF
--- a/litex/gen/fhdl/tracer.py
+++ b/litex/gen/fhdl/tracer.py
@@ -1,15 +1,40 @@
 import inspect
+from sys import version_info
 from opcode import opname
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, merged. I'm going to do some tests.

PR #43, opened 2018-01-11
fix the unsupported programmer case for kc705 and minispartan6

felixheld:


litex/boards/platforms/kc705.py
```diff
@@ -496,7 +496,8 @@ def create_programmer(self):
         elif self.programmer == "impact":
             return iMPACT()
         else:
-            raise ValueError("{} programmer is not supported".format(programmer))
+            raise ValueError("{} programmer is not supported"
+                             .format(self.programmer))
```
mithro: The normal way to wrap this would be;
```python
    raise ValueError("{} programmer is not supported".format(
        self.programmer))
```
or
```python
    raise ValueError(
        "{} programmer is not supported".format(self.programmer))
```

enjoy-digital:
merged, thanks.

PR #50, opened 2018-01-14
Import migen+misoc changes

mithro:
DO NOT MERGE YET!

This is an attempt to reduce the changes between LiteX and Migen/MiSoC.

There are a bunch of commits in this that need careful checking and rework.

This change also needs to be rebased onto the latest LiteX changes...


mithro:
I'll test this works with the HDMI2USB-litex-firmware now....

litex/boards/targets/simple.py
```diff
@@ -47,12 +47,14 @@ def __init__(self, platform, **kwargs):
         self.submodules.ethmac = LiteEthMAC(phy=self.ethphy, dw=32,
                                             interface="wishbone",
                                             with_preamble_crc=False)
-        self.add_wb_slave(mem_decoder(self.mem_map["ethmac"]), self.ethmac.bus)
+        self.add_wb_slave(self.mem_map["ethmac"], 0x2000, self.ethmac.bus)
```
enjoy-digital: Not sure we can take that without others misoc changes.

litex/boards/targets/simple.py
```diff
@@ -47,12 +47,14 @@ def __init__(self, platform, **kwargs):
...
         self.add_memory_region("ethmac", self.mem_map["ethmac"] | self.shadow_base, 0x2000)
+        self.csr_devices += ["ethphy", "ethmac"]
```
enjoy-digital: Not sure we can take that without others misoc changes.

litex/boards/targets/simple.py
```diff
@@ -47,12 +47,14 @@ def __init__(self, platform, **kwargs):
...
+        self.interrupt_devices.append("ethmac")
```
enjoy-digital: Not sure we can take that without others misoc changes.

litex/build/xilinx/ise.py
```diff
@@ -98,22 +98,34 @@ def _run_ise(build_name, ise_path, source, mode, ngdbuild_opt,
     if source:
         settings = common.settings(ise_path, "ISE_DS", ver, first="version")
         build_script_contents += source_cmd + settings + "\n"
-
-    ext = "ngc"
-    build_script_contents += """
+    if mode == "edif":
+        ext = "edif"
```
enjoy-digital: If we reintegrate edif, i think others files will be needed (fhdl/edif.py?)

litex/build/xilinx/ise.py
```diff
@@ -165,31 +178,46 @@ def build(self, platform, fragment, build_dir="build", build_name="top",
...
+                if mode in ("xst", "cpld"):
                     _build_xst_files(platform.device, sources, platform.verilog_include_paths, build_name, self.xst_opt)
-                    isemode = "xst"
+                    isemode = mode
                 else:
                     _run_yosys(platform.device, sources, platform.verilog_include_paths, build_name)
                     isemode = "edif"
                     ngdbuild_opt += "-p " + platform.device
 
+            if mode == "mist":
```
enjoy-digital: Let's remove mist support from migen

litex/build/lattice/diamond.py
```diff
@@ -57,6 +57,7 @@ def _build_files(device, sources, vincpaths, build_name):
     tcl.append("prj_run Map -impl implementation")
     tcl.append("prj_run PAR -impl implementation")
     tcl.append("prj_run Export -impl implementation -task Bitgen")
+    tcl.append("prj_run Export -impl implementation -task Jedecgen")
```
q3k: This does not work for me (Diamond 3.9 on Linux x86_64):

    Wrong # arg -task
    Jedecgen is not a valid task name.
        while executing
    "prj_run Export -impl implementation -task Jedecgen"
        (file "top.tcl" line 10)

Do we actually need to generate JEDEC files? I would think these don't even apply to modern FPGAs anymore... Any particular reason you want this present?

enjoy-digital:
Thanks. Situation is more clear now: LiteX is an alternative to MiSoC and reuse Migen. I'm closing this since i'll import the MiSoC features that can be interesting manually since this also need testing.

PR #51, opened 2018-01-16
Include the ethernet related header files conditionally

felixheld:
Only including those header files in the litex firmware is the first step to
move the firmware parts of liteeth to the liteeth tree.

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -9,8 +9,11 @@
 #include <generated/mem.h>
 #include <generated/csr.h>
 
+#ifdef CSR_ETHMAC_BASE
 #include <net/microudp.h>
 #include <net/tftp.h>
+#endif
+
 #include "sfl.h"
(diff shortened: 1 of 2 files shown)


mithro:
I strongly support the direction that this is going but it is going to need @enjoy-digital's review to check he is also happy.

felixheld:
ACK.

My goal with this is to have the firmware to control a core inside the repository of that core and for that litex needs to build without the firmware for a certain core when that core isn't used in a system.

felixheld:
One of the follow-up commits should also change the condition when the ethernet-related part of the code is compiled and not ifdef'd out. Currently the code is only included when the CSR base of the ethernet mac registers exists with a certain name. Ideally the code will not match to a name, but to the existence of at least one instance of a certain type of core in the system, but currently I'm not sure how to implement and integrate that properly.

enjoy-digital:
Thanks, I'm ok going this direction.

PR #55, opened 2018-01-19
BIOS: Flashboot without main ram

ewen-naos-nz:
Modified flashboot() to skip copy to main ram if there is no main ram, and instead execute in place out of SPI flash.  Needed to support, eg, Numato Elbert v2 which has no external SDRAM (only FPGA internal SRAM).  Untested on Numato Elbert v2 (still working on getting gateware building for Spartan 3A), but compiles for Numato Mimas v2 / Digilent Arty A7; both will still serialboot, and Mimas v2 also appears to still flashboot.

PR #57, opened 2018-01-20
WIP - BIOS: Flashboot without main ram

rohitk-singh:
Modified flashboot() to skip copy to main ram if there is no main
ram, and instead execute in place out of SPI flash.  (For this to
work the linker .ld will also need to redirect references to be
inside the SPI flash mapping.)

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -290,6 +290,17 @@ void netboot(void)
 #endif
 
 #ifdef FLASH_BOOT_ADDRESS
+
+/* On systems with exernal SDRAM we copy out of the SPI flash into the SDRAM
+   before running, as it is faster.  If we have no SDRAM then we have to
+   execute directly out of the SPI flash. */
+#ifdef MAIN_RAM_BASE
+#define FIRMWARE_BASE_ADDRESS MAIN_RAM_BASE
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Isn't it a duplicate of #55?

mithro:
@enjoy-digital - @rohitk-singh Is testing this change on the cmod-a7

enjoy-digital:
ok, we'll merge it once validated.

enjoy-digital:
@rohitk-singh: Was it validated?

rohitk-singh:
@enjoy-digital No. We ran issues with Cmod A7 and were not able to program it. Most likely issues with Openocd and the 12MHz clock sourced from FTDI, which openocd controls. I'll test it with MimasV2 and then confirm. Please do not merge yet.

enjoy-digital:
@rohitk-singh: any update?

rohitk-singh:
@enjoy-digital Sorry for delay. I just tested it. Works fine. Designer just has to make **sure** to define 'flash_boot_address` in the Python SoC target class with the correct memory-mapped location of the firmware to boot in SPI Flash. Other wise BIOS will not support flash boot.

enjoy-digital:
thanks! merged.

PR #59, opened 2018-01-23
Allow for multiple synthesis directives in specials.

q3k:
This is needed to specify timing constraints on some Lattice Diamond library specials, like the EHXPLLL.

To keep backwards compatibility we allow the directive to still be a single string. If it's not, we assume it's an iterable.


enjoy-digital:
Thanks, that seems fine.

PR #60, opened 2018-01-23
Top module selection (for Verilator and Diamond)

q3k:
When dealing with multiple Verilog sources in a design we must explicitly state which module is the top of the design. This change ensures such a selection is made when building designs using Verilator and Lattice Diamond.

While the changes have been tested to work, I am not sure if this is the right way to go about this. Please review carefully.


mithro:
@enjoy-digital - FYI

enjoy-digital:
@q3k, @mithro: sorry, i'll look at that soon.

enjoy-digital:
thanks, merged.

PR #61, opened 2018-01-23
platform/arty.py: Move Pmod definitions to 'connectors' section.

PaulSchulz:
pmod definitions should be in 'connectors' section of platform definition for the ARTY board.


enjoy-digital:
Thanks, merged.

PR #63, opened 2018-01-26
boards/platforms: Add Arty S7 Board.

cr1901:
Verified with the `simple` SoC target. I'll add the DDR3 controller in a few days.


enjoy-digital:
Thanks! Merged.

PR #65, opened 2018-02-27
platforms/tinyfpga_b: Move serial peripheral out of default I/O, make it

cr1901:
Self-explanatory and bikeshedding.

As has been done in the past w/ Migen, users should _opt into_ using connectors/not-dedicated pins as `_io` using `add_extension`, rather than having non-dedicated pins assigned static functions by default. 

I prefer to add default pin assignments in the board namespace that users can assign to the board via `add_extension`. This permits maximum flexibility for users who need the I/O pins for something else, as well as providing default pins for those who want to use the first two I/O pins as a serial port.

### DIFF
--- a/litex/boards/platforms/tinyfpga_b.py
+++ b/litex/boards/platforms/tinyfpga_b.py
@@ -9,12 +9,6 @@ _io = [
         IOStandard("LVCMOS33")
     ),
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks.

PR #67, opened 2018-03-03
xilinx/vivado: Provide a fallback mechanism for using the same root f…

cr1901:
…or Vivado and ISE toolchains.

`HDMI2USB`, `litex-buildenv`, and presumably others code repositories using `litex` expect to search for toolchains using a single global "Xilinx" root directory. The backends will use this root directory to search for the actual ISE and Vivado toolchains. This is because Vivado and ISE have different directory hierarchies, and we don't know ahead of time which toolchain we need to use.

Recently, `litex` changed the backend behavior to match `migen`. When searching for toolchains, `migen` will only search for [matching version numbers](https://github.com/m-labs/migen/blob/master/migen/build/tools.py#L36) under [the passed-in toolchain path](https://github.com/m-labs/migen/blob/master/migen/build/xilinx/common.py#L35-L39). 

Although Vivado and ISE _do_ typically share the same root directory, ISE and Vivado have different directory structures (ISE does `$VERSION/ISE_DS`, Vivado does `$ROOT/Vivado/$VERSION`). `Litex` [previously](https://github.com/enjoy-digital/litex/commit/64e4e1ce84daad98d6120f075e27f4a8a5060271#diff-1d1ee9e0156af00e70c0d66869d2ba94L38) had logic such that the same `toolchain_path` could be passed into the `build` method for both the Vivado and ISE backends.

ISE platforms work fine as-is with the migen behavior, even passing in the root `toolchain_path` (typically `/opt/Xilinx`). However, using the same root `toolchain_path` broke Vivado platform with an `OSError: no version directory for Xilinx tools found in...` etc. This patch adds a fallback to search for a `Vivado` directory underneath the passed-in `toolchain_path` to search for matching version numbers, fixing code depending on the old behavior.


enjoy-digital:
thanks.

PR #68, opened 2018-03-04
Improving error message when csr name is not found.

mithro:
Before;
```
"/usr/local/lib/python3.5/dist-packages/litex-0.1-py3.5.egg/litex/soc/integration/soc_core.py",
line 258, in get_csr_dev_address
return self.csr_map[name]
KeyError: 'core'
```

Now;
```
  File "XXXX/github/enjoy-digital/litex/litex/soc/integration/soc_core.py", line 259, in get_csr_dev_address
    return self.csr_map[name]
KeyError: 'ddrphy'

The above exception was the direct cause of the following exception:

  ...
  File "XXXX/github/enjoy-digital/litex/litex/soc/interconnect/csr_bus.py", line 199, in scan
    mapaddr = self.address_map(name, None)
  File "XXXX/github/enjoy-digital/litex/litex/soc/integration/soc_core.py", line 269, in get_csr_dev_address
    ) from e
RuntimeError: Unable to find ddrphy in your SoC's csr address map.

Check BaseSoC.csr_map in XXXX/github/enjoy-digital/litex/litex/boards/targets/arty.py

Found l2_cache, timer0, ddrphy2, buttons, sdram, identifier_mem, uart, uart_phy, leds, crg in the csr_map
```

### DIFF
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -1,3 +1,4 @@
+import inspect
 from operator import itemgetter
 
 from migen import *
@@ -256,6 +257,16 @@ class SoCCore(Module):
             name = name + "_" + memory.name_override
         try:
             return self.csr_map[name]
+        except KeyError as e:
+            raise RuntimeError("""\
+Unable to find {} in your SoC's csr address map.
+
+Check {}.csr_map in {}
+
+Found {} in the csr_map""".format(
+                name, self.__class__.__name__, inspect.getfile(self.__class__),
+                ", ".join(self.csr_map.keys()))
+            ) from e
         except ValueError:
             return None
 


enjoy-digital:
Do we really need to display the csr that was found? I'm fine with a better error message and with indicating which file to look at but i don't see the additionnal value of showing the csr that were found.

mithro:
@enjoy-digital 

The way I think about it is, 
> What is the first question I'm going to ask when trying to fix this error?

In this case the first question I'm going to ask is 
> Well what csr names *are* actually in the map?

and so that is why I included them in the error message. 

What is your worry about including them? Is it that the error message output will get too long? This list has been pretty sort on every design I've seen?

The two most common causes of this issue are;
 * Forgetting to add the name to the csr map
 * Misspelling the csr name.

Another further enhancement would be look at the edit distance between the requested CSR and the CSRs that are available. If there are any that are *really* close then it is likely someone just misspelled the name and we can provide that suggestion. clang & gcc now do this in their error messages and it is super handy
> Did not find `hello`, did you mean `_hello`?

enjoy-digital:
@mithro: i generally answer to the question "Well what csr names are actually in the map?" by looking at the target file, but yes for things with multiple level of inheritance it can be more difficult.
Merged, thanks.

PR #69, opened 2018-03-04
Adding conda environment and simple travis build

mithro:
This CL adds a simple "conda environment" example. It then tests this on Travis. This checks that all the targets at least compile. You can see [example output here](https://travis-ci.org/mithro/litex/builds/348804944) or below;
image


enjoy-digital:
Thanks, merged! I should try to make more use of Conda and Travis...

PR #70, opened 2018-03-04
Convert README to markdown for nicer rendering on GitHub

mithro:
This makes the rendering nicer and has the following advantages;
 * Links are now clickable.
 * You can link directly to a heading like this link to [Very Quick Start Guide for Newcomers](https://github.com/mithro/litex/tree/readme-md#very-quick-start-guide-for-newcomers).
 * Syntax highlighting of code snippets and such.

Before;
image

After;
image


enjoy-digital:
@mithro, thanks, but the choice of using plain text for the README is deliberate and i'd like to keep that for now. (even if there is no advantages...). I'm however open to use markdown for documentation or others things.

PR #71, opened 2018-05-24
Fix for missing connectors for arty boards

DeanoC:
The arty platform connectors weren't hooked up to the SOC, also they were in an incompatible syntax.
I've hooked them up, but lost the IOStandard field, as AFAICT thats not supported for connectors at this time.

### DIFF
--- a/litex/boards/platforms/arty.py
+++ b/litex/boards/platforms/arty.py
@@ -109,10 +109,10 @@ _io = [
 ]
 
 _connectors = [
-    ("pmoda", 0, Pins("G13 B11 A11 D12 D13 B18 A18 K16"), IOStandard("LVCMOS33")),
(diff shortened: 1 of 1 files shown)


mithro:
@cr1901 - PTAL?

enjoy-digital:
Thanks! merged.

cr1901:
@mithro Still sick, but will make a corresponding change in `arty_s7` soon. Seems what happened was initially `arty` platform defined pmod as `io` rather than a `connector`.

Then someone [refactored](https://github.com/DeanoC/litex/commit/3ac28ed6f7d4a980ca6f5803f61c0bfd6cc48402#diff-5bc78ceadf12ce9f8b5ce6dd88f81642) the pmods into the `_connections` list. They forgot to remove the wrapper classes (`Pins`, `IOStandard`) that don't apply to connectors, which @DeanoC fixed.

PR #75, opened 2018-06-18
soc: bios: fix windows build

xobs:
With this patch, litex is able to build the netv2-soc project on Windows.
Currently, the BIOS builds just fine on Windows, but afterwards tries to run
`chmod`.  This command does not exist on Windows, and is unnecessary.

Add a conditional guard to prevent this command from running on Windows.


enjoy-digital:
Thanks!

PR #77, opened 2018-07-05
Enable support for vexriscv debugging

xobs:
Enable debug support for the vexriscv.

The first patch pulls a newer version of vexriscv from upstream, which incldues a version of the verilog file with debug support enabled.

The second patch adds the CSRs to the CPU if instantiated with cpu_debugging=True.

The third patch adds cpu_debugging as an optional flag to SoCCore, allowing the user to instantiate a core with debug support enabled.

The final patch adds a reset line to the UART, which is needed when resetting the CPU.

### DIFF
--- a/litex/soc/cores/cpu/vexriscv/core.py
+++ b/litex/soc/cores/cpu/vexriscv/core.py
@@ -3,18 +3,176 @@ import os
 from migen import *
 
 from litex.soc.interconnect import wishbone
+from litex.soc.interconnect.csr import AutoCSR, CSRStatus, CSRStorage
 
 
-class VexRiscv(Module):
-    def __init__(self, platform, cpu_reset_address):
+class VexRiscv(Module, AutoCSR):
+    def __init__(self, platform, cpu_reset_address, cpu_debugging=False):
         self.ibus = i = wishbone.Interface()
         self.dbus = d = wishbone.Interface()
 
         self.interrupt = Signal(32)
 
+        # Output reset signal -- set to 1 when CPU reset is asserted
+        self.debug_reset = Signal()
+
+        i_debug_bus_cmd_payload_wr = Signal()
+        i_debug_bus_cmd_payload_address = Signal(8)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks! Could we create Vexriscv-debug variant also in LiteX and use the variant parameter instead of cpu_debugging? The reset on the UART is a workaround, we can merge it now, but i'll investigate. Can you describe how to how to use the debug mode and how to reproduce the UART issue?

xobs:
Where would you like me to describe how to use the debug mode?  Currently it requires a C-language openocd<->etherbone adapter and a patched copy of OpenOCD.  Eventually I would like to rewrite the adapter in Python -- I just need to learn enough Python to figure out how to do that.

How would I create a variant?  Would I create a second class in `core.py` called something like `VexRiscvDebug(VexRiscv, AutoCSR)` and implement the logic there?  And then add a new cpu_type in `soc_core.py` of `"vexriscvdebug"` that instantiates that class?

enjoy-digital:
Don't worry for that, i'm going to merge that and then try to add the variant myself. Once you'll have a Python version working, i think you can describe how to use it in the litex-buildenv wiki where we already started documenting things : https://github.com/timvideos/litex-buildenv/wiki
Thanks a lot for the work.

xobs:
Thanks!

As a quick sanity test -- you can verify it works by writing `(1 << 17)` to the CORE register.  This will halt the CPU.  You can then write `(1 << 25)` to the CORE register to resume it.  In theory, writing (`1 << 17)`, `(1 << 16) | (1 << 17)`, and then `(1 << 24) | (1 << 25)` to the CORE register should reset the softcore.

These bits are lightly documented at https://github.com/SpinalHDL/VexRiscv#debugplugin

enjoy-digital:
I just did the changes to use variant instead of cpu_debugging, and did some minor cleanup:
https://github.com/enjoy-digital/litex/commit/c821a0feab5035c7dd70821ecc222cc24f547afa
I just tested compilation and should not have changed functionality.

PR #79, opened 2018-07-06
vexriscv: debug: fix reading DATA register

xobs:
The REFRESH register accepts an 8-bit address and determines which
register to refresh.  Since there are only two addresses currently in
use, this register can be either 0x00 or 0x04.

A refactor replaced the compare with one that checked for any 0 bits.
Since both 0x00 and 0x04 have 0 bits, this check always evaluated as
true, causing the logic to always refresh the CORE register.

Replace this check with an explicit check for 0x00.


enjoy-digital:
Thanks, sorry for the troubles.

PR #80, opened 2018-07-17
vexriscv_debug: use csr read()/write() accessors

xobs:
CSR access widths can be different from register widths.  8-bit
registers are common.

The runtime-generated `read()` and `write()` functions handle this
mapping correctly.  When direct register accesses are handled, this
mapping is lost.

Use the accessor functions rather than directly accessing the memory
addresses, so that we work on platforms other than 32-bit-wide.


enjoy-digital:
Thanks, merged.

PR #81, opened 2018-07-27
Push Vexriscv debug directly on the Wishbone bus

xobs:
With https://github.com/SpinalHDL/openocd_riscv/commit/b649408080c8918c2d53bf9bd935fd9af70b48ba merged, openocd now natively supports the etherbone protocol provided by `litex_server`.

Convert the Vexriscv Debug bus to be available directly on Wishbone.  This provides the following benefits:

1. By using `register_mem()`, the interface offset address is always known.  With a CSR it can move around.
1. By communicating directly with Wishbone, we can monitor the `.we` signal, removing the need for a `TRIGGER` register.  This doubles throughput on reads.
1. Wishbone supports 32-bit transactions, whereas some devices require 8-bit CSRs.  This speeds up accesses by a further 4x.
1. With the integration into `openocd_riscv`, we can remove `vexriscv_debug` and let `openocd_riscv` communicate directly with `litex_server`.


enjoy-digital:
Thanks! merged.

mithro:
@xobs so awesome!

PR #86, opened 2018-08-12
Fix generating csr.csv file

pgielda:
Fix generating csr.csv file when no absolute path is given.


enjoy-digital:
Merged, thanks!

PR #88, opened 2018-08-20
board_tuning parameters added

bunnie:
there will be another PR on the litedram project to match this


enjoy-digital:
Thanks. I moved sdram_init.py to litedram yesterday and just integrated your changes.

PR #91, opened 2018-08-23
.gitignore: litex/build contains valid source, so exclude from .gitig…

cr1901:
…nore.

Self-explanatory. This was causing Atom tree-view, which respects `.gitignore`, to hide `litex/build`, so I tweaked the `.gitignore` to unhide it :).


enjoy-digital:
Thank you both.

PR #92, opened 2018-08-23
software/bios: Gate flush_l2_cache() if L2 Cache isn't present.

cr1901:
I assume this is an oversight, but without using `#ifdef L2_SIZE` to gate `flush_l2_cache()`, I would get undefined symbol errors when attempting to link a BIOS on systems without L2 Cache (such as `tinyfpga`). This commit fixes my issues.


enjoy-digital:
Thank you both.

PR #99, opened 2018-09-08
Add COPY_TO_MAIN_RAM generated Makefile variable to distinguish systems with/without "main_ram" region.

cr1901:
The `litex-buildenv` firmware linker script at present depends on `main_ram` being present, for example. This Makefile variable allows us to choose the proper linker script to use in case we boot from spi flash instead.


enjoy-digital:
It seems there is also a VexRiscv update in the pull request. Can you split that?

enjoy-digital:
Thanks.

PR #100, opened 2018-09-08
lattice/programmer: Use --program-image option with tinyprog if addre…

cr1901:
…ss is given.

This allows a user to program the concatenated gateware+bios+firmware image with a single command on platforms which support `tinyprog`.


enjoy-digital:
Thanks.

PR #108, opened 2018-09-22
Use csr accessors when generating `csr.h`

xobs:
This patchset replaces MMPTR() with csr_readl() and csr_writel().  These functions are defined as inline, unless `CSR_ACCESSORS_DEFINED` is defined when `csr.h` is included.

This name can be changed, if desired.

This can be used with `etherbone.c` available at https://github.com/xobs/wishbone-adapter/

An example of using this is to start up `litex_server` and build the following file in `dna.c`:

```c++
#define CSR_ACCESSORS_DEFINED
#include <stdio.h>
#include "csr.h"
#include "etherbone.h"

static struct eb_connection *g_eb;

void csr_writel(uint32_t value, uint32_t addr) {
        eb_write32(g_eb, value, addr);
}

uint32_t csr_readl(uint32_t addr) {
        return eb_read32(g_eb, addr);
}

int main(int argc, char **argv) {
        g_eb = eb_connect("127.0.0.1", "1234", 0);

        printf("DNA ID: %lld\n", dna_id_read());

        eb_disconnect(&g_eb);
        return 0;
}
```

Compile with:

```sh
$ gcc dna.c etherbone.c -o dna
```

This uses `dna_id_read()` as generated by litex.  It is easy to see how this could be extended to write complete drivers on a host.

### DIFF
--- a/litex/soc/integration/cpu_interface.py
+++ b/litex/soc/integration/cpu_interface.py
@@ -123,12 +123,12 @@ def _get_rw_functions_c(reg_name, reg_base, nwords, busword, read_only, with_acc
     if with_access_functions:
         r += "static inline "+ctype+" "+reg_name+"_read(void) {\n"
         if size > 1:
-            r += "\t"+ctype+" r = MMPTR("+hex(reg_base)+");\n"
+            r += "\t"+ctype+" r = csr_readl("+hex(reg_base)+");\n"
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks!

enjoy-digital:
Tested with lite_xim, works fine :), thanks again.

PR #112, opened 2018-10-04
build/platforms: Add ice40_hx8k_b_evn from Migen.

cr1901:
Adding a new board preemptively so that `litex-buildenv` platforms can [inherit](https://github.com/timvideos/litex-buildenv/issues/28) from `litex` platforms in the future.


enjoy-digital:
Thanks merged! (Note that it should not be a problem for `litex-buildenv` to rely directly in 'migen.build' (tools and platforms) if you want to avoid duplication)

PR #116, opened 2018-10-09
sim: serial: Send '\r\n' instead of just '\n'

stffrdhrn:
This fixes an issue when running with the HDMI2USB firmware which
expects \r\n to come from the UART.  Since the verilator adapter
is just sending \n commands cannot be executed.

Also, one minor whitespace cleanup. (could remove if needed)

### DIFF
--- a/litex/build/sim/core/modules/serial2console/serial2console.c
+++ b/litex/build/sim/core/modules/serial2console/serial2console.c
@@ -76,6 +76,11 @@ void read_handler(int fd, short event, void *arg)
   int i;
   read_len = read(fd, buffer, 1024);
   for(i = 0; i < read_len; i++) {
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks. This was changing behaviour of LiteX bios in the simulation (two carriage returns detected). I just changed the bios code to handle all kind of carriage return: https://github.com/enjoy-digital/litex/commit/10624c26da9577b7228e1af28ee26389bce7f905, so this is fine now. Maybe you should also add that to HDMI2USB firmware to improve compatibility with various terminals.

stffrdhrn:
Right,

Btw, We can work around the 2 \r's with a termios setting.  But I see your
point. I'll post the firmware patch to @mithro to see what his thoughts are.

PR #118, opened 2018-10-27
uart: Enable buffering the FIFO.

mithro:
On the iCE40 FPGA, adding buffering allows the SyncFIFO to be placed in
block RAM rather than consuming a large amount of resources.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -157,7 +157,7 @@ def _get_uart_fifo(depth, sink_cd="sys", source_cd="sys"):
         fifo = stream.AsyncFIFO([("data", 8)], depth)
         return ClockDomainsRenamer({"write": sink_cd, "read": source_cd})(fifo)
     else:
-        return stream.SyncFIFO([("data", 8)], depth)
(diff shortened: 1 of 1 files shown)


cr1901:
@mithro Is there any possibility that you could add a `buffered_fifo=False` parameter to the UART core that's propagated up to `BaseSoC` (using `buffered_uart_fifo=False`). This way a user can tailor whether BRAM resources or distributed RAM resources are used depending on their FPGA family.

Alternatively, instead of propagating the argument up to `BaseSoC`'s constructors, ice40 `base` targets instead instantiate their UARTs manually (since it _is_ a special case).

mithro:
@cr1901 That would add a lot of extra complexity and I'm not sure there is really any gain to doing so? Do you have any evidence that adding the buffer on other platforms negatively affects them?

cr1901:
Talked with @mithro on IRC, I retract my suggestion :).

enjoy-digital:
Thanks. I just tested implementation with both on Arty and resource usage is almost similar. We can use it as default and no need to add a parameter.

PR #123, opened 2018-11-01
PicoRV32 Enhancements

cr1901:
This PR adds the following features to the PicoRV32 core:
* Refactor PicoRV32's `core.py` file to support variants.
* Make sure IRQ vector is always located 16 bytes after the reset vector (was hardcoded to `0x00000010`).
* Clear `.bss` in `crt0` before running main program (required if using FPGA RAM that doesn't clear itself when a bitstream is loaded, such as up5k's `SPRAM`, or ASICs :))
* Emulate support for a relocatable `IRQ` vector; required for `fupy`. I use the `q2` register to hold the "true" IRQ vector location.


enjoy-digital:
Thanks! merged.

PR #124, opened 2018-11-02
build/sim/verilator: don't use --threads when $(THREADS) is unset

jfng:
The --thread option must be provided with a value or the build will fail.


enjoy-digital:
Thanks.

PR #129, opened 2018-11-27
vivado's version smaller than 2016 will not need XPM_LIBRARIES.

Readon:
It is only tested on vivado 2015.2 which is only version I have.


enjoy-digital:
Thanks for the pull request. Since XPM is only useful is a couple of case and i just added a parameter to enable it when needed. (looking the version was probably working on Linux but would need to be tested on Windows, so for now it's better to just disable XPM by default).

Issue #143, opened 2018-12-20
Support multiple period constraits with nextpnr


xobs:
Right now, nextpnr will pick the highest frequency and try to make all clock domains work at that frequency.

`nextpnr` supports varying clock domains by passing a python program in a file specified by `--pre-pack`.

This is documented at https://github.com/YosysHQ/nextpnr/blob/master/docs/constraints.md#clock-constraints

This is accomplished in migen already: https://github.com/m-labs/migen/blob/master/migen/build/lattice/icestorm.py#L28

PR #144, opened 2018-12-20
Integrate latest migen changes for lattice/icestorm.

mithro:
Integrated up to 37db6bb52532b6d1c6bc8b724c2e8c6a38546c2a.

Fixes https://github.com/enjoy-digital/litex/issues/143


enjoy-digital:
@mithro: keeping the litex.build is interesting for me, but for info, in litex-buildenv, the platform could be directly using migen.build and you would avoid that. LiteX would just be used to build the SoC.

PR #148, opened 2019-02-22
versa_ecp5: Remove negative diff IO pins

daveshah1:
In Lattice FPGAs only the positive side of differential pairs should
be specified (unlike Xilinx)

These are a warning on Diamond (which trims unused IO) and an error
with Yosys/nextpnr (which doesn't so they conflict when the positive
pin is 'expanded').

Already this is the case for the clock input, this commit performs
the same change for the DDR3 pins.


enjoy-digital:
Thanks for the patch and explanation.

PR #149, opened 2019-02-25
Add trellis build option to versa_ecp5 and bring trellis support up to date

daveshah1:
See also enjoy-digital/litedram#76

Tested with YosysHQ/Yosys#794 and YosysHQ/nextpnr#219


enjoy-digital:
Thanks!

PR #152, opened 2019-03-06
build/lattice/trellis: generate bitstream directly in svf format

gsomlo:
Before being able to program the board (e.g., with openocd), one
would have to convert the bitstream file to .svf using a python
script included with the source trellis distribution. However, the
trellis 'ecppack' utility can now generate .svf formatted bitstream
directly.


enjoy-digital:
Thanks, can we keep both .bit and .svf generated?

gsomlo:
On Wed, Mar 06, 2019 at 01:21:54PM -0800, enjoy-digital wrote:
> Thanks, can we keep both .bit and .svf generated?

Yes, updated PR now does that.

enjoy-digital:
Thanks!

PR #153, opened 2019-03-07
[fix] utils was omitted when installed from pip

chmousset:
module utils needed by litex_read_verilog and similar:

```
litex_read_verilog ../rtl/spileds.v 
  File "/home/max/ZYNQ/fpga_builder/.ve/bin/litex_read_verilog", line 6, in <module>
    from litex.utils.litex_read_verilog import main
ModuleNotFoundError: No module named 'litex.utils'
```


enjoy-digital:
Thanks.

PR #154, opened 2019-03-22
build/xilinx: Update Yosys write_edif parameters

daveshah1:
See YosysHQ/yosys#830


enjoy-digital:
Thanks, this could explain the issue we have with Yosys used for synthesis, Vivado for place and route. I'm going to do a test.

mithro:
@enjoy-digital Excited to see if this does fix the issue!

@eddiehung This might be relevant to the discussion we were having with Yosys->Vivado not working for certain designs.

PR #156, opened 2019-03-27
soc/interconnect/axi: address length cleanup

gsomlo:
Data width assertion should require AXI and Wishbone width to be
equal, regardless of the actual value (e.g., 32 or 64).
Also, use "axi.address_width" instead of hard-coded "32" for
_[read|write]_addr signals (since we already asserted that it
should be 32).

### DIFF
--- a/litex/soc/interconnect/axi.py
+++ b/litex/soc/interconnect/axi.py
@@ -60,12 +60,12 @@ class AXIInterface(Record):
 
 class AXI2Wishbone(Module):
     def __init__(self, axi, wishbone, base_address):
-        assert axi.data_width    == 32
-        assert axi.address_width == 32
+        assert axi.data_width    == len(wishbone.dat_r)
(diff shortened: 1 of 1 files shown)


litex/soc/interconnect/axi.py
```diff
@@ -60,12 +60,12 @@ def __init__(self, data_width, address_width, id_width=1, clock_domain="sys"):
 
 class AXI2Wishbone(Module):
     def __init__(self, axi, wishbone, base_address):
-        assert axi.data_width    == 32
+        assert axi.data_width    == len(wishbone.dat_r)
         assert axi.address_width == 32
```
gsomlo: maybe also replace "32" with "`assert axi.address_width == len(wishbone.adr) + 2`" ?  (to minimize hard-coded magic numbers)

enjoy-digital:
Thanks, merged.

PR #157, opened 2019-03-31
Add ifdef check for MAIN_RAM_SIZE

CBJamo:
Small boards like those with ice40s won't have main ram, so the bios fails to build.


enjoy-digital:
Thanks!

PR #158, opened 2019-04-08
Changes for litepcie support for Altera Cyclone V

vbuitvydas:
Contributing to make support for Altera Cyclone V:
- added reset synchronizer;
- changes to make top level assigment in .qsf file with build name instead of always using "top"


enjoy-digital:
Thanks! merged.

PR #162, opened 2019-04-17
Add full and full_debug CPU variant of VexRiscv

mateusz-holenko:
This adds two additional variants of VexRiscv CPU: `full` and `full_debug` with CsrPlugin configured in 'all' mode, supporting ecall instruction.


enjoy-digital:
Thanks. Is it going to be different than the Linux variant?

mateusz-holenko:
Linux variant has different mmu and interrupts handling. This one just adds support for `ecall`, so we can run Zephyr on it (https://github.com/zephyrproject-rtos/zephyr/pull/14915).

enjoy-digital:
OK thanks, merged.

PR #163, opened 2019-04-17
build/sim/core: Initialize Verilator commandArgs

gsomlo:
Required when DUT is using plusargs. Prevents Verilator simulation
from crashing with "Verilog called $test$plusargs or $value$plusargs
without testbench C first calling Verilated::commandArgs(argc,argv)".


enjoy-digital:
Merged, thanks!

PR #164, opened 2019-04-19
Litex usb server support

xobs:
This adds preliminary support to `litex_server` for the new USB protocol.

### DIFF
--- a/litex/soc/tools/remote/__init__.py
+++ b/litex/soc/tools/remote/__init__.py
@@ -1,4 +1,5 @@
 from litex.soc.tools.remote.comm_uart import CommUART
 from litex.soc.tools.remote.comm_udp import CommUDP
 from litex.soc.tools.remote.comm_pcie import CommPCIe
+from litex.soc.tools.remote.comm_usb import CommUSB
 from litex.soc.tools.remote.litex_client import RemoteClient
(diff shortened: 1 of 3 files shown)


xobs:
I added a diagram to comm_usb.py describing the protocol.  I also made it more robust against reconnections.

enjoy-digital:
Thanks, merged.

xobs:
I noticed that `litex/litex/__init__.py` has:

```python
from litex.soc.tools.remote import RemoteClient
```

This calls `litex/soc/tools/remote/__init__.py` which now adds:

```python
from litex.soc.tools.remote.comm_usb import CommUSB
```

Because of this, the `usb` module is now a dependency for litex.  Is this a thing that we want, or should I remove it from `remote/__init__.py`?  Or should I wrap the `import usb` in a `try` block?

enjoy-digital:
Thanks for reporting. I just reorganized a bit the tools and fix this issue with: https://github.com/enjoy-digital/litex/commit/9ee6c35b42be6eadcf9fdd3c4497e1523900f802

In an init we were doing an import if CommXXYY to have shorter name when used in scripts, but that's no longer needed since most of the scripts are now using RemoteClient directly.

PR #165, opened 2019-04-19
Vexriscv cpu reset address

xobs:
Add the ability to specify the CPU reset address as a CSR.  This is useful for debugging, as we can change the reset address and then hit the "resetIt" bit on the debug port.


enjoy-digital:
Thanks.

PR #167, opened 2019-04-20
litex_server: check socket flags exist before using them

xobs:
Some flags are only available on certain platforms.  Verify these flags
exist prior to using them when opening a socket.

The latest patch doesn't work anymore on Windows.  This fixes that.

See
https://stackoverflow.com/questions/14388706/socket-options-so-reuseaddr-and-so-reuseport-how-do-they-differ-do-they-mean-t
for more information


enjoy-digital:
Thanks.

PR #170, opened 2019-04-22
build/xilinx/vivado: only try Xilinx setup if vivado is not already i…

ldoolitt:
…n the path

Only affects the non-Windows code path.
Uses python distutils, already used elsewhere.


enjoy-digital:
Thanks yes that's useful, merged.

PR #171, opened 2019-04-23
Install development packages in the user directory

keesj:
When in development mode install the packages in the user directory using the
--user flag from pip. This allows to install and run without the need for root
access.


enjoy-digital:
Thanks, i merged it by made it an option for now with https://github.com/enjoy-digital/litex/commit/228f2867476a76508de8ecb8c15026df9b52ff4f since it was breaking CI.

PR #173, opened 2019-04-24
build: handle exceptional case when litex/migen not deployed as git repo

gsomlo:


enjoy-digital:
Merged, thanks!

litex/build/tools.py
```diff
@@ -95,15 +95,21 @@ def get_migen_git_revision():
     import migen
     d = os.getcwd()
     os.chdir(os.path.dirname(migen.__file__))
-    r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    try:
+        r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    except:
+        r = "--------"
```
mithro: It would have been better to have this as `UNKNOWN`?

litex/build/tools.py
```diff
@@ -95,15 +95,21 @@ def get_migen_git_revision():
     import migen
     d = os.getcwd()
     os.chdir(os.path.dirname(migen.__file__))
-    r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    try:
+        r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    except:
```
mithro: A bare accept is generally bad... Could you catch the right exceptions?

litex/build/tools.py
```diff
@@ -95,15 +95,21 @@ def get_migen_git_revision():
...
     return r
 
 def get_litex_git_revision():
     import litex
     d = os.getcwd()
     os.chdir(os.path.dirname(litex.__file__))
-    r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    try:
+        r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
+    except:
```
mithro: Same as above.

litex/build/tools.py
```diff
@@ -95,15 +95,21 @@ def get_migen_git_revision():
...
+        r = "--------"
```
mithro: Same as above...

gsomlo:
Thanks for the feedback! I think we'd have to open a new PR, since
this one was already applied. That said, see replies inline:

On Wed, Apr 24, 2019 at 05:55:23PM -0700, Tim Ansell wrote:
> +    except:
> +        r = "--------"
> 
> It would have been better to have this as UNKNOWN?

"--------" had already been used in a previous/intermediate version of
get_*_git_revision(), so I just went with it. OTOH, I'm OK with picking
something else that's more informative.

> +    try:
> +        r = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"])[:-1].decode("utf-8")
> +    except:
> +        r = "--------"
> 
> A bare except is generally bad... Could you catch the right exceptions?

I wanted a default value for r in case `git rev-parse --short HEAD`
failed for ANY reason whatsoever (git command not found, git command
failed due to not-a-repo or any other reason, etc.). Narrowing it down
to just "the right" exceptions would require me to know which ones I
want to cover, and also that there are others I *don't* want to cover.
AFICT, that second set is empty. But I'm only a middling Python
programmer, so I can be convinced I'm wrong :)

Thanks again,
--Gabriel

enjoy-digital:
Thanks @mithro.

As @gsomlo said, we just want here a simple way to get migen/litex git sha1 when possible. (deployed as a git repo). We could still improve that in the future to handle the cases where it's not deployed as a git repo, and then try to catch the exceptions.

PR #174, opened 2019-04-26
soc/integration/soc_core: Avoid an overlap of IRQ sources in PicoRV32.

mateusz-holenko:
In [PicoRV32](https://github.com/cliffordwolf/picorv32#custom-instructions-for-irq-handling) external IRQ sources 0/1/2 are shared with internal interrupts.

This remaps `NMI`, `timer0` and `uart` interrupts to avoid an overlap.

### DIFF
--- a/litex/boards/targets/ac701.py
+++ b/litex/boards/targets/ac701.py
@@ -77,11 +77,6 @@ class EthernetSoC(BaseSoC):
     }
     csr_map.update(BaseSoC.csr_map)
 
-    interrupt_map = {
-        "ethmac": 3,
-    }
-    interrupt_map.update(BaseSoC.interrupt_map)
-
     mem_map = {
         "ethmac": 0x30000000,  # (shadow @0xb0000000)
     }
@@ -89,7 +84,7 @@ class EthernetSoC(BaseSoC):
 
     def __init__(self, phy="rgmii", **kwargs):
         assert phy in ["rgmii", "1000basex"]
-        BaseSoC.__init__(self, **kwargs)
+        BaseSoC.__init__(self, soc_interrupts=["ethmac"], **kwargs)
 
         if phy == "rgmii":
             self.submodules.ethphy = LiteEthPHYRGMII(self.platform.request("eth_clocks"),
--- a/litex/boards/targets/arty.py
+++ b/litex/boards/targets/arty.py
@@ -81,18 +81,13 @@ class EthernetSoC(BaseSoC):
     }
     csr_map.update(BaseSoC.csr_map)
 
-    interrupt_map = {
-        "ethmac": 3,
-    }
-    interrupt_map.update(BaseSoC.interrupt_map)
-
     mem_map = {
         "ethmac": 0x30000000,  # (shadow @0xb0000000)
     }
     mem_map.update(BaseSoC.mem_map)
 
     def __init__(self, **kwargs):
-        BaseSoC.__init__(self, **kwargs)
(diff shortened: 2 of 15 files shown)


mateusz-holenko:
Now I see that in https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/arty.py#L85 `ethmac` is mapped to IRQ 3 so it creates a conflict with my proposed mapping for PicoRV32.

It seems it would be better to have `NMI`, `timer0` and `uart` IRQ mapping stable (i.e., independent of the actual core) so it's easier to define targets. What do you think?

futaris:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/include/base/irq.h

https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-picorv32.S

https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/picorv32/core.py#L117

enjoy-digital:
@mateusz-holenko thanks, i'll look at that soon.

enjoy-digital:
@mateusz-holenko: what about, 

- reserving IRQs 0 to 7 for the CPU and SoCCore (allowing user IRQs from 8 to 31).
- for each CPU, create a `reserved_interrupts` property that would provide the interrupts reserved for the CPU, in the case of PicoRV32 it would be:
````python
    @property
    def reserved_interrupts(self):
        return {0: "timer", 1: "ebreak_ecall", 2: "bus_error"}
````
This would also eliminate the `reserve_nmi_interrupt` for mor1kx.
- SoCCore would then choose free interrupts in the 8 first IRQs for timer0 and uart.

mateusz-holenko:
I implemented a solution based on your suggestions. 
I had to rework interrupts significantly, so please verify if this is what you meant.

I'm not sure about `nmi` though. It seemed not to be `mor1kx` related - it was enabled by default for all types of cores and the only place disabling it was `litedram/examples/litedram_gen.py`. Right now I reserved it only for `mor1kx`, but it is a breaking change.

enjoy-digital:
Thanks, i did a quick review and will try to integrate that tomorrow.
I'm not sure we want to change the way the interrupts are defined by the user (otherwise it will break multiple designs). I'll find the best way to do what you want and avoid breaking the API.

enjoy-digital:
@mateusz-holenko can you check if https://github.com/enjoy-digital/litex/commit/8caa38bc256da5098f969449b44662d63e3672fb and https://github.com/enjoy-digital/litex/commit/47dc87584f2bc29e837b4319f516ad3369473fd9 are doing what you want?

I think it should do what you want and keep retro-compatibility. The interrupts specific to SoCCore (uart, timer0) are allocated automatically on the remaining interrupts after CPU and User interrupts are reserved.

mateusz-holenko:
@enjoy-digital Great, I think it does the job. Your solution is definitely less intrusive!

I understand that the conflicts are still possible for a core reserving IRQ #3, but currently it's not an issue. The solution would be simply to redefine targets not to provide explicit number for `ethmac` interrupt.

Thanks for your support.

enjoy-digital:
Yes you can define interrupts like that: https://github.com/enjoy-digital/litex/commit/3f09af6d6e7432e8592c30178ad974acd8f07ed5 and will no longer have the conflict.

mateusz-holenko:
> Yes you can define interrupts like that: [3f09af6](https://github.com/enjoy-digital/litex/commit/3f09af6d6e7432e8592c30178ad974acd8f07ed5) and will no longer have the conflict.

That's even better!

PR #178, opened 2019-05-04
vexriscv: Fix some floating signals

daveshah1:
Yosys was optimising the 1'bx here very eagerly (but correctly) and breaking the CPU as these signals were floating. Tieing explicitly to 1'b0 fixes this.


enjoy-digital:
Thanks!

PR #179, opened 2019-05-08
soc/integration/cpu_interface: more arch-specific address size fixes

gsomlo:
When generating arch-specific include files (generated/[mem|csr].h)
ensure address literal defines are suffixed by 'L', denoting their
'unsigned long' type. This inhibits compiler warnings when values
computed based on these constants are cast to pointers.

Also ensure csr_[read|write][b|w|l]() function declarations have
'unsigned long' address arguments.

Finally, restore the correct (32-bit, (unsigned *)) expected
behavior of the MMPTR() macro, inadvertently converted to an
arch-specific sized access (unsigned long *) by commit 5c2b8685.


enjoy-digital:
Thanks, merged.

PR #182, opened 2019-05-17
boards/nexys4ddr: ethernet support fix-up

gsomlo:
Commit 5f6e7874 added ethernet support, let's now also expose it via
the "--with-ethernet" command line argument.


enjoy-digital:
Thanks!

PR #183, opened 2019-05-21
Use 0x43/0xc3 for USB bridge magic packet

xobs:
The previous value -- 0xc0 -- is used by Windows to request WCID values, which causes the bridge to be completely unusable under Windows.  Furthermore, it breaks enumeration.  So we had to change it to 0x43/0xc3, which is not yet used.

This patchset also adds error printing when the user attempts to open a device when they don't have access to it.


enjoy-digital:
Thanks!

PR #184, opened 2019-05-22
soc/cores : fix readback of current timer value

TomKeddie:


TomKeddie:
@enjoy-digital if you have a moment to comment on my style choices in https://github.com/TomKeddie/prj-litex/tree/master/arty/blink/hw I'd really appreciate it.

enjoy-digital:
Hi @TomKeddie,

i'm not sure to understand the aim of the proposed patch. With the current behaviour, the user need to write the update_value register to trigger an update (and this is the desired behavior). Your patch removes the need for this write to the update_value, can you explain why you want this?

TomKeddie:
Hi @enjoy-digital thanks for your patience.  As you can tell I'm still digging into the intricacies of litex/migen.  I couldn't see the relationship between _value and value but I think you're saying it's a feature of the arichtecture and results in generation of both a value register and an update_value register.  I tried reading top.v but I'm a vhdl guy and my verilog isn't here yet.  Perhaps there is some migen doco to describe  the _value/value relationship?  Thanks.

enjoy-digital:
In this case, the prefix `_` is used to for the CSR (register) and no prefix is use for internal signals.
`_value` is the register that will be read by the software. `value` is the internal counter (that is still running. For the timer, we want to control the update to the register. (since the 32 bits are read in chunks of 8 bits by the software), so when we write to `_update` register, we want the `_value` register to take the internal counter `value`.

PR #185, opened 2019-05-23
tools/litex_sim: restore functionality of '--with-sdram' option

gsomlo:
After LiteDRAM commit #50e1d478, an additional positional argument
('databits') is required by the PhySettings() constructor.

The value used here (32) will generate a 64MByte simulated SDRAM.


enjoy-digital:
Thanks!

PR #188, opened 2019-05-30
Miscellaneous cleanup patches

gsomlo:


enjoy-digital:
Thanks. CSRBank is used by others designs, can you only keep the cpu_interface changes in this pull request?

gsomlo:
> Thanks. CSRBank is used by others designs, can you only keep the cpu_interface changes in this pull request?

Just to confirm, are these two instances of GenericBank():
https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/csr.py#L331
and
https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/wishbone.py#L671
_**NOT**_ redundant w.r.t. each other? If so, I'm happy to drop the second patch and force-push (or to have you pull just the first of the two, whatever makes the most sense) -- Thanks!

enjoy-digital:
Yes, the two instances are not redundant. You can just provide the first patch, thanks.

enjoy-digital:
Thanks.

PR #193, opened 2019-06-04
software/libbase: memcpy: simple, arch-width agnostic implementation

gsomlo:
Remove optimizations targeted specifically at rv32 architecture,
allowing memcpy to work on all word sizes.

Since this is "only" the BIOS, it is also arguably better to
optimize for size rather than performance, given that control
will be quickly handed over to some other program being loaded.


enjoy-digital:
Thanks.

PR #198, opened 2019-06-10
boards/arty : Add directly connected spi clk pin 

TomKeddie:
Avoids need for STARTUPE2.  Config pins tristate after download by default, alternate works well through R262.

shot_2019_06_10_08_34_31

### DIFF
--- a/litex/boards/platforms/arty.py
+++ b/litex/boards/platforms/arty.py
@@ -79,13 +79,15 @@ _io = [
         IOStandard("LVCMOS33"),
     ),
 
-    ("spiflash4x", 0,  # clock needs to be accessed through STARTUPE2
+    ("spiflash4x", 0,
         Subsignal("cs_n", Pins("L13")),
+        Subsignal("clk", Pins("L16")),
         Subsignal("dq", Pins("K17", "K18", "L14", "M14")),
         IOStandard("LVCMOS33")
     ),
-    ("spiflash", 0,  # clock needs to be accessed through STARTUPE2
+    ("spiflash", 0,
         Subsignal("cs_n", Pins("L13")),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks. Can you explain a bit more what you want to do? I'm don't think it's possible to drive L16 pins without using a special STARTUPE2 primitive, at least that's what i read and understood a few years ago. Do you know another way to drive this pin? If not, then we still need to use the STARTUPE2 primitive and having this pin defined in the platform file is misleading.

TomKeddie:
Hi, here is a wider view of the schematic. QSPI_SCK is connected to two pins.  E9 and L16 (via R262).  You need to use STARTUPE2 to drive E9 but it's much clearer to drive E16 directly.  I'm doing that in this PR https://github.com/tinyfpga/TinyFPGA-Bootloader/pull/30  and in this example https://github.com/TomKeddie/prj-litex/blob/master/arty/usb/hw/arty-usb.py 

It might still be necessary to use STARTUPE2 to drive E9 high-Z after config, I need to research this further.  I thought the config pins went tristate after download by default but I can't find it in UG470 just now.  It seems to work but that isn't sufficient of course.

shot_2019_06_10_09_43_23

enjoy-digital:
Ah ok thanks, i wasn't aware of that! If the STARTUPE2 drives the CLK to  High-Z, then yes we could do that. (that's probably the case, i'll also have a look).

enjoy-digital:
Merged, thanks. I'll now look if the same can be done on the others Digilent boards.

enjoy-digital:
Strangely, the trick does not seem to be used on the others board.

TomKeddie:
Thanks. :+1:

PR #199, opened 2019-06-12
bios: Fix build when ethphy is present but ethmac is not.

ambrop72:
While testing my Ethernet DMA, I renamed the `ethmac` module to `ethmac_dma` so that it wouldn't be used from the BIOS, but I got an undefined reference to `eth_init` because `bios.c` checks different CSR defines than the code that defines `eth_init`.


enjoy-digital:
Thanks for catching/fixing it, merged.

PR #201, opened 2019-06-18
tools/litex_sim: fix default endianness for mem_init

gsomlo:
Initializing ROM and/or RAM content requires knowing the CPU
endianness before the SimSoC->SoCSDRAM->SoCCore constructor
sequence is invoked (before the SoC's self.cpu.endianness
could be accessed). Given that the majority of supported CPU
models use "little", set it as the new default, and override
only for the two models that use "big" endianness.


enjoy-digital:
Thanks, that makes sense, merged.

PR #202, opened 2019-06-20
soc: cores: add up5kspram module

xobs:
The ICE40UP5K has 128 kB of SPRAM that's designed to be used
as memory for a softcore.  This memory is actually 4 16-bit
chunks that we can gang together to give us either 64 kB or
128 kB.

Add a module that will allow us to use this memory in an ICE40.


enjoy-digital:
Thanks!

PR #204, opened 2019-06-25
fw (flash write) command

mateusz-holenko:
This adds a new command to BIOS CLI - *fw* (flash write) - that allows to write uint to flash having a simillar signature to the existing *mw* (memory write) command.


enjoy-digital:
Thanks!

PR #214, opened 2019-07-08
soc_core: additional csr_alignment follow-up fixes

gsomlo:
- Update a few additional places to use DFII_ADDR_SHIFT instead of
  a hard-coded 4, which assumed 32-bit alignment.

- Force 64-bit alignment Rocket -- the only supported configuration!

This is a fixup for commit f4770219, tested on Rocket and 64bit Linux.


enjoy-digital:
Thanks for the review and fixes.

PR #216, opened 2019-07-12
Rework booting Linux on VexRiscv

mateusz-holenko:
This modifies the bios booting logic when running on a platform with VexRiscv in linux configuration.

It first tries to use separate kernel/rootfs/device tree/emulator images and when it fails, falls back to a single `boot.bin` image.

The layout of images in flash is also modified so it coresponds to the location in RAM (mind that images in flash still needs size/crc descriptor in the first 8 bytes).

### DIFF
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -530,6 +530,8 @@ class SoCCore(Module):
             self._constants.append(((name + "_" + constant.name).upper(), constant.value.value))
         for name, value in sorted(self.config.items(), key=itemgetter(0)):
             self._constants.append(("CONFIG_" + name.upper(), value))
+            if isinstance(value, str):
+                self._constants.append(("CONFIG_" + name.upper() + "_" + value, 1))
 
         # Connect interrupts
         if hasattr(self, "cpu"):
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -239,6 +239,7 @@ class SoCCore(Module):
 
         # Add CPU
         self.config["CPU_TYPE"] = str(cpu_type).upper()
+        self.config["CPU_VARIANT"] = str(cpu_variant).upper()
         if cpu_type is not None:
             # CPU selection / instance
             if cpu_type == "lm32":
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -224,6 +224,48 @@ const char *filename, char *buffer)
 
 static const unsigned char macadr[6] = {0x10, 0xe2, 0xd5, 0x00, 0x00, 0x00};
 
+#define ROOTFS_IMAGE_RAM_OFFSET      0x00800000
+#define DEVICE_TREE_IMAGE_RAM_OFFSET 0x01000000
+#define EMULATOR_IMAGE_RAM_OFFSET    0x02000000
+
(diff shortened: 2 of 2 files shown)


enjoy-digital:
Thanks, i'll review that soon. Are you taking into account that the first part of the Flash contains the FPGA bistream? (this means we need to add a offset and mapping in SPI Flash need to be optimized).

mateusz-holenko:
@enjoy-digital No, I haven't thought about other data in the Flash - I tested the code in Renode emulator where I don't need an FPGA bitstream.

The flash layout change is not essential (I just did it to make a booting code more consistent), so if it causes problems I can use the previous offests.

mateusz-holenko:
I wrapped `try_get_kernel_rootfs_dtb_emulator` and `copy_image_from_flash_to_ram` with ifdefs to avoid warnings about unused functions when compiled for targets other than `VexRiscv.linux`.

mateusz-holenko:
@enjoy-digital Now that I think about the changes in flash layout I'm not sure it's done right.
`EMULATOR_IMAGE_OFFSET` is set to 0x02000000 which means that always more than 32MB of flash is necessary, regardless of an actual size of all images.
Denser layout where images are put one after another would utilize flash in optimal way and allow to run on wider range of boards.

enjoy-digital:
By putting images one after another, you need to reflash all the linux binaries if you want to update only one of them, so this has some drawbacks. I think we should allocate a fixed spaced for each binary and ensure we have some margin for the file to grow a bit. Possible overlap could be checked by the tool that will be used to flash the binaries.

I suggest that we use a mapping a bit similar what was done initially in a first time, can you change that? I'll then test on netboot/flashboot on hardware and merge. We'll make the mapping evolve later if needed.

mateusz-holenko:
You are right. I changed the flash layout to use the same slot sizes for images as originally, but with a base address taken from `FLASH_BOOT_ADDRESS` (as previously it was hardocded to 0x50400000).

mateusz-holenko:
@enjoy-digital How are tests on HW going? Have you found anything I should look at?

enjoy-digital:
Sorry for the delay, its merged with the following changes:
- https://github.com/enjoy-digital/litex/commit/4b686dbdb22c8208035a31d7a82d59245bef3dad
- https://github.com/enjoy-digital/litex/commit/af61688d1d86f80210d550fefc24ca247d5754b3

For now, emulator.bin still lives in a separate ram, we need to test first in linux-on-litex-vexriscv with emulator.bin located in main_ram and then apply your change.

Thanks for your work.

mateusz-holenko:
Thanks for the merge @enjoy-digital!

I have one quesiton though - where should `EMULATOR_RAM_BASE` come from? Right now it's used, but not declared anywhere (not even generated in `mem.h`/`csr.h`), so the compilation fails.

enjoy-digital:
It's generated when building https://github.com/litex-hub/linux-on-litex-vexriscv but i'll fix your issue.

enjoy-digital:
Fixed with https://github.com/enjoy-digital/litex/commit/462d12bacc6357b619db9816e31de494358eb96f.

mateusz-holenko:
> It's generated when building https://github.com/litex-hub/linux-on-litex-vexriscv but i'll fix your issue.

Oh, I see. I was using litex-buildenv's targets - that's why I couldn't find it.

> Fixed with 462d12b.

Thanks!

PR #217, opened 2019-07-12
spi: change CSR to CSRStorage

sergachev:
Unless there is a reason not doing so, I would change CSR to CSRStorage (it is only used for writes anyway); this allows using the core in SoCs with csr_data_width=8. 
This is the way it actually works for me.


enjoy-digital:
Indeed, thanks! (i was only using it on designs with csr_data_width=32)

PR #223, opened 2019-07-25
support vivado incremental implementation

sergachev:
Add an option to use read_checkpoint vivado tcl command.
Off by default. To turn it on use 'self.toolchain.incremental_implementation = True' in a platform.


enjoy-digital:
Thanks.

PR #243, opened 2019-08-14
build/xilinx/vivado: improve directive support

sergachev:
1) add directives to all vivado commands needing them - synth, place, route, opt
2) deprecate with_phys_opt -> should be replaced with  vivado_post_place_phys_opt_directive = 'AddRetime'
3) remove a bit of code duplication near 'if platform.verilog_include_paths:'

The default behaviour not changed (all directives set to defaults). with_phys_opt still works as before, but throws a warning.
Directive use example:     platform.toolchain.vivado_synth_directive = 'AreaOptimized_high'


enjoy-digital:
Thanks!

PR #246, opened 2019-08-23
software: use native toolchain for same host, target architectures

gsomlo:
LiteX rightfully assumes that most often the target software must
be cross-compiled from an x86 host platform. However, LiteX can be
also built on a 'linux-riscv64' platform (e.g. Fedora's riscv64
port), where the software for riscv64 targets should be compiled
using the native toolchain.


enjoy-digital:
Thanks!

PR #251, opened 2019-08-30
atlantic JTAG UART working module

AnttiLukats:
there is also changed one property on PolarFire platform, Libero 12.x does only support enhanced constraint flow, but in generic the MicroSemi platform support needs much more work to be usable


enjoy-digital:
Thanks, that's merged and simplified with https://github.com/enjoy-digital/litex/commit/a2938a7ae7dd2cee2befcaf10b2441a50af5f426. JTAGAtlantic is now just a regular UART PHY and can be selected by setting uart_name="jtag_atlantic". This avoid duplication of UART code.

PR #253, opened 2019-09-02
WIP: Enable link-time-optimization on LiteX builds

mithro:
LTO can make both the BIOS and bare metal firmware ~30% smaller than current. 

The only issue seems to be that the compiler needs to include `-lgcc` for the link to succeed.

### DIFF
--- a/litex/soc/software/bios/Makefile
+++ b/litex/soc/software/bios/Makefile
@@ -30,7 +30,8 @@ bios.elf: $(BIOS_DIRECTORY)/linker.ld $(OBJECTS)
 		-L../libnet \
 		-L../libbase \
 		-L../libcompiler_rt \
-		-lnet -lbase-nofloat -lcompiler_rt
+		-lnet -lbase-nofloat -lcompiler_rt \
+		-lgcc
 ifneq ($(OS),Windows_NT)
 	chmod -x $@
 endif
--- a/litex/soc/software/bios/Makefile
+++ b/litex/soc/software/bios/Makefile
@@ -24,7 +24,7 @@ endif
 bios.elf: $(BIOS_DIRECTORY)/linker.ld $(OBJECTS)
 
 %.elf: ../libbase/crt0-$(CPU)-ctr.o ../libnet/libnet.a ../libbase/libbase-nofloat.a ../libcompiler_rt/libcompiler_rt.a
-	$(LD) $(LDFLAGS) -T $(BIOS_DIRECTORY)/linker.ld -N -o $@ \
+	$(CC) $(LDFLAGS) -T $(BIOS_DIRECTORY)/linker.ld -N -o $@ \
 		../libbase/crt0-$(CPU)-ctr.o \
 		$(OBJECTS) \
 		-L../libnet \
--- a/litex/soc/software/common.mak
+++ b/litex/soc/software/common.mak
@@ -49,7 +49,7 @@ INCLUDES = -I$(SOC_DIRECTORY)/software/include/base -I$(SOC_DIRECTORY)/software/
 COMMONFLAGS = $(DEPFLAGS) -Os $(CPUFLAGS) -g3 -fomit-frame-pointer -Wall -fno-builtin -nostdinc $(INCLUDES) -flto -fuse-linker-plugin -Os -ffunction-sections -fdata-sections -flto -nostartfiles -nostdlib -nodefaultlibs
 CFLAGS = $(COMMONFLAGS) -fexceptions -Wstrict-prototypes -Wold-style-definition -Wmissing-prototypes
 CXXFLAGS = $(COMMONFLAGS) -std=c++11 -I$(SOC_DIRECTORY)/software/include/basec++ -fexceptions -fno-rtti -ffreestanding
-LDFLAGS = -nostdlib -nodefaultlibs -L$(BUILDINC_DIRECTORY)
+LDFLAGS = -nostdlib -nodefaultlibs -Os $(CPUFLAGS) -L$(BUILDINC_DIRECTORY)
 #--plugin $(shell $(CC) --print-prog-name liblto_plugin.so)
 
 define compilexx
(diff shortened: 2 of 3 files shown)


mithro:
@mateusz-holenko - Can you take a look at this and see if this works properly and the exact size reduction this gives.

mithro:
@xobs - Can you take a look at this and see if it works for you?

xobs:
@mithro This looks reasonable to me, but I still don't want to use the bios in Foboot, because we don't need it.  The bios includes things such as `printf`, which will at least get reduced down to `puts` in this instance.

mithro:
@xobs This should enable LTO for any firmware built by LiteX (not just the BIOS).

litex/soc/software/bios/Makefile
```diff
@@ -24,13 +24,14 @@ endif
...
-	$(LD) $(LDFLAGS) -T $(BIOS_DIRECTORY)/linker.ld -N -o $@ \
+	$(CC) $(LDFLAGS) -T $(BIOS_DIRECTORY)/linker.ld -N -o $@ \
 		../libbase/crt0-$(CPU)-ctr.o \
 		$(OBJECTS) \
 		-L../libnet \
 		-L../libbase \
 		-L../libcompiler_rt \
-		-lnet -lbase-nofloat -lcompiler_rt
+		-lnet -lbase-nofloat -lcompiler_rt \
+		-lgcc
```
mithro: The requirement for `-lgcc` here when LTO is enabled concerns me...

mateusz-holenko:
@mithro Sorry for a delay, but I've been out for a while.

I tested your commits on top of the current litex master (430fee4) and
here are the results for BIOS for VexRiscv:

```
LTO | disabled | enabled
----+----------+---------
elf | 192 804b | 93 160b (52% smaller)
----+----------+---------
bin |  24 532b | 21 324b (13% smaller)
```

I used the following command to build bios:

    ./litex/litex/boards/targets/arty.py --cpu-type vexriscv --csr-csv csr.csv --with-ethernet --no-compile-gateware

I tested litex-buildenv as well.
I had to modify `firmware/Makefile` a little and force using external toolchain (as the one provided by conda does not support -lgcc switch).
I built my own toolchain from https://github.com/riscv/riscv-gnu-toolchain using:

    ./configure --prefix=/opt/riscv32 --with-arch=rv32g --with-abi=ilp32
    make

Finally I got it working:

```
LTO          | disabled | enabled
-------------+----------+----------
firmware elf | 439 716b | 307 392b (30% smaller)
-------------+----------+----------
firmware bin |  57 088b |  38 068b (33% smaller)
-------------+----------+----------
bios elf     | 210 360b | 102 272b (51% smaller)
-------------+----------+----------
bios bin     |  26 020b |  22 660b (13% smaller)
```

I run emulation in Renode and binaries built with LTO work fine.

mithro:
@mateusz-holenko These results look great! The big question we have is around why the `-lgcc` is needed?

mateusz-holenko:
We've tested again your commits on top of LiteX master from 16.01.2020 (990870d).
Removing `-lgcc` causes linker errors on `__lshrdi3` function. We've managed to
fix this by disabling optimisation for code size (`-Os`) in compiler flags.
Then we replaced `-Os` with `-O2` and other flags that, according to GCC manual,
are enabled by `-Os`. These flags are:
```
-fno-align-functions -fno-align-jumps -fno-align-loops \
-fno-align-labels -fno-prefetch-loop-arrays -findirect-inlining \
-fira-region=one -fira-hoist-pressure -fno-tree-ch \
-freorder-blocks-algorithm=simple -mtune=size
```
None of these is causing the linking process failure as it was with `-Os`.
We've tested code sizes in 4 different configurations:

1. without LTO
1. with LTO, with `-lgcc` and with `-Os`
1. with LTO, without `-lgcc`, with `-O2`
1. with LTO, without `-lgcc`, with `-O2` and other flags (mentioned above)

Here are the results:
```
Configuration |    1.    |    2.    |    3.    |    4.
--------------+----------+----------+----------+----------
bios.elf      | 139 628b |  84 836b |  76 988b |  76 732b (45% smaller than 1.)
--------------+----------+----------+----------+----------
bios.bin      |  19 324b |  17 188b |  18 324b |  17 852b (7.6% smaller than 1.)
```

Only the `-fno-tree-ch` flag added in 4. resulted in lowering `bios.bin` size.
It was from `18 324b` to `17 852b` (472b difference)
However, there are still some flags that are listed in GCC manual as enabled
by `-Os`, such as:
```
-mq-class -msize-level=3 -mdual-nops
```
But when we try to use them, compiler says that they are unrecognized.

mithro:
It seems that the function is the following - https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html;
```
Runtime Function: int __lshrsi3 (int a, int b)
Runtime Function: long __lshrdi3 (long a, int b)
Runtime Function: long long __lshrti3 (long long a, int b)
These functions return the result of logically shifting a right by b bits.
```

I guess LTO is turning something into a logic right shift?

xobs:
I was experiencing errors with `__lshrdi3` as well.  They were due to the use of 64-bit arithmetic inside every CSR access.  This should be fixed with the latest update to CSR accessors once it gets merged.

mateusz-holenko:
We rebased this branch onto the newest master and added the commit replacing compilation flag `-Os` by `-O2` and flags that are enabled by `-Os` according to RISCV GCC manual in order to avoid the need of using `-lgcc`: https://github.com/antmicro/litex/commit/49634cc3f209217e9ad8493e2b27c59cb7c2ebb0

@mithro: this is your PR (based on your `enable-lto` branch), so you can update it to match our version. You might also want to squash commits as we reverted some of your changes.

The LiteX BIOS binary size is as follows:
```
Configuration |    LiteX master    |    LTO with -lgcc    |    LTO without -lgcc   
--------------+--------------------+-----i----------------+----------------------
bios.bin      |            25 140b |              21 844b |              23 700b 
```

enjoy-digital:
Replaced with https://github.com/enjoy-digital/litex/pull/401 that has been merged.

PR #254, opened 2019-09-02
Add @xobs' smaller CRC version

mithro:
@xobs wrote a smaller (code size wise) version of the crc functions in libbase. This merges these implementations into upstream litex which will be enabled if someone adds the `SMALL_CRC` define. This is useful when you have a space constrained location to store the BIOS.


mithro:
@xobs PTAL.

xobs:
@mithro Very nice! This will come in handy on small systems that need crc16/crc32, and would rather trade speed for size.

enjoy-digital:
Thanks!

PR #255, opened 2019-09-09
fix crc32

sergachev:
This 'return 0' breaks BIOS CRC and firmware boot - "BIOS CRC failed (expected a296e241, got 00000000)".


enjoy-digital:
Thanks!

PR #257, opened 2019-09-13
soc/interconnect/csr: add CSRField/documentation support, do some simplification on CSRStorage

enjoy-digital:
This is an experiment to add field and documentation support to CSR.

### DIFF
--- a/litex/soc/interconnect/csr.py
+++ b/litex/soc/interconnect/csr.py
@@ -298,9 +298,6 @@ class CSRStorage(_CompoundCSR):
         Allow the design to update the CSRStorage value. *Warning*: The atomicity of reads by the
          CPU is not guaranteed.
 
-    alignment_bits : int
-        ???
-
     name : string
         Provide (or override the name) of the ``CSRStatus`` register.
 
@@ -325,22 +322,21 @@ class CSRStorage(_CompoundCSR):
         ???
     """
 
-    def __init__(self, size=1, reset=0, fields=[], atomic_write=False, write_from_dev=False, alignment_bits=0, name=None, description=None):
+    def __init__(self, size=1, reset=0, fields=[], atomic_write=False, write_from_dev=False, name=None, description=None):
         if fields != []:
             self.fields = CSRFieldAggregate(fields, "read-write")
             size  = self.fields.get_size()
             reset = self.fields.get_reset()
         _CompoundCSR.__init__(self, size, name)
-        self.alignment_bits = alignment_bits
         self.storage_full = Signal(self.size, reset=reset)
-        self.storage = Signal(self.size - self.alignment_bits, reset=reset >> alignment_bits)
-        self.comb += self.storage.eq(self.storage_full[self.alignment_bits:])
+        self.storage = Signal(self.size, reset=reset)
+        self.comb += self.storage.eq(self.storage_full)
         self.atomic_write = atomic_write
         self.re = Signal()
         if write_from_dev:
             self.we = Signal()
-            self.dat_w = Signal(self.size - self.alignment_bits)
-            self.sync += If(self.we, self.storage_full.eq(self.dat_w << self.alignment_bits))
+            self.dat_w = Signal(self.size)
+            self.sync += If(self.we, self.storage_full.eq(self.dat_w))
         for field in [*fields]:
             field_assign = getattr(self.fields, field.name).eq(self.storage[field.offset:field.offset + field.size])
             if field.pulse:
@@ -359,13 +355,7 @@ class CSRStorage(_CompoundCSR):
             lo = i*busword
             hi = lo+nbits
             # read
-            if lo >= self.alignment_bits:
-                self.comb += sc.w.eq(self.storage_full[lo:hi])
-            elif hi > self.alignment_bits:
-                self.comb += sc.w.eq(Cat(Replicate(0, hi - self.alignment_bits),
-                    self.storage_full[self.alignment_bits:hi]))
-            else:
-                self.comb += sc.w.eq(0)
+            self.comb += sc.w.eq(self.storage_full[lo:hi])
             # write
             if nwords > 1 and self.atomic_write:
                 if i:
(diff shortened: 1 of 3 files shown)


enjoy-digital:
@xobs: can you check is this allow you to do what you want? This is not yet documented, but we could integrate the documentation did since the Field API should be very similar (readable/writable is replace with access parameter). Do you see some feature that are still missing?

litex/soc/interconnect/csr.py
```diff
@@ -239,6 +309,12 @@ def __init__(self, size=1, reset=0, atomic_write=False, write_from_dev=False, al
             self.we = Signal()
             self.dat_w = Signal(self.size - self.alignment_bits)
             self.sync += If(self.we, self.storage_full.eq(self.dat_w << self.alignment_bits))
+        for field in [*fields]:
+            field_assign = getattr(self.fields, field.name).eq(self.storage[field.offset:field.offset + field.size])
+            if field.pulse:
+                self.comb += If(self.storage.re, field_assign)
```
xobs: The `self.storage.re` signal should be `Replicate()`d out to `field.size`, though in reality I can't think of a usecase where `pulse` would be anything but a 1-bit signal.

litex/soc/interconnect/csr.py
```diff
@@ -122,6 +122,66 @@ def do_finalize(self, busword):
...
+                assert field.access == "read-only"
+            elif access == "read-write":
+                assert field.access in ["read-write", "write-only"]
+            setattr(self, field.name, field)
+
+    @staticmethod
+    def check_names(fields):
+        names = []
+        for field in fields:
+            if field.name in names:
```
xobs: Should we make sure names are lowercase? It is possible to make `CSRField("name")` and `CSRField("Name")`, which won't violate a check here but will generate a CPP error in the resulting `csr.h` file.

litex/soc/interconnect/csr.py
```diff
@@ -122,6 +122,66 @@ def do_finalize(self, busword):
         raise NotImplementedError
 
 
+class CSRField(Signal):
+    def __init__(self, name, size=1, offset=None, reset=0, description=None, pulse=False, access=None, values=None):
+        assert name == name.lower()
+        assert access in [None, "write-only", "read-only", "read-write"]
```
xobs: Would it be better to use an enum here, or a text field as here?  I'm unfamiliar with Python Best Practices.

Valid SVD values are (from https://siliconlabs.github.io/Gecko_SDK_Doc/CMSIS/SVD/html/group__svd__xml__fields__gr.html):

```
Describe the manipulation of data written to a field. If not specified, the value written to the field is the value stored in the field. The other options are bitwise operations:

    oneToClear: write data bit of one shall clear (set to zero) the corresponding bit in the field.
    oneToSet: write data bit of one shall set (set to one) the corresponding bit in the field.
    oneToToggle: write data bit of one shall toggle (invert) the corresponding bit in the field.
    zeroToClear: write data bit of zero shall clear (set to zero) the corresponding bit in the field.
    zeroToSet: write data bit of zero shall set (set to one) the corresponding bit in the field.
    zeroToToggle: write data bit of zero shall toggle (invert) the corresponding bit in the field.
    clear: after a write operation all bits in the field are cleared (set to zero).
    set: after a write operation all bits in the field are set (set to one).
    modify: after a write operation all bit in the field may be modified (default).
```

However, this is such a minor thing that I don't think it should hold up adoption of this patchset.

litex/soc/interconnect/csr.py
enjoy-digital: In fact the way we are doing it here, we do the full assignment on `self.storage.re`, so it should be fine, but yes, we should probably limit pulse usage to 1-bit fields.

litex/soc/interconnect/csr.py
enjoy-digital: The check is done directly in the field: https://github.com/enjoy-digital/litex/pull/257/commits/8c080e5fb6cf52342e4e4b5963d9d85e2e90a57a#diff-ff1bf16eefbf92d8e4569d807b041554R127 but indeed, we it would be more consistent to move it to check_names. We are converting the name to uppercase when generating the `csr.h`, so in fact i'm not sure this is really useful. I'll check that.

litex/soc/interconnect/csr.py
enjoy-digital: In fact i'm not sure to understand what is your intend for values: do you want to store possible values of the field + description as you did in https://github.com/xobs/dreg/blob/master/dcsr.py#L290-L301 or the manipulation of data in the field as seem to describe the modifiedWriteValues SVD values? I think both could be useful, but maybe in a first time we should just support what you intended to do in dcsr?

litex/soc/interconnect/csr.py
xobs: I see.  That's a much better way of doing it than I was originally doing it.

litex/soc/interconnect/csr.py
xobs: Ah, that's what `lower()` does.  That works, then!

litex/soc/interconnect/csr.py
xobs: Both.

`values` stores the possible values, plus each description of that value.  Later on we may want to add some validation, ensuring the values are actual valid numbers.  But either way, this describes what the contents of the register can be, and what each value means.

`access` indicates what the value does.  For example, the bits in the `EventManager.pending` are all `oneToClear`, since writing a 1 to that CSR clears the `pending` bit.

The only additional fields we may want to add are `min` and `max`, which would define ranges (such as for clock dividers).

enjoy-digital: Thanks, i agree on the features, maybe i'll first merge the rest of the code and will open another pull-request to discuss this more.

enjoy-digital:
Thanks @xobs for the ideas/feedback. That's now integrated (with some others simplification on CSRStorage). We'll just have to see what is the best for the values parameter in the future, for now it's what you originally did.

PR #259, opened 2019-09-18
timer: add documentation

xobs:
Now that CSRs have documentation support, add documentation to the basic
`Timer` module.

### DIFF
--- a/litex/soc/cores/timer.py
+++ b/litex/soc/cores/timer.py
@@ -10,11 +10,22 @@ from litex.soc.interconnect.csr_eventmanager import *
 
 class Timer(Module, AutoCSR):
     def __init__(self, width=32):
-        self._load = CSRStorage(width)
-        self._reload = CSRStorage(width)
-        self._en = CSRStorage()
-        self._update_value = CSR()
-        self._value = CSRStatus(width)
+        self._load = CSRStorage(width, description="""This is the initial value loaded into the
+                                        timer.  You can make a one-shot timer by disabling the
+                                        timer, writing to this register, and then re-enabling
+                                        the timer.  For a recurring timer, set this to the same
+                                        value as `reload`, or to 0.""")
+        self._reload = CSRStorage(width, description="""The internal timer value will be updated
+                                        with this value whenever it reaches 0.  Use this to create
+                                        a periodic timer that fires whenever this transitions from
+                                        0 to >0.  To create a one-shot timer, leave this value as 0.""")
+        self._en = CSRStorage(fields=[CSRField("en", description="Write a `1` here to start the timer running")])
+        self._update_value = CSRStorage(fields=[CSRField("update", description="""Writing to this register causes
+                                        the `value` register to be updated with with the current countdown
+                                        value.""")])
+        self._value = CSRStatus(width, description="""Last snapshotted value of the countdown
+                                        timer.  This value is only updated when a `1` is written
+                                        to `update_value`.""")
 
         self.submodules.ev = EventManager()
         self.ev.zero = EventSourceProcess()


xobs:
It might make sense to combine `en` and `update_value` into a single CSR, however that is a breaking ABI change and so I've omitted it.

enjoy-digital:
Thanks! Since timer is already used on very various places, that's indeed better to avoid breaking behavior.

enjoy-digital:
@xobs: i change the style a bit and avoided some  redundancy: https://github.com/enjoy-digital/litex/commit/28885064f73edea5d5efb3d31a2464e3b17ee7f4, i hope it's fine for you.

xobs:
Minor typo: " this register need to be set to 0" -> " this register needs to be set to 0".

I'm working on generating documentation based on these values, and I'm finding the `description=` field is completely ignored.  So I propose this additional patch to `csr.py`:

```patch
diff --git a/litex/soc/interconnect/csr.py b/litex/soc/interconnect/csr.py
index e3d2f1ae..228290fa 100644
--- a/litex/soc/interconnect/csr.py
+++ b/litex/soc/interconnect/csr.py
@@ -280,6 +280,7 @@ class CSRStatus(_CompoundCSR):
             size  = self.fields.get_size()
             reset = self.fields.get_reset()
         _CompoundCSR.__init__(self, size, name)
+        self.description = description
         self.status = Signal(self.size, reset=reset)
         for field in fields:
             self.comb += self.status[field.offset:field.offset + field.size].eq(getattr(self.fields, field.name))
@@ -350,6 +351,7 @@ class CSRStorage(_CompoundCSR):
             size  = self.fields.get_size()
             reset = self.fields.get_reset()
         _CompoundCSR.__init__(self, size, name)
+        self.description = description
         self.storage = Signal(self.size, reset=reset)
         self.atomic_write = atomic_write
         self.re = Signal()
```

xobs:
Actually, by using the `CSRField` mechanism, we can generate Wavedrom diagrams directly.  Let me come up with an example for you.

enjoy-digital:
Thanks, i fixed the two previous points (description  still needed to be attached to CSRStatus/CSRStorage).

xobs:
This is an example of using the CSRFields for those bits, rather than a blanket description.

This is an automatically-generated document.  This device has only a Timer and no other CSRs: https://doctest.xobs.io/timer.html#timer0-en

mithro:
@xobs We should definitely make this a sphinx plugin.

xobs:
@enjoy-digital I prefer the wording before 28885064f73edea5d5efb3d31a2464e3b17ee7f4 -- it's less terse, and parses out into a reference manual better.

enjoy-digital:
@xobs: i agree, but in fact realized while doing it that part of the documentation you added should probably be part of a general documentation of the module. What do you think about having a general description of the module and behaviour and then short description in the fields? For example for the Timer, do you find that better: https://github.com/enjoy-digital/litex/commit/a86bd3b9a9ee559d13d169f709e5b5ec61edb157 ?

The automatic documentation could then reuse the general description and give the list of the registers and the description of the fields.

xobs:
@enjoy-digital I do think we should have a more general overview of the module, and I'm not sure how to do that.  The problem with using docstrings that way is it conflates Python documentation with SoC documentation.

For example, when we document the UART (https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/uart.py#L184), we will want to note the Signals that go in and out: rx, tx, phy, possibly a debug bridge... Those belong in Python documentation and not in the resulting SoC documentation.

I think it would be better to have a class-level attribute called something like `module_documentation` or `csr_documentation`.  This would include more advanced things such as waveforms, general "how to use", and any notes about the module.

enjoy-digital:
@xobs: if possible, for readibility of the code, i would prefer having long strings/descriptions recognized as python's comments by text editors. What about using custom docstring sections for that?

For example:
````
    """Timer

    Provides a generic Timer core.

    The Timer is implemented as a countdown timer that can be used in various modes:
    - Polling : Returns current countdown value to software.
    - One-Shot: Loads itself and stops when value reaches 0.
    - Periodic: (Re-)Loads itself when value reaches 0.

    How-to-Use
    ----------
    `en` register allows the user to enable/disable the Timer. When the Timer is enabled, it is
    automatically loaded with the value of `load' register.

    When the Timer reaches 0, it is automatically reloaded with value of `reload` register.

    The user can latch the current countdown value by writing to `update_value` register, it will
    update `value` register with current countdown value.
    ...
 ````
 The Python documentation would collect the whole docstring, and the SoC documentation would just collect the sections that are interesting for using it.

**Comment by xobs** [comment]
I like that approach, and I'd take it one step further and designate a prefix for user-level sections.

For example, the ddr module might have a section on calibration and a different section on memory choices. Meanwhile, litevideo might have a whole section dedicated to TMDS timings.

These kinds of documentation ought to be included in the generated reference manual so a user can verify the signals are good.

This would be useful even for Modules that have no CSRs, such as Spibone or the USB Wishbone bridge, which have things that really ought to be in a reference manual but otherwise aren't documented, or aren't bundled with the soc.

**Comment by enjoy-digital** [comment]
OK good, i'm fine with that and the others possible use-cases.

**Comment by xobs** [comment]
What about a sentinel that indicates "Everything after this is is for the reference manual"?  I'm thinking similar to perldoc `=cut` and `=pod` directives which switch between documentation and code.  Maybe have `=rm` and `=mod` to switch between reference manual documentation and module documentation?

The thing I'm trying to avoid is needing to add a dependency on `doctools` to simply parse documentation.  With a sentinel, we can simply do:

```python
to_rm=False
for line in module.__doc__.splitlines():
    if line == "=rm":
        to_rm = True
    elif line == "=mod":
        to_rm = False
    if to_rm:
        print(line, file=modfile)
```

xobs:
Also, is there a better place to discuss this than a merged PR?

enjoy-digital:
You could create an issue to discuss that if think it's better. We should that what is easier for you, so i'm fine with what you are suggesting. (we could still switch to something else easily in the future).

mithro:
@xobs I would recommend using rst headers / blocks. You can then use docutils to easily extract the sections.

mithro:
An rst directive looks like the following;

```rst
.. glossary::
   :option:

   environment
      A structure where information about all documents under the root is
      saved, and used for cross-referencing.  The environment is pickled
      after the parsing stage, so that successive runs only need to read
      and parse new and changed documents.

   source directory
      The directory which, including its subdirectories, contains all
      source files for one Sphinx project.
```

So you could put the following in a docstring;

```rst

Timer

Provides a generic Timer core.

The Timer is implemented as a countdown timer that can be used in various modes:
- Polling : Returns current countdown value to software.
- One-Shot: Loads itself and stops when value reaches 0.
- Periodic: (Re-)Loads itself when value reaches 0.

.. usage-docs::
   :option-1:
   :option-2:

   `en` register allows the user to enable/disable the Timer. When the Timer is
   enabled, it is automatically loaded with the value of `load' register.

   When the Timer reaches 0, it is automatically reloaded with value of
   `reload` register.

   The user can latch the current countdown value by writing to `update_value`
   register, it will update `value` register with current countdown value.

...
```

sphinx / other Python documentation tools will successfully understand the rst stuff.

If you want to extract it manually, you can do the following (which is almost
identical to how you would do your perldoc version).

```python

lines = textwrap.dedent(docstring).splitlines()
# Strip everything before ".. usage-docs::" identifier.
while not lines[0].startswith(".. usage-docs::"):
  lines.pop(0)

lines.pop(0)

# Get any options around how to include / what to do with this usage docs.
options = []
while lines[0].strip().startswith('    :'):
  assert lines[0][-1] == ':'
  options.append(lines.pop(0)[5:-1])

section_lines = []
while (not lines.strip()) or lines[0].startswith('    '):
  section_lines.append(lines.pop(0))

return textwarp.dedent("\n".join(section_lines))
```

xobs:
Added an Issue at https://github.com/enjoy-digital/litex/issues/265

PR #260, opened 2019-09-18
soc_sdram: fix build when l2 cache disabled

gsomlo:
When L2 is disabled (via l2_size=0), we must still connect the
sdram crossbar port to a wishbone slave interface. In this case,
that should be the interface directly facing the CPU, instead of
the one connecting to the L2 cache.

### DIFF
--- a/litex/soc/integration/soc_sdram.py
+++ b/litex/soc/integration/soc_sdram.py
@@ -78,9 +78,10 @@ class SoCSDRAM(SoCCore):
         self.add_wb_sdram_if(wb_sdram)
         self.register_mem("main_ram", self.mem_map["main_ram"], wb_sdram, main_ram_size)
 
+        # insert optional l2_cache inbetween DRAM and exposed wb interface:
+        port = self.sdram.crossbar.get_port()
+        port.data_width = 2**int(log2(port.data_width)) # Round to nearest power of 2
         if self.l2_size:
-            port = self.sdram.crossbar.get_port()
-            port.data_width = 2**int(log2(port.data_width)) # Round to nearest power of 2
             l2_size         = 2**int(log2(self.l2_size))    # Round to nearest power of 2
             l2_cache = wishbone.Cache(l2_size//4, self._wb_sdram, wishbone.Interface(port.data_width))
             # XXX Vivado ->2018.2 workaround, Vivado is not able to map correctly our L2 cache.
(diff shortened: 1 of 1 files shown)


gsomlo:
For context, I'm tinkering with the way the cpu connects to the l2 cache connects to LiteDRAM, in an attempt to comprehend how it all hangs together. During said tinkering I noticed I get an error if I try to just disable the l2 cache, so I fixed that, and figured I might as well share that fix while I'm at it :)

mithro:
Disabling the cache in LiteDRAM seems like a useful debugging too, so I would support this.

gsomlo:
On Wed, Sep 18, 2019 at 09:07:13AM -0700, Tim Ansell wrote:
> Disabling the cache in LiteDRAM seems like a useful debugging too, so I would
> support this.

The presence of the "l2_size" argument to SoCSDRAM's __init__(), along
with the pre-existing "if self.l2_size: <create_and_connect l2 cache>"
code path suggested it was already supposed to be supported, except
that support was broken (likely due to extremely infrequent/unlikely
use).

My personal interest is in bypassing the wishbone "main bus"
altogether, and connecting the (AXI) LiteDRAM slave interface
directly to Rocket's cached-memory AXI master interface. But
I gotta walk before I can run, and understanding how things
work right now (with the occasional "drive-by" fix) is part
of that process... :)

enjoy-digital:
In fact the l2 cache is used for 2 things here:
- cache (no secret :))
- data_width adaptation.
And we don't currently have another good option for bi-directionnal data_width adaptation, so in the best case, we still need to have the l2 cache, but with a minimal size just to do the data_width adaptation. Maybe it would be best to change the code in this direction: don't make the l2_cache optional, but set the size the minimum if l2_size is 0.

This new assertion in LiteDRAM code should allow catch the issue your code: https://github.com/enjoy-digital/litedram/commit/d93dded6249332c049d27bb784e3d6f530a130e4

gsomlo:
On Wed, Sep 18, 2019 at 12:18:21PM -0700, enjoy-digital wrote:
>   • data_width adaptation.
>     And we don't currently have another good option for bi-directionnal
>     data_width adaptation, so in the best case, we still need to have the l2
>     cache, but with a minimal size just to do the data_width adaptation. Maybe
>     it would be best to change the code in this direction: don't make the
>     l2_cache optional, but set the size the minimum if l2_size is 0.

Or, we could simply `assert l2_size >= 1024` (or whatever you think
the acceptable minimum should be for strictly "UpDownConverter"
purposes?

Speaking of, what about the soc.interconnect.wishbone.Converter module? Not that I'm advocating for its use here, but just curious why it wouldn't work.

gsomlo:
turns out, soc.interconnect.wishbone.Converter seems to work if all we need is to convert between the width of the wishbone system bus and the width of LiteDRAM port.
I'm not insisting on this change (still happy to just add an "assert l2_size >= minimum" and be done with it), but since my goal is to learn about the internals of Lite[X|DRAM], here we go :)

enjoy-digital:
With https://github.com/enjoy-digital/litex/commit/e97c1e36fb5773c3b5e09b854ddeca979ef9728d, we now use a minimal l2_size computed from the sdram data_width.

The Wishbone Converter should be very similar to the L2 cache with the minimal size (UpConverter was probably coded reusing some code from the L2 cache), so i don't think that's worth selecting between L2 Cache and Converter.

For your case, if you want to have a 64-bit data path between the CPU and the SDRAM here is what i can think of:
- 1) only use 64-bit Wishbone for all peripherals of the SoC. (pass data_width parameter to all wishbone.Interface() and check/remove limitation is some modules are hardcoded for 32-bit data_width).
- 2) have 2 Wishbone buses in your SoC the original 32-bit one + a 64-bit one. The 64-bit one would be directly connected to the CPU and L2 Cache and would bridge to the 32-bit one to access the others peripherals.

1) would probably be better in the long term, but 2) is probably easier to do.

gsomlo:
I think I'd like to do *both*:
- use 64bit wb for peripherals, and connect only the "mmio" AXI port from Rocket to that bus;
- have a second bus, which in this case can be native 64-bit AXI, connecting the "mem" AXI port from Rocket directly to LiteDRAM's data port.

Thanks for the explanations -- I'm closing this PR now that it's essentially obsoleted by commit e97c1e36.

PR #261, opened 2019-09-19
csr_eventmanager: add `name` and `description` args

xobs:
Add `name` and `description` as optional arguments to the various
EventSource types.  These default to `None`, so this should be a
backwards-compatible change.

Use the same trick as CSRs, where we default the `name` to be the
instantiated object name as read from the Migen `get_obj_var_name()`
call.


enjoy-digital:
Thanks!

PR #262, opened 2019-09-20
vivado just needs to be in the path for the programmer as well

jersey99:
Maintaining parity with https://github.com/enjoy-digital/litex/pull/170


enjoy-digital:
Thanks.

PR #264, opened 2019-09-23
Enable to run Linux on mork1x

mateusz-holenko:
This increases integrated ROM size to fit bios with ethernet for mork1x and adapts memory map for this CPU to enable mainline Linux.


enjoy-digital:
Thanks, merged.

PR #266, opened 2019-09-24
Add ModuleDoc and AutoDoc

xobs:
This adds two new classes: `ModuleDoc` and `AutoDoc`.

`ModuleDoc` is used to document a section of a `Module`.  When derived as a subclass, such as what this patch does with `Timer`, it produces a documentation section such as https://doctest.xobs.io/timer0.html#timer

`AutoDoc` adds the ability to attach additional sections to a Module by adding them to `self`.  This gives the ability to have dynamic documentation based on runtime configuration.  For example, in https://github.com/xobs/spibone/blob/master/spibone.py#L122 we examine the `wires` argument and attach the appropriate documentation to `self`.  This results in accurate documentation as seen at https://doctest.xobs.io/spibone.html, reflecting the correct protocol and number of wires.

### DIFF
--- a/litex/soc/integration/doc.py
+++ b/litex/soc/integration/doc.py
@@ -1,0 +1,150 @@
+# This file is Copyright (c) 2019 Sean Cross <sean@xobs.io>
+
+from migen.fhdl.module import DUID
+from migen.util.misc import xdir
+
+from litex.soc.interconnect.csr_eventmanager import EventManager
+
+import textwrap
+import inspect
+
+class ModuleDoc(DUID):
+    """Module Documentation Support
+
+    ModuleDoc enables you to add documentation to your Module.  This documentation is in addition to
+    any CSR-level documentation you may add to your module, if applicable.
+
+    There are two ways to use :obj:`ModuleDoc`:
+
+    1. Inherit :obj:`ModuleDoc` as part of your class.  The docstring of your class will become the
+    first section of your class' module documentation
+    2. Add a :obj:`ModuleDoc` object to your class and inherit from :obj:`AutoDoc`.
+
+    If you inherit from :obj:`ModuleDoc`, then there is no need to call ``__init__()``
+
+    Synopsis
+    --------
+
+    ::
+
+        class SomeClass(Module, ModuleDoc, AutoDoc):
+            \"\"\"Some Special Hardware Module
+
+            This is a hardware module that implements something really cool.
+            \"\"\"
+
+            def __init__(self):
+                self.other_section = ModuleDoc(title="Protocol Details", body="This section details more
+                information about the protocol")
+
+    """
+
+    def __init__(self, body=None, title=None, file=None, format="rst"):
+        """Construct a :obj:`ModuleDoc` object for use with :obj:`AutoDoc`
+
+        Arguments
+        ---------
+
+        body (:obj:`str`): Main body of the document.  If ``title`` is omitted, then the
+        title is taken as the first line of ``body``.
+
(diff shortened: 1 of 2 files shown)


xobs:
This also adds the ability to pull from external files (as requested by @bunnie), where one could include a module's documentation outside of the Python file.  Finally, it also includes the ability to set a format other than rst.  Presumably this would be "md", which is very popular on github and on the Internet in general.

enjoy-digital:
Thanks!

litex/soc/integration/doc.py
```diff
@@ -0,0 +1,150 @@
...
+        file (:obj:`str` Optional): It is possible to load the documentation from an external
+        file instead of specifying it inline.  This allows for the use of an external text
+        editor.  If a ``file`` is specified, then it will override the ``body`` argument.
+
+        format (:obj:`str` Optional): The text format.  Python prefers reStructured Text, so this
+        defaults to `rst`.  If specifying a `file`, then the suffix will be used instead of
+        `format`.  If you specify a format other than `rst`, you may need to install a converter.
+        """
+
+        import os
```
mithro: Shouldn't import inside a function.

litex/soc/integration/doc.py
```diff
@@ -0,0 +1,150 @@
...
+            This is a hardware module that implements something really cool.
+            \"\"\"
+
+            def __init__(self):
+                self.other_section = ModuleDoc(title="Protocol Details", body="This section details more
+                information about the protocol")
+
+    """
+
+    def __init__(self, body=None, title=None, file=None, format="rst"):
```
mithro: `setup.py` uses `content_type='text/markdown'` etc. See https://packaging.python.org/guides/making-a-pypi-friendly-readme/#including-your-readme-in-your-package-s-metadata

litex/soc/integration/doc.py
```diff
@@ -0,0 +1,150 @@
...
+        """Construct a :obj:`ModuleDoc` object for use with :obj:`AutoDoc`
+
+        Arguments
+        ---------
+
```
mithro: You could use https://docs.python.org/3/library/typing.html here.

litex/soc/integration/doc.py
xobs: Why not?  Is there a `Python Best Practices` I could read that has information about things like this?

litex/soc/integration/doc.py
xobs: The idea behind using `format=` is that it can be inferred from the file extension.  Is there a common way to go from file extensions to mimetypes?

litex/soc/integration/doc.py
xobs: Are you suggesting we add Python 3.5 types to all new code?  That sounds like a fine suggestion.

litex/soc/integration/doc.py
mithro: pylint / pep8 will normally complain about this.

https://www.python.org/dev/peps/pep-0008/#imports

litex/soc/integration/doc.py
mithro: There is https://docs.python.org/3/library/mimetypes.html

litex/soc/integration/doc.py
mithro: I think heading towards typing on new code is probably better then adding it in the documentation.

mithro: It'll end up in the same place in the documentation and can be checked via things like mypy (http://mypy-lang.org/).

PR #269, opened 2019-10-01
soc: cores: support sending custom bitstream to ICAP

mateusz-holenko:
This adds FIFO that can be used to send any
sequence of commands to the ICAP controller.

NOTE: This commit changes ICAP's CSRs breaking compatibilty with the older software.


enjoy-digital:
Thanks. I merged it but also would like to keep the old ICAP module, so renamed this new one to ICAPBitstream. I also did some simplifications, can you verify it's still working correctly?

mateusz-holenko:
Thanks for the merge! Our inital tests show that the simple scenario works fine, but uploading a bigger bitstream does not. We will investigate it further and propose fixes.

PR #271, opened 2019-10-06
RFC: optional '-nowidelut' flag for yosys synth_ecp5

gsomlo:
NOTE: this is a RFC, and probably should be re-spun as a more elegant, less
"hacky" modification to the ecp5versa trellis build harness. I think I'd like to gather
feedback and force-push this at least once more before it getting applied...
Thanks,
--Gabriel

Passing '-nowidelut' to yosys' synth_ecp5 command improves area utilization
to the point where a (linux variant) rocket-chip based design will fit on a
versa_ecp5 board. Usually '-nowidelut' incurs a timing penalty, but that is
then mitigated by using DSP inference (enabled by default from yosys commit
8474c5b3).

Off by default, this flag can be enabled by adding '--yosys-nowidelut=True'
to the litex/boards/targets/versa_ecp5.py command line.

### DIFF
--- a/litex/boards/targets/versa_ecp5.py
+++ b/litex/boards/targets/versa_ecp5.py
@@ -11,6 +11,8 @@ from migen.genlib.resetsync import AsyncResetSynchronizer
 
 from litex.boards.platforms import versa_ecp5
 
+from litex.build.lattice.trellis import yosys_args, yosys_argdict
+
 from litex.soc.cores.clock import *
 from litex.soc.integration.soc_sdram import *
 from litex.soc.integration.builder import *
@@ -133,6 +135,7 @@ def main():
         help='gateware toolchain to use, diamond (default) or  trellis')
     builder_args(parser)
     soc_sdram_args(parser)
+    yosys_args(parser)
     parser.add_argument("--sys-clk-freq", default=75e6,
                         help="system clock frequency (default=75MHz)")
     parser.add_argument("--with-ethernet", action="store_true",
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks. Maybe the `versa5g_args` should be renamed in `yosys_args` and moved to `build/lattice/trellis.py` and could just be imported in the trellis targets? This way we could avoid duplication if others targets need this.

gsomlo:
On Sun, Oct 06, 2019 at 03:09:29AM -0700, enjoy-digital wrote:
> Maybe the versa5g_args should be renamed in yosys_args and moved to
> build/lattice/trellis.py and could just be imported in the trellis
> targets?

OK, I force-pushed something along these lines (still only importing
it from targets/versa_ecp5.py for now -- we can add it to other
targets eligible for trellis later, once we decide we like it).

Speaking of which, should the "yosys_args" be renamed once more, to
e.g., "trellis_args", since maybe in the future we'd like to pass in
optional flags for nextpnr as well, not just yosys?

gsomlo:
> Speaking of which, should the "yosys_args" be renamed once more, to
e.g., "trellis_args", since maybe in the future we'd like to pass in
optional flags for nextpnr as well, not just yosys?

I guess we can always do that when we actually _do_ need to add our first optional nextpnr flag... :)

enjoy-digital:
Sorry, i wanted to answer but pushed the wrong button and merged... Before my first answer i was also wondering what was the best between `yosys_args` and `trellis_args` but maybe we can keep it like that and if we need to set specific nextpnr flags we'll also add a `nextpnr_args`. We'll also probably want to have that for iCE40, so `yosys_args` and `nextpnr_args` is maybe better.

PR #272, opened 2019-10-06
fix comments in icap.py

sergachev:
Remove a couple of mistypes.


enjoy-digital:
Thanks for catching this.

PR #274, opened 2019-10-08
builder: use the SoC's existing shadow base with get_csr_header()

gsomlo:
Both the SoC and get_csr_header() have independently set defaults
for the value of 'shadow_base'. If the SoC's value was modified,
ensure that get_csr_header() uses the modified value instead of
its own default.

Signed-off-by: Gabriel Somlo <somlo@cmu.edu>


enjoy-digital:
Thanks.

PR #276, opened 2019-10-09
cpu/rocket: swap main_mem and io regions

gsomlo:
The total size of RAM (main_mem) can be expected to vary significantly,
and often exceed the size needed for MMIO allocations by a large margin.

As such, place Rocket's MMIO (io regions) below 0x8000_0000, and start
the RAM (main_mem) at 0x8000_0000, with nothing above it to limit its
future growth.

Also, bump the pre-built Rocket verilog submodule to an updated version,
which also comes with matching changes to the way MMIO and RAM accesses
are mapped and routed to their respective AXI interfaces.


enjoy-digital:
Thanks, merged.

PR #277, opened 2019-10-10
[feature] Add SystemVerilog support for the Vivado builder

MartinCornil:
The actual Verilog version used by Vivado is 2001. This version doesn't support features like the `$bits` "system call" which has been added in Verilog 2005. To be able to use these nice features, Xilinx recommends to add these files as SystemVerilog files.
This PR adds the possibility to declare files as SystemVerilog files with `add_sources` and `add_source_dir` (new `language` parameter). The ".sv" file extension has also been added for inferred file type identification.

### DIFF
--- a/litex/build/generic_platform.py
+++ b/litex/build/generic_platform.py
@@ -334,7 +334,7 @@ class GenericPlatform:
         for f in filenames:
             self.add_source(os.path.join(path, f), language, library)
 
-    def add_source_dir(self, path, recursive=True, library=None):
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks! That's merged with minor modifications:
- https://github.com/enjoy-digital/litex/commit/f3ba078882e2516a2a4982b3d1c7cad9ad5a457e
- https://github.com/enjoy-digital/litex/commit/ed3c53d7cd2afd1fe18400064c252151cb237ac3

PR #283, opened 2019-10-19
bios: Increment address when writing to flash

kbeckmann:
When writing multiple words using the "fw" command from bios, the address it wrote to wasn't incremented so the same address was just written to multiple times.


enjoy-digital:
Merged, thanks!

PR #284, opened 2019-10-23
cpu/rocket, soc_sdram: Bypass WB bus for LiteDRAM <-> Rocket link

gsomlo:
Rocket provides its own L1 i/d cache, and two AXI master ports:
cached RAM accesses are routed over a dedicated 'mem_axi' port,
while uncached MMIO accesses are separately routed over another
dedicated port, 'mmio_axi'.

We bypass the Litex Wishbone bus when Rocket accesses the main
RAM, by connecting 'mem_axi' directly to LiteDRAM.

Remaining CPU models will continue to share the Litex Wishbone
bus for both RAM and MMIO (a.k.a. CSR) accesses.

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -88,10 +88,9 @@ class RocketRV64(CPU):
         self.mem_axi   =  mem_axi = axi.AXIInterface(data_width=64, address_width=32, id_width=4)
         self.mmio_axi  = mmio_axi = axi.AXIInterface(data_width=64, address_width=32, id_width=4)
 
-        self.mem_wb    =  mem_wb = wishbone.Interface(data_width=64, adr_width=29)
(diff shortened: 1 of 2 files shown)


gsomlo:
As it turns out, on the ecp5 versa the liteDRAM data width is 128bit, which does NOT match my 64bit mem_axi port on the Rocket chip, so I need to take this back to the drawing board...

daveshah1:
Worth noting the TrellisBoard has a 32-bit rather than 16-bit DDR3 side bus so I imagine the system side will be 2x wider too (I haven't actually checked)

gsomlo:
> Worth noting the TrellisBoard has a 32-bit rather than 16-bit DDR3 side bus so I imagine the system side will be 2x wider too (I haven't actually checked)

Before I re-open this I should probably figure out if I can't specify the mem_axi data width from the Rocket end of things. Be nice if we could have Rocket access the RAM directly in the largest chunk size supported by liteDRAM...

PR #285, opened 2019-10-24
[RFC] Linker region concept

mateusz-holenko:
I was recently working on supporting `icefun` platform in [LiteX-BuildEnv](https://github.com/timvideos/litex-buildenv) and [Renode](https://github.com/renode/renode).

The problem I stumbled upon was that `icefun` keeps both LiteX BIOS and the firmware in flash and executes the code directly from it. The platform is defined in a way that it generates 3 memory regions:
* [spiflash](https://github.com/timvideos/litex-buildenv/blob/master/targets/icefun/base.py#L88),
* [rom](https://github.com/timvideos/litex-buildenv/blob/master/targets/icefun/base.py#L88)
* [user_flash](https://github.com/timvideos/litex-buildenv/blob/master/targets/icefun/base.py#L93)
where `rom` and `user_flash` are the subregions of `spiflash`.

BTW: Shouldn't [this code](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc_core.py#L381-L382) prevent such configurations?

I believe that the problem is that `rom` is needed by [the linker script](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/linker.ld#L13) and `user_flash` is needed by [another one](https://github.com/timvideos/litex-buildenv/blob/master/firmware/linker-xip.ld#L13), but currently the only way of generating those symbols is by using `add_memory_region`.

This PR propose a new way of handling this problem. The idea behind "linker regions" is to allow defining memory subregions for the use of linker scripts without generating actual overlapping memory regions.

What do you think about it?

### DIFF
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -181,6 +181,7 @@ class SoCCore(Module):
 
         # Regions / Constants lists
         self._memory_regions = []  # (name, origin, length)
+        self._linker_regions = []  # (name, origin, length)
         self._csr_regions = []     # (name, origin, busword, csr_list/Memory)
         self._constants = []       # (name, value)
 
@@ -428,6 +429,9 @@ class SoCCore(Module):
             raise FinalizeError
         self._csr_masters.append(csrm)
 
+    def add_linker_region(self, name, origin, length):
+        self._linker_regions.append((name, origin, length))
+
     def add_memory_region(self, name, origin, length):
         def in_this_region(addr):
             return addr >= origin and addr < origin + length
@@ -448,6 +452,9 @@ class SoCCore(Module):
 
     def get_memory_regions(self):
         return self._memory_regions
+
+    def get_linker_regions(self):
+        return self._linker_regions
 
     def check_csr_range(self, name, addr):
         if addr >= 1<<(self.csr_address_width+2):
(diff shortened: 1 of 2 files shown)


litex/soc/integration/soc_core.py
```diff
@@ -428,6 +429,9 @@ def add_csr_master(self, csrm):
             raise FinalizeError
         self._csr_masters.append(csrm)
 
+    def add_linker_region(self, name, origin, length):
+        self._linker_regions.append((name, origin, length))
```
mateusz-holenko: This should probably verify it the new `linker region` is included in any defined `memory region`.

enjoy-digital:
@mateusz-holenko: can you check https://github.com/enjoy-digital/litex/commit/9fcf29738710f755ed45ad967eacd2287cd91a67? The memory region overlap check should be fixed. (your code should no longer be able to compile) and i added a `linker_region` parameter to `add_memory_region` to specify that the memory region is for the linker (we don't do the overlap check for linker regions). So in your code, you should only have to add `linker_region=True` to the memory regions that were declared just for the linker.

I'll add a check soon to be sure that a memory declared as `linker_region` is part of an already declared memory.

mateusz-holenko:
The idea with `linker_region` parameter looks nice.

My only question is if the information about the type of memory region (if it's a linker region or not) is propagated to the configuration output (i.e., csv/json)?
This is crucial for me to be able to distinguish between those in the configuration file.

enjoy-digital:
@mateusz-holenko: All your requirements should now be covered so i think we can close this.

mateusz-holenko:
You are right. Thanks for the support @enjoy-digital!

PR #286, opened 2019-10-24
build/lattice/trellis: optionally allow failure if p&r timing not met

gsomlo:
When timing requirements are strict, allow the build process to fail upon
failure to meet timing. This facilitates running the build process from a
loop, repeatedly, until a "lucky" p&r solution is found, e.g.:

  while true; do
    litex/boards/targets/versa_ecp5.py --gateware-toolchain trellis \
      --sys-clk-freq=60e06 --cpu-type rocket --cpu-variant linux \
      --with-ethernet --yosys-nowidelut \
      --nextpnr-timingstrict
    if [ "$?" == "0" ]; then
      echo "Success" | mail -s "Build Succeeded" your@email.here
      break
    fi
  done

This augments commit #683e0668, which unconditionally forced p&r to
succeed, regardless of whether timing was met, via '--timing-allow-fail'.


enjoy-digital:
Thanks merged. In the future, it some yosys/nextpnr parameters are common for CE40/ECP5 (and others), we'll maybe regroup this in a common file (but parameter names should not change so that's fine for now).

Issue #299, opened 2019-11-04
Rocket: 'mem_axi'  <--> LiteDRAM data-width impact on performance


gsomlo:
Assuming a dedicated Rocket mem_axi <-> LiteDRAM data port link, and a
direct, "native" point-to-point AXI connection between the two, will an
increased AXI data_width improve performance?

Given that Rocket transfers entire cache lines over 'mem_axi', would
shorter, wider bursts actually improve performance as compared to longer,
narrower ones?

I built Rocket Litex/Linux variants using "WithMedCores", with 'mem_axi'
data width set to three different values: standard (64-bit), wide (128-bit)
and double-wide (256-bit). By picking subsets of ('dm', 'dq', 'dqs_p') pins
in litex-boards/litex_boards/partner/platforms/trellisboard.py, I forced
the LiteDRAM data port to have a data width matching the 'axi_mem' data
width of the Rocket version under test, thus enabling native point-to-point
AXI linkage between Rocket and LiteDRAM.

I built a fourth bitstream, where the data width gap difference between the
default 'mem_axi' (64bit) and default trellisboard LiteDRAM port (256bit)
was bridged using the Wishbone data width converter. This is presumably
inefficient, and a native AXI converter might do a (much?) better job
(to be determined by some benchmarking upon implementation). However, I
wanted to throw it in as an additional data point for comparison.

All CPUs are otherwise identical, except for the axi data width (or for
the presence of a Wishbone-based width converter). The system clock runs
at 65MHz in all four cases.

I then loaded a BBL-wrapped Linux kernel with an embedded, Busybox based
initrd. The BBL-embedded DT is configured to tell Linux that 128MB of RAM
are available, which works because it never exceeds the actual physical
memory exposed by the trellisboard LiteDRAM (minimum 256MB at 64-bit
width, maximum 1GB at 256-bit width).

On top of this setup, I ran Coremark, Linpack (single and double precision),
and NBench (integer and fp performance as a fraction of a 90MHz Pentium,
and memory, integer, and fp performance as a fraction of a 233MHz AMD K6).

NOTE1: All CPU variants tested use bbl-based FPU emulation, so fp benchmark
results are therefore simply an indirect measure of *integer* performance,
through the execution of BBL's fp emulation trap handler!

NOTE2: Rocket's 'WithMedCores' is designed to fit on the ecp5versa, and has
a relatively small L1 cache. Performance would almost certainly be improved
[...]

PR #300, opened 2019-11-04
RFC: Direct link between Rocket/mem_axi <--> LiteDRAM dataport

gsomlo:
Attn @enjoy-digital -- do not (immediately) pull into upstream LiteX, let's have a conversation and review first -- maybe it isn't even the right thing to do :) Please also see the related experiment regarding data width at https://github.com/enjoy-digital/litex/issues/299

Use a direct AXI point-to-point link between rocket's cached-ram axi port (mem_axi) and LiteDRAM, whenever the data width of the two ports matches.

For now, as a workaround, use Wishbone based data width conversion when the port widths are mismatched (which is rather inefficient). A future improvement might be to implement a native axi data-width converter.

Additionally, this series prepares Rocket's wrapper file to allow future variants with different, parameterized axi_mem data widths.

Finally, remove the limitation on the maximum size of usable ram (to allow full utilization of the 1GB of RAM present on the trellis board)

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -88,10 +88,9 @@ class RocketRV64(CPU):
         self.mem_axi   =  mem_axi = axi.AXIInterface(data_width=64, address_width=32, id_width=4)
         self.mmio_axi  = mmio_axi = axi.AXIInterface(data_width=64, address_width=32, id_width=4)
 
-        self.mem_wb    =  mem_wb = wishbone.Interface(data_width=64, adr_width=29)
         self.mmio_wb   = mmio_wb = wishbone.Interface(data_width=64, adr_width=29)
 
-        self.buses     = [mem_wb, mmio_wb]
+        self.buses     = [mmio_wb]
 
         # # #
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
@gsomlo: i think this is a good compromise for now and we could merge it. In the future, i'd like to 

1. implement support for bi-directional mode on  LiteDRAMNativePortConverter (https://github.com/enjoy-digital/litedram/blob/master/litedram/frontend/adaptation.py#L312) and/or
2. add AXI converters.

Implementing 1. is probably easier as a first step (the LiteDRAM Native interface is a simplified AXI) and would allow requesting a LiteDRAM port of specific data-width in SoCSDRAM with the converter directly integrated in LiteDRAM.

gsomlo:
@enjoy-digital: thanks for the feedback, and please go ahead and merge this if you think it's OK. As it is now, the 64-bit linux rocket variant will work "natively" over AXI on the nexys4ddr, which exposes a 64-bit wide liteDRAM port.

I'm currently looking into how I could most "eloquently" parameterize the mem_axi port's data width on the Rocket/Chisel side of things (https://github.com/chipsalliance/rocket-chip/issues/2185).

I'll generate a follow-up PR that actually adds more variants (with different mem_axi width options) to rocket-litex-verilog once I'm happy with how to do that (and also maybe once they bump firrtl to _really_ solve https://github.com/chipsalliance/rocket-chip/issues/2168 (they fixed firrtl upstream, but didn't bump the Rocket submodule yet, afaict.

Issue #289, opened 2019-10-28
Make CSR Memory paged access behavior better visible to developers


sergachev:
I spent some time figuring out why one Memory instance was correctly writable only up to 512 bytes and catching very random bugs appearing on writes above 512 B, until I figured out that there is a 'magic' 512 constant here https://github.com/enjoy-digital/litex/blob/ab8af28213c5d744e6749b76452f5053feadcf58/litex/soc/interconnect/csr_bus.py#L89 and that the generated csr.h contains something like xxx_page_write().

Ideas coming to my mind to make this behavior better visible:
 - define page size along with XXX_MEM_BASE (where does 512 B limit actually come from? can we change it?)
 - update https://github.com/timvideos/litex-buildenv/wiki/LiteX-for-Hardware-Engineers#memory
 - generate accessor functions in csr.h which set page number based on address 
 - produce some warnings if memory size is > 512

I do not mind implementing and submitting some of that, just want to get ideas what could be best.

enjoy-digital:
Thanks for reporting, i'll have closer look at it in the next days to see what is the best.

enjoy-digital:
@sergachev: the actual limitation of 512 bytes is coming from the fact that the CSR region size is 0x800 (4096) and default csr_data_width is 8, so this means the maximum size is 512 bytes. With https://github.com/enjoy-digital/litex/commit/9c3c43c94ade654a1e386ff6dd131195dd6d63d7 i added a check for it. For now this is limitation is not easy to remove, so i would suggest just having this check. In the future we'll see if it's interesting to remove this limitation. Is it fine for you?

sergachev:
@enjoy-digital Thank you for the explanation. 9c3c43c94ade654a1e386ff6dd131195dd6d63d7 prohibits creating memories > 512, which is not ideal - they can be used, just with paged access. I would rather make it a warning.

enjoy-digital:
@sergachev: can you create a pull request with the behaviour you think should be the best? I'm not using csr.SRAM heavily and would probably recommend using `wishbone.SRAM` for large memories.

sergachev:
I do not use csr.SRAM directly, rather wishbone.SRAM indeed, following this guide: [Wishbone ro, logic wo](https://github.com/timvideos/litex-buildenv/wiki/LiteX-for-Hardware-Engineers#wishbone-ro-logic-wo)

```
class RxMem(Module, AutoCSR):
    ...
    def __init__(self):
        ...
[...]

PR #307, opened 2019-11-15
change >512 B CSR memory exception to a warning

sergachev:
So regarding https://github.com/enjoy-digital/litex/issues/289 I suggest, at least for now, to demote the exception to a warning, because actually > 512 B memories work and are handled by the code just below this exception.


enjoy-digital:
Thanks, merged.

PR #308, opened 2019-11-17
soc_sdram, bios/sdram: support sdram init for csr_data_width <= 32

gsomlo:
Read leveling tested with csr_data_width [8, 16, 32] on the
ecp5-versa5g and trellisboard (using yosys/trellis/nextpnr),
and on the nexys4ddr (using Vivado).

FIXME:
    - Not tested sdrrdbuf(), sdrrderr(), and sdrwr();
    - Still need to update write_level() routine!!!

### DIFF
--- a/litex/soc/integration/soc_sdram.py
+++ b/litex/soc/integration/soc_sdram.py
@@ -29,8 +29,8 @@ class SoCSDRAM(SoCCore):
     def __init__(self, platform, clk_freq, l2_size=8192, **kwargs):
         SoCCore.__init__(self, platform, clk_freq, **kwargs)
         if not self.integrated_main_ram_size:
-            if self.cpu_type is not None and self.csr_data_width != 8:
-                 raise NotImplementedError("BIOS supports SDRAM initialization only for csr_data_width=8")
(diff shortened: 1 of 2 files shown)


enjoy-digital:
@gsomlo: thanks! i'll review the changes and will do the write leveling adaptations.

enjoy-digital:
@gsomlo: i just did some tests and it seems indeed to work fine. I'll try to do the write-leveling adaptations in the next days so that we can merge.

gsomlo:
I just took a stab at write leveling, but I'm flying blindly here
since I don't have anything to test it on. I'm thinking it should
work though, so please give it a test.

Happy to back it out if it turns out to be wrong :)

enjoy-digital:
Thanks, i've been able to test it on Kintex7 and the write leveling is done correctly with csr_data_width=8 or 32. We can merge the changes. For info, your changes are useful for your usecase but also very useful for others, like creating a SoC with LitePCIe (which assumes csr_data_width=32) and LiteDRAM (for which for now the initialization was only working with csr_data_width=8), so thanks!

PR #309, opened 2019-11-20
soc/cores/clock: add lock reg and assign reset

mateusz-holenko:
It was necessary to add drp_locked CSR for reading LOCK signal from
MMCM. Additionally, input signal RESET from MMCM was not driven by
any signal to do a proper reset of MMCM module thus it was impossible
to perform entirely correct dynamic clock reconfiguration.

### DIFF
--- a/litex/soc/cores/clock.py
+++ b/litex/soc/cores/clock.py
@@ -92,6 +92,7 @@ class XilinxClocking(Module, AutoCSR):
 
     def expose_drp(self):
         self.drp_reset = CSR()
+        self.drp_locked = CSR()
         self.drp_read  = CSR()
         self.drp_write = CSR()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks it's merged. 

For the `reset,` i was aware of it and remember having a strange behaviour with it, but i just re-tested if and it seems to work fine, so i applied it to all Xilinx devices with: https://github.com/enjoy-digital/litex/commit/36107cdfd79752289b24fb6200a90094d78ad597

For the `drp_locked` register, i changed it to a CSR Status and also connected it to the locked signal with: https://github.com/enjoy-digital/litex/commit/6883a43680c4619dd02e89b1d74a6133212fe31f

PR #310, opened 2019-11-25
spi_flash: correct documentation on SPI mode

xobs:
The SPI mode is actually mode3, since the output value is updated on the
falling edge of CLK and the input value is updated on the rising edge.

This also clarifies some of the documentation based on experience with
the core.


enjoy-digital:
Thanks!

PR #312, opened 2019-11-30
new VGA terminal implementation

Frank-Buss:


mithro:
I'm wondering if this should be part of [litevideo](https://github.com/enjoy-digital/litevideo)?

enjoy-digital:
@mithro, @Frank-Buss: indeed, LiteVideo would probably be a better location for this. Otherwise the code/example is really nice, thanks! @Frank-Buss : If you do a pull request to LiteVideo, you can put the terminal directory near csc/input/ouput,... and can you reduce the size of the pictures ?(we could probably reduce each down to 100KB instead of 1MB).

Frank-Buss:
Ok.

PR #313, opened 2019-12-03
Yosys - ISE flow fix

mmicko:
These changes are needed due to recent better support introduced in Yosys specially regarding Spartan6 flow.

Tested this with https://github.com/mmicko/linux-on-litex-vexriscv/tree/spartan6 for creating Linux running on WaxWing Spartan6 board from Numato

### DIFF
--- a/litex/build/xilinx/ise.py
+++ b/litex/build/xilinx/ise.py
@@ -85,9 +85,17 @@ def _run_yosys(device, sources, vincpaths, build_name):
     for filename, language, library in sources:
         ys_contents += "read_{}{} {}\n".format(language, incflags, filename)
 
+    family = ""
+    if (device.startswith("xc7") or device.startswith("xa7") or device.startswith("xq7")):
+        family = "xc7"
+    elif (device.startswith("xc6s") or device.startswith("xa6s") or device.startswith("xq6s")):
+        family = "xc6s"
+    else:
+        raise OSError("Unsupported device")
+
     ys_contents += """hierarchy -top top
-synth_xilinx -top top -family xc6s -ise
-write_edif -pvector bra {build_name}.edif""".format(build_name=build_name)
+synth_xilinx -top top -family {family} -ise
+write_edif -pvector bra {build_name}.edif""".format(build_name=build_name, family=family)
 
     ys_name = build_name + ".ys"
     tools.write_to_file(ys_name, ys_contents)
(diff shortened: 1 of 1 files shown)


litex/build/xilinx/ise.py
```diff
@@ -207,7 +218,8 @@ def build(self, platform, fragment, build_dir="build", build_name="top",
                     _build_xst_files(platform.device, platform.sources, platform.verilog_include_paths, build_name, self.xst_opt)
                     isemode = mode
                 else:
-                    _run_yosys(platform.device, platform.sources, platform.verilog_include_paths, build_name)
+                    if run:
```
mithro: Where does `run` come from?

mithro: Oh wait, it is an argument....

litex/build/xilinx/ise.py
```diff
@@ -114,7 +122,10 @@ def _run_ise(build_name, ise_path, source, mode, ngdbuild_opt,
         settings = common.settings(ise_path, ver, "ISE_DS")
         build_script_contents += source_cmd + tools.cygpath(settings) + "\n"
     if mode == "edif":
-        ext = "edif"
+        ext = "ngo"
```
mithro: Why is this needed now? I'm a little worried it will break other edif use cases?

litex/build/xilinx/ise.py
mmicko: Right. This is added since it executes yosys right now even when you do not build but load on board

litex/build/xilinx/ise.py
mmicko: ngo and ngc are supported by ngdbuild and other commands in flow. Adding edif2ngd is cleaner way to do the flow instead of using edif directly (Think for actually did not work fine at all for me , but away right now from developer machine, so can confirm tomorrow)

mmicko: Just tested. Was right, this was broken before. Without change it fails at this line :

netgen -ofmt verilog -w -sim top.edif top_synth.v

Release 14.7 - netgen P.20131013 (lin64)
Copyright (c) 1995-2013 Xilinx, Inc.  All rights reserved.
ERROR:NetListWriters:286 - Invalid file argument 'top.edif'.

Thing is netgen works with ngc or ngo, and edif was sent, that is why I add step to convert edif to ngc and made this work actually.

mithro: Great!

litex/build/xilinx/ise.py
mithro: Great!

enjoy-digital:
Thanks @mmicko and @mithro.

Issue #314, opened 2019-12-10
We need to document LiteX CSRs!


gsomlo:
I'd like to add a doc/CSR.md file to LiteX, documenting how CSRs work from both the hardware and software (programmer's API) perspective. I'm going to add a new comment below with a .md-ish formatted document, which we can collectively work on until it's more or less ready, after which I'll post a PR to add it to LiteX. If there's a better way to edit a doc before including it under ./doc, please let me know and I can start doing that (I'm a bit awkward with github, having been used to linux's mailing list based model of development and feedback solicitation, so apologies for any clunkiness :).
Comments, feedback, and direct edits of the following comment (if github allows it) much appreciated!

gsomlo:
1. LiteX CSR definition and terminology

LiteX CSRs are MMIO Configuration/Status Registers used to interact with
and control various hardware peripheral devices included on a LiteX SoC.
NOTE: LiteX CSRs are *NOT* to be confused with the concept of CSRs as
defined within the RISC-V ISA! To further complicate things, it is
entirely possible for the LiteX SoC to be configured with a RISC-V soft
CPU, in which case the term "CSR" will be used in a RISC-V specific way
when discussing the CPU's behavior and internals, and in a LiteX specific
way (see below) when discussing "hardware peripheral MMIO registers" that
belong to the SoC, and are accessed by the CPU at the direction of various
pieces of system software (e.g., the Bios, bootloader, or OS kernel).

2. LiteX CSR Addressing

2.1. CSR Address Space: Base Address and `csr_address_width`

Currently, LiteX CSRs are placed in an address segment of size 0x1000000
bytes (16MB, see `litex/soc/integration/soc_core.py:SoCCore:__init__()`
-- search for `self.register_mem("csr", ...)`).

This segment starts at a Base Address provided as `mem_map["csr"]`, also
in SocCore, where it defaults to 0x82000000. The Base Address value may
be overridden by other components (typically CPUs) as they are added to
the SoC configuration.

Assuming an address space of 32 bits, the LiteX CSR Base Address occupies
the 8 MSBits, leaving the remaining 24 bits for addressing actual MMIO
registers within the CSR segment. However, assuming that the smallest
addressable individual memory location is 32-bit wide, that also removes
the 2 LSBits, leaving the actual number of bits for addressing individual
CSR elements (`csr_address_width`) to be max. 22 (enforced via an
`assert 2**(csr_address_width + 2) <= 0x1000000` in `SocCore:__init__()`).
[...]

PR #315, opened 2019-12-12
soc_core: additional CSR safety assertions

gsomlo:
Since csr_data_width=64 has probably never worked properly, remove
it as one of the possible options (to be fixed and re-added later).
Add csr_data_width=16, which has been tested and does work.

Additionally, ensure csr_data_width <= csr_alignment (prevent subregister slices
wider than the CPU's native word size or XLen).

Note: somewhat related to Issue #314


enjoy-digital:
Thanks!

PR #320, opened 2019-12-21
Misc. Rocket and CSR cleanup

gsomlo:
this is (hopefully) non-controversial stuff I'd like to apply before posting a CSR accessor update RFC


enjoy-digital:
Thanks.

PR #321, opened 2019-12-21
cpu/rocket: variants with double (128b) and quad (256b) wide mem_axi

gsomlo:
Various development boards' native LiteDRAM ports may have
native data widths of either 64 (nexys4ddr), 128 (ecp5versa),
or 256 (trellisboard) bits. Add Rocket variants configured
with mem_axi ports of matching data widths, so that a point
to point connection between the CPU's memory port and LiteDRAM
can be accomplished without any additional data width conversion
gateware.


enjoy-digital:
Thanks!

PR #324, opened 2019-12-23
RFC: Updated CSR accessors

gsomlo:
@enjoy-digital Please review, but don't apply until the intra-subregister endianness issue has been elucidated :)
Once we know more about that, I'd also like us to consider the possibility of adding memcpy-_**like**_ methods for CSRs that are wider than 64 bits, such as `sdram_dfii_pix_rddata_addr` and `sdram_dfii_pix_wrdata_addr`, and `CSR_IDENTIFIER_MEM_BASE`, so that we can get rid of `MMPTR` altogether.

I'm hoping that ultimately this could serve as an "OS/kernel programmer's reference", so we can implement a flexible set of accessors in e.g. Linux, with csr_data_width passed in as a DTB parameter. (see https://github.com/litex-hub/linux-on-litex-vexriscv/pull/60).

### DIFF
--- a/litex/soc/software/include/hw/common.h
+++ b/litex/soc/software/include/hw/common.h
@@ -4,8 +4,8 @@
 #include <stdint.h>
 
 /* To overwrite CSR accessors, define extern, non-inlined versions
- * of csr_read[bwl]() and csr_write[bwl](), and define
- * CSR_ACCESSORS_DEFINED.
+ * of csr_rd_uint[8|16|32|64]() and csr_wr_uint[8|16|32|64](), and
+ * define CSR_ACCESSORS_DEFINED.
  */
 
 #ifndef CSR_ACCESSORS_DEFINED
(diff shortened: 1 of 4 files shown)


gsomlo:
@enjoy-digital @mateusz-holenko : Just added a method to read a (large, >64bit) CSR as an array of uint[8|16|32|64] values, that should return the same result independently of the value of csr_data_width. Now working on a corresponding write method, at which time I'm hoping the software-side CSR overhaul will be done and useful to implement a truly generic Linux SoC driver :)

Any comments and feedback (from anyone reading this, really) much appreciated!

gsomlo:
ok, we now have both "classic" accessors (read/write standard C uintX_t types), as well as arrays of uintX_t superimposed over a CSR (should be useful for consistently manipulating CSRs that total over 64 bits and therefore can't be read/written with one of the "classic" accessors).

I'm giving this another few days to "settle", after which I'll remove the "testing" commit and the "RFC" status, and ask @enjoy-digital to apply :)

Thanks much, and have a good weekend!

mithro:
Can you post this to the litex-linux mailing list? See https://groups.google.com/forum/?nomobile=true#!forum/linux-litex

litex/soc/software/include/hw/common.h
```diff
@@ -14,37 +16,255 @@
...
+			buf[i] = _csr_rd(addr, sizeof(uintX_t)); \
+			addr += num_subregs(sizeof(uintX_t)); \
+		} \
+	} else { \
+		/* multiple elements per subregister (2, 4, or 8) */ \
+		nsubs = num_subregs(sizeof(uintX_t) * cnt); \
+		n_sub_elem = CSR_DW_BYTES / sizeof(uintX_t); \
+		for (i = 0; i < nsubs; i++) { \
+			r = _bswp_dyn_X(addr[i], CSR_DW_BYTES); \
+			for (j = 0; j < n_sub_elem; j++) { \
```
gsomlo: I think if I go backwards `for (j = n_sub_elem - 1; j >= 0; j--)` I might be able to forego having to use byteswap routines at all, will try that and force-push an updated patch if successful...

gsomlo:
latest force-push removes need for byte swapping, so I'm optimistic about this getting close to something I'd like to see upstream :)

enjoy-digital:
@gsomlo: thanks for the hard work on this!

gsomlo:
Resubmitted as PR #339 , without the testing patch applied to the bios main function. I wanted that to still be available publicly, so I'm leaving it here but closing this PR.

mithro:
@gsomlo 
 - Would you mind adding Python equivalents somewhere in LiteX? They would be useful when using wishbone bridge?
 - Could you also add a PR too https://github.com/litex-hub/wishbone-utils ? I'm sure @xobs would be interested in this functionality in wishbone-tool

gsomlo:
Hi @mithro:

>   • Would you mind adding Python equivalents somewhere in LiteX? They would be
>     useful when using wishbone bridge?

Not sure what you mean here. I was thinking of updating
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/identifier.py#L15
(and the corresponding software side routine at
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/id.c#L14
) in a follow-up PR to use the new byte-array CSR accessor, so that'd be adding at least some of the Python equivalent functionality. But not sure what/where
else you're thinking of.

>   • Could you also add a PR too https://github.com/litex-hub/wishbone-utils ?
>     I'm sure @xobs would be interested in this functionality in wishbone-tool

I'm not familiar with that code base, and after a quick cursory look I
can't tell which files I'd be patching. I also don't know how I'd test
any changes I'd be proposing, so this would probably be better done if
someone familiar with `wishbone-utils` cut'n'pasted (some of) the new
LiteX accessors into the appropriate place in the code base...

PR #326, opened 2019-12-25
WIP: Add support for ICE40 HX8K Breakout Board

fallen:
This is just a stash of the changes I did to integrate the ICE40 HX8K BB support from litex-buildenv to litex repo.

I had to do this in order to use the board in an out-of-tree Litex target file.
I had to do this since litex-buildenv is not a Python package.
After discussing with @mithro he told me that board support files are supposed to end up being integrated in Litex repo.

Maybe this can be used as a basis to iterate on.


fallen:
After re-reading the tweet, it's the wrong repo, IIUC this is better to be integrated in https://github.com/litex-hub/litex-boards.
Let's close this then?

mithro:
@fallen Yes please.

PR #327, opened 2019-12-30
moving RAM offsets outside of CSR_ETHMAC define

zakgi:
the macro for RAM offsets are within the CSR_ETHMAC #ifdef blocked, which is trimmed when building the firmware for boards without ethernet.
This change or something similar would allow to build on such boards.

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -232,6 +232,19 @@ int serialboot(void)
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Hi @zakgi,

theses RAM offsets are only used by netboot (at least in LiteX), so with Ethernet support, so it should be fine. Can you tell me what you were building and which errors you were having?

zakgi:
Hi @enjoy-digital,

the offsets are also used in the flashboot() function farther down in boot.c:414-.
Building a configuration with flash but without ethernet results in an error.

enjoy-digital:
Indeed, thanks!

PR #328, opened 2020-01-01
add the possibility for a "precise" clock solution

bunnie:
If clocks and multipliers are planned well, we can have
a zero-error solution for clocks. Suggest to change < to <= in
margin comparison loop, so that a "perfect" solution is allowed
to converge.


enjoy-digital:
That's true, thanks for the feedback/PR!

PR #330, opened 2020-01-02
Document CTRL and fix TIMER0 Documentation

xobs:
This fixes documentation of `TIMER0` based on the output I've seen at https://rm.fomu.im/timer0.html.  With this patch, the lists format correctly: 
image

It also adds some documentation to `CTRL`.  I'm not sure what the fields are used for, so I took a guess.  `SCRATCH` appears largely unused, and nothing appears to actually use `RESET`.  If these descriptions are incorrect, I can change them to something that is correct.

### DIFF
--- a/litex/soc/cores/timer.py
+++ b/litex/soc/cores/timer.py
@@ -18,31 +18,34 @@ class Timer(Module, AutoCSR, ModuleDoc):
     Provides a generic Timer core.
 
     The Timer is implemented as a countdown timer that can be used in various modes:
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks! It's merged, i just rephrased a bit the SoCController CSR descriptions with: https://github.com/enjoy-digital/litex/commit/b65a36e7e81ca7a7d7ef205010bea2d608ebe3e3

xobs:
Ah, so that's what it's used for!  Good to know.  Thanks for the fix!

PR #331, opened 2020-01-03
WIP: add support for DRP on XADC

bunnie:
The design is backward-compatible in functionality for users
who don't want to use DRP. That is, on power on, the XADC will
scan the supply and temperature and store them in CSRs.

If drp_enable is set, the scanning stops, and the XADC is now
controlled by the DRP bus.

Wher drp_enable is reset, the XADC may return to an auto-sample
mode, but only if the internal registers are configured to do this.
If you return to drp_enable without, for example, turning on
the continuous sequence and setting which channels to check,
the results will be unpredictable (mostly either it'll scan just
once and stop, or it'll not scan all the channels, depending on
the register settings).

At this point, the backward compatibility was confirmed in testing,
the DRP API is still a work in progress as the application this
is being developed for needs to support fun stuff like real time
sampling of signals to a buffer.

Down the road, this block may have to be modified again to support an
output FIFO, so we're not railing the CPU trying to do real time
sampling of ADC data. This will probably be added as a True/False flag
of some sort in the parameter list, because the FIFO will be expensive
as far as BRAM goes to implement and applications that don't need the
FIFO buffer can probably use that BRAM for better things.

### DIFF
--- a/litex/soc/cores/xadc.py
+++ b/litex/soc/cores/xadc.py
@@ -9,6 +9,8 @@ from litex.soc.interconnect.csr import *
 
 # XADC ---------------------------------------------------------------------------------------------
 
+analog_layout = [("vauxp", 16), ("vauxn", 16), ("vp", 1), ("vn", 1)]
+
 class XADC(Module, AutoCSR):
     def __init__(self, analog_pads=None):
         # Temperature(°C) = adc_value*503.975/4096 - 273.15
@@ -26,12 +28,6 @@ class XADC(Module, AutoCSR):
         # Alarms
         self.alarm = Signal(8)
         self.ot    = Signal()
-
-        # Analog
-        if analog_pads == None:
-            self.analog = Record([("vauxp", 16), ("vauxn", 16), ("vp", 1), ("vn", 1)])
-        else:
-            self.analog = analog_pads
 
         # # #
 
@@ -65,10 +61,10 @@ class XADC(Module, AutoCSR):
             o_CHANNEL   = channel,
             o_EOC       = eoc,
             o_EOS       = eos,
-            i_VAUXN     = self.analog.vauxn,
-            i_VAUXP     = self.analog.vauxp,
-            i_VN        = self.analog.vn,
-            i_VP        = self.analog.vp,
+            i_VAUXP     = 0 if analog_pads is None else analog_pads.vauxp,
+            i_VAUXN     = 0 if analog_pads is None else analog_pads.vauxn,
+            i_VP        = 0 if analog_pads is None else analog_pads.vp,
+            i_VN        = 0 if analog_pads is None else analog_pads.vn,
             i_CONVST    = 0,
             i_CONVSTCLK = 0,
             i_RESET     = ResetSignal(),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the PR! Can you give me permissions to push change to the PR to ease work on this?:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

This way i could just commit small changes directly and we could continue discussing here and merge when ready.

bunnie:
I /think/ this access has been enabled...that checkbox for "allow edits from maintainers" should do it, right?

image

enjoy-digital:
Yes it should be fine, thanks.

bunnie:
OK, I've added the patch you sent back. Also, I saw I could add you as a maintainer to the litex branch inside betruted, so I did that. hopefully that gives you commit access now.

enjoy-digital:
OK thanks, it should be good for the permission now. So now we need to see if it's still behaving as you expect and see if all the features are there. Once done, we could merge it.

bunnie:
For the "P" side of the analog channels, actually, connecting
a digital line to them has "no meaning". The docs say that
either you connect an analog pin to a pad, or vivado "ties it off
appropriately". I wish it were the case that tying a pin to 0 or 1
would actually connect it to a power or ground, because it means
that even in unipolar mode you have to burn two pins to break out
the signal of interest *and* the ground reference analog pad
(I thought I could just connect it to "0" and the pin would be
grounded, but that doesn't happen -- it's just ignored if it's
not wired to a pad).

For the pad specifier, is it OK to leave it with an optional
argument of analog_pads=None? I tried assigning to the
self.analog property after instantiation, but this doesn't
seem to work, the default values are preferred. It looks like
if you don't want to do the analog_pads= optional argument
the other way to do it would be to add code on the instiating
module that tampers with the properties of the instance directly,
but I think that's sort of ugly.

Also, I noticed you stripped out the layout specifier for
the analog_pads. I thought it would be nice to provide that
in the file, so the caller doesn't have to infer what the
pad layout is by reading the code...what's the motivation for
removing that?

enjoy-digital:
Sorry, i tried to simplified things but i missed the fact analog parameter were some pads. We can probably keep what you've just done and revert the layout in the top yes. Can you point me to some code using the new features of the XADC? (I just want to be sure to understand). I'll do the changes to revert the layout.

bunnie:
Sure, I just pushed my code as it stands:

XADC instantiation goes through the info block:
https://github.com/betrusted-io/betrusted-soc/blob/master/gateware/info/__init__.py#L24

But the pads are originated in the top level design file:
https://github.com/betrusted-io/betrusted-soc/blob/master/betrusted-soc.py#L462

Maybe there is a better way to handle the pad mapping, but I'm still figuring out Records.

enjoy-digital:
OK, thanks. Can you check https://github.com/enjoy-digital/litex/pull/331/commits/378722a7efe8c539466da0618adbc4215882d5c7 is fine for you?

bunnie:
yes, this works, confirmed in hardware. thank you!

enjoy-digital:
OK good. Do you want to add others features in this PR? Or should we merge and you'll create another PR with the extra-features later?

bunnie:
i'd say let's do the merge, and i'll open a new PR later when i get to adding a FIFO option to the XADC. 

At the moment, I've got what I need to find the bugs in the PCB design, so I think I'm going to push forward with more PCB-level bug squashing before i dive into adding features...

enjoy-digital:
OK, thanks. I'm merging then :)

PR #337, opened 2020-01-09
soc/cores/spi_flash: Don't tristate DQ2/DQ3 when bitbanging

gregdavill:
This change keeps DQ2/DQ3 as outputs when bit-banging is enabled.

Current behaviour when (bitbang_en=1 & dir=1), all the DQ pins get tristated. However bit-banging currently only support SPI modes.

This can cause issues for unconfigured flash chips, which treat DQ2 as ~WP, DQ3 as ~HOLD. 
This results in no output or bad output when attempting to read any data from the FLASH in bitbang mode, specifically: JEDEC ID codes, and programming status bits.

### DIFF
--- a/litex/soc/cores/spi_flash.py
+++ b/litex/soc/cores/spi_flash.py
@@ -8,6 +8,7 @@
 
 from migen import *
 from migen.genlib.misc import timeline
+from migen.fhdl.specials import Tristate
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the clear explanation and fix!

PR #338, opened 2020-01-10
Add optional 'ignore-loops' flag to nextpnr

DurandA:
This PR adds optional passing of `--ignore-loops` flag to NextPNR for the Trellis toolchain.


enjoy-digital:
Thanks!

PR #349, opened 2020-01-19
Timing improvement

bunnie:
Hard IP blocks are fixed in location, so long/deep combinational paths routing to multiple hard IP blocks can lead to timing closure problems. 

XADC and MMCM DRPs currently have their DEN pins triggered by the ".re" output of a CSR. This is asynchronously derived from a fairly complicated set of logic that involves a logic path that goes all the way back through the cache and arbitration mechanisms of the wishbone bus. On more complex designs, this is leading to a failure of timing closure for these paths, because the hard IP blocks can be located in disparate portions of the chip which "pulls" the logic cluster in opposite directions in an attempt to absorb the routing delays to these IP blocks, leading to non-optimal placement for everything else and thus timing closure problems. 

This pull request proposes that we add a pipeline delay on these critical paths. This delays the commit of the data to the DRP by one cycle, but greatly relieves timing because the pipeline register can be placed close to the cluster of logic that computes addresses, caching, and arbitration, allowing for the routing slack to the hard IP blocks to be absorbed by the path between the pipe register and the hard IP block. 

In general, this shouldn't be a problem because the algorithm to program the DRP is to hit the write or read CSR, and then poll the drdy bit until it is asserted (so the process is already pretty slow).  The MMCM in particular should have almost no impact, because MMCM updates are infrequent and the subsequent lock time of the MMCM is pretty long. The XADC is potentially more problematic because it can produce data at up to 1MSPS; but if sysclk is around 100MHz, adding 10ns to the read latency is relatively small compared to the theoretical maximum data rate of one every 1,000ns.

Note that the xadc patch requires introducing a bit of logic into the non-DRP path. This is because without explicitly putting an "if" statement around the logic, you fall back to the non-blocking semantics of the verilog operator, which ultimately leads to a pretty hefty combinational path. By having a default "if" that should get optimized out when DRP is not enabled, when the DRP path /is/ enabled the synthesizer knows it can safely push the async signal into a simple mux as opposed to worrying about enforcing the non-blocking operator semantics to get the desired result.

### DIFF
--- a/litex/soc/cores/clock.py
+++ b/litex/soc/cores/clock.py
@@ -102,22 +102,17 @@ class XilinxClocking(Module, AutoCSR):
 
         # # #
 
-        den_pipe = Signal()
-        dwe_pipe = Signal()
-
         drp_drdy = Signal()
         self.params.update(
             i_DCLK  = ClockSignal(),
-            i_DWE   = dwe_pipe,
-            i_DEN   = den_pipe,
+            i_DWE   = self.drp_write.re,
+            i_DEN   = self.drp_read.re | self.drp_write.re,
             o_DRDY  = drp_drdy,
             i_DADDR = self.drp_adr.storage,
             i_DI    = self.drp_dat_w.storage,
             o_DO    = self.drp_dat_r.status
         )
         self.sync += [
-            den_pipe.eq(self.drp_read.re | self.drp_write.re),
-            dwe_pipe.eq(self.drp_write.re),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks for the PR and detailed explanation. This make senses and has been merged with https://github.com/enjoy-digital/litex/commit/eae0e00496741b6afffb157b44576695237d7696. I did a manual rebase to keep a single commit and used your description as the commit description. This way we keep a trace of your explanation in the git.

PR #351, opened 2020-01-23
Fix sram size argument

mateusz-holenko:
Renaming the argument is a breaking change and might cause some user scripts to fail.
On the other hand this option is broken (wrong type), so I guess those scripts would fail anyway ;)


enjoy-digital:
Thanks for catching it!

PR #357, opened 2020-01-24
Add clk ce

bunnie:
Request to add a couple of extra clock buffer types to the MMCM module.

Gating requires an optional clock gate argument, maybe there is a better way to do it...


enjoy-digital:
Thanks! That's merged with minor changes: https://github.com/enjoy-digital/litex/commit/19ef19ce0dda86219da013bd378b523a31a6020a

`clk_ce` is renamed to `ce` (since the others parameters don't have the `clk_` prefix) and error reporting for unsupported clock buffers or missing `ce` Signal is improved.

PR #359, opened 2020-01-25
soc/software/bios/sdram: On ECP5 strobe dly_sel after read leveling

gregdavill:
This change appears to fix this issue I've seen with some of my ECP5 based OrangeCrab boards https://github.com/enjoy-digital/litedram/issues/103

I'm not entirely sure why this fixes things. But I suspect strobing the DQSBUFM.PAUSE lines after all the delays have been set is getting things to work.

### DIFF
--- a/litex/soc/software/bios/sdram.c
+++ b/litex/soc/software/bios/sdram.c
@@ -943,6 +943,13 @@ int sdrlevel(void)
 		printf("\n");
 	}
 
+#ifdef ECP5DDRPHY
+	/* Toggle all dly_sel lines. 
+	 * Which toggles all DQSBUFM.PAUSE lines, this ensures they're using the correct delays. */
+	ddrphy_dly_sel_write(0xFF);
+	ddrphy_dly_sel_write(0);
+#endif
+
 	return 1;
 }
 #endif
--- a/litex/soc/software/bios/sdram.c
+++ b/litex/soc/software/bios/sdram.c
@@ -419,6 +419,12 @@ static void read_delay_rst(int module) {
 
 	/* unsel module */
 	ddrphy_dly_sel_write(0);
+
+#ifdef ECP5DDRPHY
+	/* Sync all DQSBUFM's, By toggling all dly_sel (DQSBUFM.PAUSE) lines. */
+	ddrphy_dly_sel_write(0xFF);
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the investigation and PR. The fear i have with the fix is that we don't really know if the delays have been applied correctly during the read leveling, so if the read leveling is reliable.

The issue could be that the PAUSE we are doing when updating the delay is too short. Could you try adding a delay here: 
- https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/sdram.c#L419
- https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/sdram.c#L430

and see if it improves things?

If not, can you try to do the toggling at the end of these two functions?

gregdavill:
I'm always getting the bit-pattern |11100000|, for both modules. So I assume that the delays are doing something. They just don't seem to be applied once we switch to hardware-control?

This was the test with `cdelay(100)` added where you've indicated. Doesn't seem like PAUSE length is the issue.
```
--========== Initialization ============--
Initializing SDRAM...
SDRAM now under software control
Read leveling:
m0, b0: |11100000| delays: 01+-01
best: m0, b0 delays: 01+-01
m1, b0: |11100000| delays: 01+-01
best: m1, b0 delays: 01+-01
SDRAM now under hardware control
Memtest bus failed: 2/256 errors
Memtest data failed: 524288/524288 errors
Memtest addr failed: 8192/8192 errors
Memory initialization failed
```

gregdavill:
Adding This to the end of each function

```C
ddrphy_dly_sel_write(1 << module);
ddrphy_dly_sel_write(0);
```
Makes no difference.

But If we strobe all modules, this **does** work.
```C
ddrphy_dly_sel_write(0xFF);
ddrphy_dly_sel_write(0);
```
I'm still not sure on the exact specifics. But it seems that for some of my ECP5 boards, it's important to strobe PAUSE simultaneously for all modules before switching LiteDRAM back to hardware controlled?

I'll update this PR with the new changes.

enjoy-digital:
Thanks for the update and test, that's merged.  That's better to do it this way, so that we know that after each rst/inc all DQSBUFM have been updated and not only at the end.

Issue #365, opened 2020-01-29
CSR accessor performance / feature / build Regressions


xobs:
Due to a recent change, the CSR accessor methods have undergone a large performance regression.  Additionally, the Etherbone library is now incompatible.

# Performance Regression

Previously, CSRs were defined similarly to:

```cpp
static inline void csr_writel(uint32_t value, uint32_t addr)
{
	*((volatile uint32_t *)addr) = value;
}
```

This allowed the compiler to inline writes very easily.  For example, in Fomu we have the following function:

```cpp
static void spi_single_tx(uint8_t out) {
	int bit;

	for (bit = 7; bit >= 0; bit--) {
		if (out & (1 << bit)) {
			lxspi_bitbang_write((0 << PIN_CLK) | (1 << PIN_MOSI));
			lxspi_bitbang_write((1 << PIN_CLK) | (1 << PIN_MOSI));
			lxspi_bitbang_write((0 << PIN_CLK) | (1 << PIN_MOSI));
		} else {
			lxspi_bitbang_write((0 << PIN_CLK) | (0 << PIN_MOSI));
			lxspi_bitbang_write((1 << PIN_CLK) | (0 << PIN_MOSI));
			lxspi_bitbang_write((0 << PIN_CLK) | (0 << PIN_MOSI));
		}
	}
}
```

This would get compiled down to the following function:

```asm
(gdb) disassemble spi_single_tx
Dump of assembler code for function spi_single_tx:
   0x0000014c <+0>:     li      a4,7
   0x00000150 <+4>:     lui     a5,0xe0008
   0x00000154 <+8>:     li      a6,2
   0x00000158 <+12>:    li      a2,1
   0x0000015c <+16>:    li      a7,3
   0x00000160 <+20>:    li      a1,-1
   0x00000164 <+24>:    sra     a3,a0,a4
   0x00000168 <+28>:    andi    a3,a3,1
   0x0000016c <+32>:    beqz    a3,0x188 <spi_single_tx+60>
   0x00000170 <+36>:    sw      a2,-2048(a5) # 0xe0007800
   0x00000174 <+40>:    sw      a7,-2048(a5)
   0x00000178 <+44>:    sw      a2,-2048(a5)
   0x0000017c <+48>:    addi    a4,a4,-1
   0x00000180 <+52>:    bne     a4,a1,0x164 <spi_single_tx+24>
   0x00000184 <+56>:    ret
   0x00000188 <+60>:    sw      zero,-2048(a5)
[...]

PR #366, opened 2020-01-29
software, integration/export: (re-)expose CSR subregister accessors

gsomlo:
Expose a pair of `csr_[read|write]_sub()` subregister accessors, and
restore the way dedicated accessors are generated in "generated/csr.h"
to use hard-coded combinations of shifts and subregister accessor call.

This restores downstream ability to override CSR handling at the
subregister accessor level.

This is a followup to PR #339, and should hopefully address issue #365.

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -136,7 +136,7 @@ def get_soc_header(constants, with_access_functions=True):
     r += "\n#endif\n"
     return r
 
-def _get_rw_functions_c(reg_name, reg_base, nwords, busword, read_only, with_access_functions):
+def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, read_only, with_access_functions):
     r = ""
 
     addr_str = "CSR_{}_ADDR".format(reg_name.upper())
@@ -146,7 +146,7 @@ def _get_rw_functions_c(reg_name, reg_base, nwords, busword, read_only, with_acc
 
     size = nwords*busword//8
     if size > 8:
-        # FIXME: maybe implement some "memcpy-like" semantics for larger blobs?
+        # downstream should select appropriate `csr_[rd|wr]_buf_uintX()` pair!
         return r
     elif size > 4:
         ctype = "uint64_t"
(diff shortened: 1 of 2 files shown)


mithro:
FYI - @xobs

litex/soc/integration/export.py
```diff
@@ -176,14 +191,8 @@ def get_csr_header(regions, constants, with_access_functions=True):
...
-        r += "extern void csr_wr_uint8(uint8_t v, unsigned long a);\n"
-        r += "extern void csr_wr_uint16(uint16_t v, unsigned long a);\n"
-        r += "extern void csr_wr_uint32(uint32_t v, unsigned long a);\n"
-        r += "extern void csr_wr_uint64(uint64_t v, unsigned long a);\n"
-        r += "extern uint8_t csr_rd_uint8(unsigned long a);\n"
-        r += "extern uint16_t csr_rd_uint16(unsigned long a);\n"
-        r += "extern uint32_t csr_rd_uint32(unsigned long a);\n"
-        r += "extern uint64_t csr_rd_uint64(unsigned long a);\n"
+        r += "extern void csr_write_sub(unsigned long v, unsigned long a);\n"
+        r += "extern unsigned long csr_read_sub(unsigned long a);\n"
```
mithro: These being marked as `extern` will mean the compiler can't optimize them without lto turned on (see https://github.com/enjoy-digital/litex/pull/253 for that).

litex/soc/software/include/hw/common.h
```diff
@@ -35,12 +34,22 @@
...
  */
 
-/* CSR data width (subregister width) in bytes, for direct comparson to sizeof() */
+/* CSR data width (subreg. width) in bytes, for direct comparson to sizeof() */
 #define CSR_DW_BYTES (CONFIG_CSR_DATA_WIDTH/8)
 
 /* CSR subregisters are embedded inside native CPU word aligned locations: */
 #define MMPTR(a) (*((volatile unsigned long *)(a)))
 
+static inline void csr_write_sub(unsigned long v, unsigned long a)
```
mithro: Maybe `csr_write_part` or `csr_write_section` are better names than `csr_write_sub`?

litex/soc/integration/export.py
gsomlo: this is the case where they're externally provided by downstream, and downstream defines CSR_ACCESSORS_DEFINED to let us know we shouldn't be using the built-in ones.

That's the way it was before I ever got involved :) The litex-provided accessors (in common.h) are definitely _NOT_ marked as external

litex/soc/software/include/hw/common.h
gsomlo: 'sub' is for "subregister" (how about 'subreg' then)? I could also go with "slice". Both "part" and "section" are IMHO terms that are too overloaded to ensure we don't confuse future readers of the source code...

enjoy-digital: Or should we call it `csr_write_simple` or to mach the `simple_csrs` used in https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/csr.py? (if simple is confusing we could change it to something better in both software/gateware later)

gsomlo: > Or should we call it `csr_write_simple` or to mach the `simple_csrs` ...

That works for me, thanks! Force-pushed an amended commit reflecting that.

xobs:
This works for me.  It should get inlined easily, and should still work on both 32 and 64 bit architectures.  And the new name should prevent confusion, and is guaranteed to work on any new architectures we come across.

Maybe we'll find someday that we need to have it accept a `far *` pointer.  That would be silly.

Looks good!

enjoy-digital:
Looks also good for me, thanks @gsomlo, @xobs and @mithro.

PR #369, opened 2020-01-30
WIP: Attempt at doing a better add_memory_region

mithro:
alloc_memory_region seems to not work properly however?

### DIFF
--- a/litex/soc/integration/common.py
+++ b/litex/soc/integration/common.py
@@ -1,12 +1,15 @@
 # This file is Copyright (c) 2019 Florent Kermarrec <florent@enjoy-digital.fr>
 # License: BSD
 
-import os
-import math
-import json
-import time
-import struct
 import datetime
+import enum
+import json
+import math
+import os
+import struct
+import time
+
+from collections import namedtuple
 
 from migen import *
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Sorry @mithro, but i'm not sure to understand what you want to do.

The idea with https://github.com/enjoy-digital/litex/commit/53bc18cc3f739ed02b95a9e9e530915e22b3e9d2 is to prepare for an API change where `add_mem_region` will replace `add_memory_region` to avoid user the need to provide the mem origin with the call. If a matching name can be found in the `mem_map,` then this one is used, otherwise it is automatically allocated (we first limit the allocation region depending the type of the memory (Cached, IO, etc...) and then try to find the first available region that is large enough).

This is the same reasoning that is used with `add_csr` or `add_irq.` From what i understand, in this PR, you are doing the same, but to avoid breaking API, you add a complex handling of the arguments and then complex checks to be sure the old API was behaving correctly, which complicates the code a lot and is going to be very complicated to maintain.

From the test i did, current `alloc_mem_region` and `add_mem_region` was behaving correctly, but i just needed to do more testing on real design to be sure there were no regression. So i would just suggest i do more testing on that, and then we could keep for now add_memory_region compatibility but add a deprecated warning to it (but no need improve it with checks since the idea is deprecate it and checks will no longer be needed).

What do you think?

enjoy-digital:
Thanks for the update! The usage seems closer to what i had in mind than your previous PR and the implementation is probably better than my solution, i'll will review it and test it.

litex/soc/integration/soc_core.py
```diff
@@ -402,67 +402,69 @@ def check_regions_overlap(regions):
...
     def alloc_mem_region(self, name, length, type):
-        # Use type to limit search regions
-        search_regions = []
-        if "io" in type:
-            for _origin, _length in self.soc_io_regions.items():
-                search_regions.append(SoCMemRegion(origin=_origin, length=_length, type=type))
+        # Linker only regions have to be explicitly specified.
+        assert SoCMemRegion.Properties.linker_only not in type
+
+        CACHED_BOUNDRY = 0x80000000
```
mithro: I think we should probably defined a `self.soc.cached_region` and a `self.soc.io_region` type things?

enjoy-digital: We already have a `self.io_regions` and `self.soc_io_regions`: https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc_core.py#L113 but no `cached_region` since non `io_regions ` we considered cached. But we could add it if you think it's useful.

enjoy-digital:
@mithro: after having a closer look at the PR, i think i'd like to have time to finish testing my implementation since i find it simpler, with a codestyle closer to the others part of SoCCore, and from the tests i did, was working, i just need to spend a bit more time on it testing on complex designs. In the future, i'd like we rework the SoC classes (maybe have a single SoC class, allow more SoC flexibility, with cleaner code and more elegant coding style than the current code) and i think the changes you want to do now would make more sense at this point. So to keep things consistent, let's keep the current implementation for now, i'll finish testing it and we'll improve all the others points while creating the new SoC class. I hope you understand i'm not rejecting your changes, just that i think we should wait before revisiting too much these part of the code.

enjoy-digital:
I'm closing this, but will re-open an issue soon to discuss the way we could improve the SoC classes (try to have a single one) and we could discuss there how we could improve all the CSR/IRQ/Memory allocation functions there.

mithro:
@enjoy-digital - At the moment it seems to be allocating memory regions that are already in use.

FYI -- I would like to use the `SoCMemRegion` style stuff for flash regions in the near future too.

enjoy-digital:
@mithro: OK, i'll add a unit test for this to ease the work.

mithro:
@enjoy-digital - I reopened the pull request temporarily to continue some discussion even if the pull request will never be merged.

litex/soc/integration/soc_core.py
```diff
@@ -345,6 +345,15 @@ def add_csr(self, csr_name, csr_id=None, allow_user_defined=False):
...
 
+    def add_memory(self, name, XXX):
+        pass
+
+    def add_linker_region(self, name, XXXX):
+        pass
+
+    def add_flash_region(self, name, XXXX):
+        pass
+
```
mithro: @enjoy-digital - I was thinking the new API should look something like this?

litex/soc/integration/common.py
```diff
@@ -76,20 +80,290 @@ def get_mem_data(filename_or_regions, endianness="big", mem_size=None):
...
-    def __init__(self, origin, length, type):
-        assert type in ["cached", "io", "cached+linker", "io+linker"]
-        self.origin = origin
-        self.length = length
-        self.type   = type
 
-    def __str__(self):
-        return "<SoCMemRegion 0x{:x} 0x{:x} {}>".format(
-            self.origin, self.length, self.type)
+class SoCRegion:
```
mithro: I would find the `SoCRegion` class improvements really useful in LiteX-BuildEnv even if the changes to `litex/soc/integration/soc_core.py` aren't used.

enjoy-digital:
@mithro: i just created https://github.com/enjoy-digital/litex/issues/372 to have a more general discussion about the roadmap and features, we should probably move the discussion there. We could discuss the general ideas/direction before touching too much the code for now and i would like to take some time to prototype a refreshed and more generic SoC class.

enjoy-digital:
Let's continue the discussion about that in #372 with the new SoC builder.

PR #374, opened 2020-02-03
SVD export

Disasm:
SVD format is used among different ARM microcontrollers for describing the chip peripherals.
Peripheral descriptions in this format can also be used to generate Rust bindings with [svd2rust](https://github.com/rust-embedded/svd2rust/).

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -39,7 +39,7 @@ class Builder:
     def __init__(self, soc, output_dir=None,
                  compile_software=True, compile_gateware=True,
                  gateware_toolchain_path=None,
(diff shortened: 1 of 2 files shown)


mithro:
PTAL - @xobs

xobs:
This is possible with lxsocdoc (https://github.com/litex-hub/lxsocdoc).  This project needs to get integrated into litex.

xobs:
Note that with those SVD files, registers are completely documented, and as a result we get autogenerated Rust documents that look like: https://docs.rs/fomu-pac/0.0.3/fomu_pac/

Disasm:
Oh, indeed, thank you for pointing out!

Issue #367, opened 2020-01-29
Make lxsocdoc part of LiteX directly?


mithro:
[lxsocdoc](https://github.com/litex-hub/lxsocdoc) from @xobs is really cool. 

[It generates really nice documentation for CSR registers](https://rm.fomu.im/). It also generates nice SVD files for usage by things like rust bindings. 

As it seems **generally** useful to anyone using the litex CSR registers, I think it should be included as part of LiteX. It will also help promote the idea of better documenting the CSR registers too.

@xobs / @enjoy-digital - What do you think?

enjoy-digital:
@mithro: I also find it really cool and already started using it on different projects, so would be happy to have it directly integrated if that's the intention of @xobs. (I also like a lot lxbuildenv :))

xobs:
Sure thing.  Once you get used to documenting registers, life becomes so much simpler!  Future-you will thank today-you.

The current codebase is a bit of a mess, but that can be fixed over time.  I've just fixed basic support for non-8-bit CSRs (it was mostly compatible, just had `8` hardcoded in a few places), and need to add support for register lanes.

What would be the process to bring in lxsocdoc?

enjoy-digital:
@xobs we could integrate the current version with the others tools in: https://github.com/enjoy-digital/litex/tree/master/litex/tools and install it with LiteX as we are doing with the others tools: https://github.com/enjoy-digital/litex/blob/master/setup.py#L45-L48. I'm happy to work with you on improving it over time. Can you create a PR for that if that's fine for you?

xobs:
All of the tools under `litex/tools` are executables, whereas `lxsocdoc` is more of a pipeline flow.  That is, it's meant to be added just following execution of `Builder`.

Might it make more sense to put it under a new directory `litex/gen/doc/`?

enjoy-digital:
In fact `litex/gen` is more related to Migen level specific features. What about `litex/soc/doc`? We could also move the current `litex/soc/integration/doc.py` to it (eventually renamed to` autodoc.py`?). This also makes things consistent: `lxsocdoc` would be the short name for `litex/soc/doc` and the import could be done with: `from litex.soc.doc import *` or just as you were already doing `from lxsocdoc import *`

enjoy-digital:
Integrated with https://github.com/enjoy-digital/litex/pull/375.

PR #375, opened 2020-02-04
Add lxsocdoc

xobs:
As discussed in #367 #367


litex/soc/doc/__init__.py
```diff
@@ -0,0 +1,329 @@
...
+def sub_csr_bit_range(busword, csr, offset):
+    nwords = (csr.size + busword - 1)//busword
+    i = nwords - offset - 1
+    nbits = min(csr.size - i*busword, busword) - 1
+    name = (csr.name + str(i) if nwords > 1 else csr.name).upper()
+    origin = i*busword
+    return (origin, nbits, name)
+
+
+def print_svd_register(csr, csr_address, description, length, svd):
```
mithro: The `svd` stuff should probably be in the same place we generate the json and c headers rather than in the doc?

xobs: Is there an easy way in python to convert a print() to a string concatenation?  The existing litex stuff creates one big string inside the functions in `export.py`, then writes them all out at once.  lxsocdoc, meanwhile, takes a stream-based approach and generally writes to the files as it goes.

enjoy-digital: Maybe something similar to what we are doing in the build backends:
````
from litex.build import tools
svd = []
svd.append("line1")
svd.append("line2")
tools.write_to_file("registers.svd", "\n".join(svd))
````

Comment (xobs) [reply litex/soc/doc/__init__.py]: As per your suggestion, it's now in `export.py`.

Comment (enjoy-digital) [reply litex/soc/doc/__init__.py]: Thanks!

**Comment by enjoy-digital** [comment]
Thanks @xobs, this seems fine for an initial integration. Do you have some code snippet demonstrating how to use it on a existing SoC?

**Comment by xobs** [comment]
We do this on Fomu: https://github.com/im-tomu/foboot/blob/master/hw/foboot-bitstream.py#L449-L452

```python
    vns = builder.build()
    soc.do_exit(vns)
    lxsocdoc.generate_docs(soc, "build/documentation/", project_name="Fomu Bootloader", author="Sean Cross")
    lxsocdoc.generate_svd(soc, "build/software", vendor="Foosn", name="Fomu")
```

There are instructions inside the lxsocdoc [README file](https://github.com/litex-hub/lxsocdoc#usage), however those obviously will be lost with the merge into litex.

enjoy-digital:
I'm merging and will add the README file of lxsocdoc in `litex/doc` as `lxsocdoc.md`, thanks.

PR #376, opened 2020-02-04
build/sim: allow to use environment's {C,LD}FLAGS

mglb:
There are use cases where additional flags should be added to CFLAGS or
LDFLAGS, e.g. when using Conda environment.


enjoy-digital:
Thanks.

PR #378, opened 2020-02-06
Request to merge I2S and SPIOPI cores

bunnie:
Request to merge I2S and SPIOPI cores. If you want me to split into two separate PR's, I can back this out and do that too.

### DIFF
--- a/litex/soc/cores/i2s.py
+++ b/litex/soc/cores/i2s.py
@@ -131,11 +131,11 @@ class i2s_slave(Module, AutoCSR, AutoDoc):
         # interrupts
         self.submodules.ev = EventManager()
         if hasattr(pads, 'rx'):
-            self.ev.rx_ready = EventSourcePulse()  # rising edge triggered, indicates FIFO is ready to read
-            self.ev.rx_error = EventSourcePulse()  # indicates a rx error has happened
+            self.ev.rx_ready = EventSourcePulse(description="Indicates FIFO is ready to read")  # rising edge triggered
+            self.ev.rx_error = EventSourcePulse(description="Indicates an Rx error has happened (over/underflow)")
         if hasattr(pads, 'tx'):
-            self.ev.tx_ready = EventSourcePulse()  # indicates space available in Tx buffer for next quantum
-            self.ev.tx_error = EventSourcePulse()  # indicates a tx error has happened
+            self.ev.tx_ready = EventSourcePulse(description="Indicates enough space available for next Tx quanta of {} words".format(fifodepth))
+            self.ev.tx_error = EventSourcePulse(description="Indicates a Tx error has happened (over/underflow")
         self.ev.finalize()
 
 
--- a/litex/soc/cores/i2s.py
+++ b/litex/soc/cores/i2s.py
@@ -1,0 +1,435 @@
+# This file is Copyright (c) 2020 bunnie <bunnie@kosagi.com>
+# License: BSD
+
+from litex.soc.interconnect.csr_eventmanager import *
+from litex.soc.interconnect import wishbone
+from litex.soc.integration.doc import AutoDoc, ModuleDoc
+from migen.genlib.cdc import MultiReg
+
+class i2s_slave(Module, AutoCSR, AutoDoc):
+    def __init__(self, pads, fifodepth=256):
+        self.intro = ModuleDoc("""
+        Intro
+        *******
+        
+        I2S slave creates a slave audio interface instance. Tx and Rx interfaces are inferred
+        based upon the presence or absence of the respective pins in the "pads" argument.
+        
+        The interface is I2S-like, but note the deviation that the bits are justified
+        left without a 1-bit pad after sync edges. This isn't a problem for talking to the LM49352
+        codec this was designed for, as the bit offset is programmable, but this will not work well if
+        are talking to a CODEC without a programmable bit offset!
+        
+        System Interface
+        =================
+        
+        Audio interchange is done with the system using 16-bit stereo samples, with the right channel
+        mapped to the least significant word of a 32-bit word. Thus each 32-bit word is a single
+        stereo sample. As this is a slave I2S interface, sampling rate and framing is set by the programming 
+        of the audio CODEC chip. A slave situation is preferred because this defers the generation of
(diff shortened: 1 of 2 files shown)


bunnie:
not sure I follow? can you give more details?

litex/soc/cores/i2s.py
```diff
@@ -0,0 +1,435 @@
...
+               self.bus.ack.eq(wr_ack),
+            ).Else(
+                self.bus.ack.eq(rd_ack),
+            )
+        ]
+
+        # interrupts
+        self.submodules.ev = EventManager()
+        if hasattr(pads, 'rx'):
+            self.ev.rx_ready = EventSourcePulse()  # rising edge triggered, indicates FIFO is ready to read
```
xobs: You can add a description to an EventSource object to get it documented, so you can turn the comment into documentation.

https://github.com/enjoy-digital/litex/blob/ae085782aed7f448cd205628532a7573781a34a6/litex/soc/interconnect/csr_eventmanager.py#L48

litex/soc/cores/spiopi.py
```diff
@@ -0,0 +1,1080 @@
...
+                      NextValue(spicount, spicount - 1),
+                      NextValue(spi_clk_en, 1),
+                      ).Elif(spicount <= 1 & spi_req,
+                             NextValue(spi_clk_en, 1),
+                             NextValue(spicount, 7),
+                             NextValue(spi_so, spi_do),  # reload the so register
+                             NextValue(spi_ack_pipe, 1),  # finally ack the cycle
+                             NextValue(spi_dummy, has_dummy),
+                             ).Else(
+                       NextValue(spi_clk_en, 0),
```
xobs: Here is an example of strange indentation

bunnie: oh geez...i see it now. I did a copy/paste to a new file before merging to not break the legacy spi stuff. Pycharm went and auto-indented a bunch of areas. I'll fix that now.

litex/soc/cores/i2s.py
bunnie: just added comments to these, but left the # comment about rising/falling edge triggering because I can never remember which polarity is "Process" versus "Pulse". I don't know why they don't just use EventSourceRising(), EventSourceFalling(), and EventSourceLevel() as API names...

bunnie:
> 
> 
> You wrapping in these files is a bit weird....

can you confirm it's better now that I am not relying on PyCharm to auto-format things?

litex/soc/cores/spiopi.py
```diff
@@ -0,0 +1,1063 @@
+from litex.soc.interconnect.csr_eventmanager import *
+from litex.soc.interconnect import wishbone
+from litex.soc.integration.doc import AutoDoc, ModuleDoc
+from migen.genlib.cdc import MultiReg
+
+class SpiOpi(Module, AutoCSR, AutoDoc):
```
mithro: This device is Series7 specific right? It would be nice if the device specific part was separated from the generic stuff.

litex/soc/cores/spiopi.py
```diff
@@ -0,0 +1,1063 @@
...
+                             p_CINVCTRL_SEL="FALSE", p_HIGH_PERFORMANCE_MODE="FALSE", p_REFCLK_FREQUENCY=200.0,
+                             p_PIPE_SEL="FALSE", p_IDELAY_VALUE=dq_delay_taps, p_IDELAY_TYPE=delay_type,
+
+                             i_C=ClockSignal(), i_CINVCTRL=0, i_REGRST=0, i_LDPIPEEN=0, i_INC=0, i_CE=0,
+                             i_LD=self.delay_update,
+                             i_CNTVALUEIN=self.delay_config.fields.d, o_CNTVALUEOUT=self.delay_status.fields.q,
+                             i_IDATAIN=dq.i[i-1], o_DATAOUT=dq_delayed[i],
+                    ),
+                else: # don't wire up o_CNTVALUEOUT for others
+                    self.specials += Instance("IDELAYE2",
```
mithro: I would suggest wrapping right after the first brace so everything isn't indented so much.
(Same with the state machine below).

litex/soc/cores/spiopi.py
bunnie: Yes, to get it to close timing at 100MHz, you need to do vendor-specific I/O.

To clarify, are you suggesting that the core should not be included here because it is device-specific (which is fine with me), or are you suggesting there should be a "generic" implementation for every vendor-specific part? If it's the latter, the problem is, you won't meet timing, or you'd have to reduce the clock rate on your system so much the core doesn't make sense any more to use...

enjoy-digital: @mithro: i'm fine doing this and adding a banner to specify it's only for 7-Series for now. @bunnie: That's fine having device specific cores/parts, as @mithro suggest, some parts of the core could probably be reused for others devices, but i'm happy doing the work on easing reuse and don't want to bother you with that.

enjoy-digital:
Thanks @bunnie! i just merged the code and will do the cosmetic changes myself.

enjoy-digital:
Here are the cosmetic changes:
https://github.com/enjoy-digital/litex/commit/f58e8188b7ed18593f762ccefff2f96b5848e454
https://github.com/enjoy-digital/litex/commit/62f3537db0bb159444aa4d245b94fc34ea306075
https://github.com/enjoy-digital/litex/commit/1d70ef6958bfc11cafc80d29e0ad5550d9d0fc89

On I2S: i2s_slave is renamed to S7I2SSlave and fifodepth changed to fifo_depth (for consistency with the others cores.)
On SPI OPI: the core is renamed to spi_opi and module to S7SPIOPI.

@bunnie: can you adapt your code for that and test your project with the cores from uptstream LiteX?

litex/soc/cores/spiopi.py
mithro: @bunnie - This looks really useful and a great addition! I was suggesting that the core could be split into 7-series and generic parts. **Other** people can then create the equivalent of the 7-series bits for the iCE40 / ECP5 / etc.

bunnie:
yep, confirmed to work with the upstream LiteX with full regression to hardware. thanks!

PR #381, opened 2020-02-11
spi_flash: Add quadspi support

mateusz-holenko:
This commit adds:
* quadspi support to bitbang mode,
* quadspi memory mapped read and write,
* flash register read and write via CSR.

### DIFF
--- a/litex/boards/targets/arty.py
+++ b/litex/boards/targets/arty.py
@@ -14,7 +14,7 @@ from litex.soc.cores.clock import *
 from litex.soc.integration.soc_core import *
 from litex.soc.integration.soc_sdram import *
 from litex.soc.integration.builder import *
-
+from litex.soc.cores.spi_flash import *
 from litedram.modules import MT41K128M16
 from litedram.phy import s7ddrphy
 
@@ -50,6 +50,11 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
+    mem_map = {
+        "spi_flash4x0": 0x50000000,
+    }
+    mem_map.update(SoCSDRAM.mem_map)
+    
     def __init__(self, sys_clk_freq=int(100e6), with_ethernet=False, with_etherbone=False, **kwargs):
         platform = arty.Platform()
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks for the quadspi support, we have to be careful since this introduce an API change, maybe we'll have to keep support for the old API but i could do that. Otherwise, can you explain your usecase for this and have been able to evaluate the speed improvements for reads(bitbang)/writes(mmap)? Have you created a testbench to create these changes that we could reuse?

mateusz-holenko:
We can rework the code to avoid unnecessary API change by introducing a new constructor argument to `SpiFlash` that will allow to choose between the old implementation (without quad support, no API change) or the new one (that needs a different API due to the new data lines). 

What do you think?

enjoy-digital:
Thakns, yes that would be a lot easier to merge. Writes on SPI Flashes are generally very slow, and i'm suspecting we don't have speed improvement with QUAD SPI support on writes. Have you been able to evaluate speed improvements? Or do you need QUAD SPI for another purpose? (no need to change mode between reads/writes?) Sorry for asking this, i'm just trying to understand the use-case.

mateusz-holenko:
Our main aim is to allow booting rootfs directly from flash. To make it efficient we need to avoid bit-banging, so we decided to implement `memmap` approach for writing. This allows us to use 50MHz SPI clock (was 60Hz in bitbang mode).

Adding quad support was a bonus that additionally gave us about 15% faster reads.

mithro:
@mateusz-holenko - You should **not** need write support for booting from flash?

mateusz-holenko:
@mithro - Yes, but if we want the rootfs to be persistent we need to be able to write it.

enjoy-digital:
@mateusz-holenko thanks for the additional informations. Is it  15% faster for writes? (instead of reads). If so, then yes Quad support is justified for writes and we could merge if it's still possible to use the old API.

mateusz-holenko:
@enjoy-digital: We are currently reworking the proposed solution so it's easier to benchmark it and compare single- and quad- modes.

enjoy-digital:
@mateusz-holenko: thanks, that will be very useful to find the best trade off.

mateusz-holenko:
We reworked the `spi_flash` module and split our changes into separate commits to make it easier to review and decide which features should be merged to master.

When it comes to benchmarking the performance of QUAD writes, in the meantime we discovered problems with handling timers in the Zephyr and are fixing this now. Once it's done we will run the benchmarks again to see what is the performance speed-up.

enjoy-digital:
@mateusz-holenko: thanks for the update. BTW, are you aware of the work currently done in https://github.com/litex-hub/litespi? The aim is to create a SPI core that will be more flexible and configurable than current `spi_flash` and will probably replace it in the future. Both work are complementary and your improvements/benchmarcks to `spi_flash` will be used for now and will provide a reference for LiteSPI.

mateusz-holenko:
@enjoy-digital Thanks for the info! Yes, I'm aware of the existence of `LiteSPI` but as you mentioned `spi_flash` can serve as a nice reference in the future.

mateusz-holenko:
@enjoy-digital I created a separate PR #478 that does not include full quad support (as we need first to assess the performance of full quad writes, but this in turns waits for fixing Zephyr timer support).

I think this PR should be wait for now and eventually be rebased so that it contains only quad spi related changes.

enjoy-digital:
Closing, part of the PR has already been merged with https://github.com/enjoy-digital/litex/pull/478. We'll create another up to date PR is needed.

Issue #372, opened 2020-02-01
LiteX 2020 roadmap and new LiteX SoC class


enjoy-digital:
LiteX has evolved a lot and the current `SoCCore`, `SoCSDRAM` inheritated from MiSoC have evolved quite a bit in the last years and could be greatly improved now that we have a better idea of the SoCs we want to be able build with LiteX and the features we want to support.

A new refreshed LiteXSoC class could be created and could be flexible enough to:
**In the short term:**

- [ ] allow SoCCore/SoCSDRAM/SoCZynq to inherit from it for retro-compatibility with existing designs.
- [x] support different types of main SoC bus (Wishbone, AXI, etc...) with different data-widths.
- [x] support automatic bridging between bus types and data-width adaptation when adding a bus Master or Slave to the main bus (ex: AXI 64-bit to Wishbone 32-bit, Wishbone 32-bit to AXI 32-bit, etc...)
- [x] support AXI-Lite ROM/SRAM/Interconnect when main bus is AXI (requires adding AXI-Lite ROM, SRAM, Interconnect components to LiteX). 
- [x] support mixed dynamic/static CSRs regions allocation (some CSR regions could be user-defined or reserved by the CPU (and then static), the others could dynamically allocated).
- [x] support AXI-Lite to CSRs direct bridging when the main bus is AXI. (we are currently doing AXI->Wishbone->CSR which is not efficient).
- [ ] support mixed dynamic/static CSRs allocation within a CSR region. (It's currently difficult to use static CSRs regions since CSRs can move within a module when a new CSR is added).
- [x] support mixed dynamic/static IRQs numbering (some IRQs could be user-defined or reserved by the CPU (and then static), the others could dynamically allocated).
- [x] support mixed dynamic/static Memory regions allocation (some Memory regions could be user-defined or reserved by the CPU (and then static), the others could dynamically allocated).
- [x] support properties on Memory regions: IO/Cached/Linker, allow the CPU (or main Master of the bus) to define the IO/Cached regions and checks regions properties when adding a new peripheral to the memory map.
- [ ] ease multiple instances of a similar component (SDRAM, Ethernet, etc...): It's already possible, but requires workarounds.
[...]

PR #382, opened 2020-02-11
Add new SoC/LiteXSoC classes and use it for SoCCore/SoCSDRAM

enjoy-digital:
As described in https://github.com/enjoy-digital/litex/issues/372, a refactor of the SoC classes was needed to cleanup the code and allow future improvements. In this PR New SoC/LiteXSoC classes are introduced and used as base to provide SoCCore/SoCSDRAM full retro-compatbility. No changes should be required on User's designs.


enjoy-digital:
This has been tested in retro-compatibility mode with most of the designs i maintain or have access to. This already provides useful reporting while building the SoC so i'm merging to go further and start using new features/possibilities.

enjoy-digital:
FYI @mithro, @xobs, @mateusz-holenko : the only thing i saw that is not supported by the retro-compatibility layer is the support of `address_decoder` provided to `add_wb_slave`. I'm not sure it's easy to provide retro-compatibility for it and the change on the user's side is easy: just pass the `address` instead of the `address_decoder` as done in https://github.com/litex-hub/litex-boards/commit/c94360c2e0c0bc95a059b66199e5fbbf033bb75c. I saw you were using it in your design, so will probably have to do these small changes when using upstream LiteX. I hope you won't have others changes to do, please open PR if so.

xobs:
There was a regression in soc/doc and the svd generator, which I fixed.

Because this breaks lxsocdoc, I've posted a note saying the project is now deprecated and integrated into litex: https://github.com/litex-hub/lxsocdoc

enjoy-digital:
@xobs thanks for the feedback and lxsocdoc update. The regression was not volontary and we could also have added retro-compatibility to SoCCore if needed, sorry for that. Now that lxsocdoc is merged we would have probably deprecated it at some point to ease maintenance, so if you think that's the right time to do it, let's do that.

PR #383, opened 2020-02-12
Unify output directory handling in builder

Xiretza:
All directories are now set as instance variables, allowing easy override using command-line arguments.


enjoy-digital:
Great, thanks!

PR #390, opened 2020-02-20
Import LiteSDCard support in to LiteX, using nexys4ddr as the initial test target

gsomlo:
Modify targets/nexys4ddr to facilitate easy selection of any combination of
'--with-ethernet', '--with-sdcard', both, or none, through the target builder
command line.

Import the current LiteSDCard demo's bios command-prompt initialization
and testing options into the "upstream" LiteX bios.

The idea is that, eventually, those initialization and test routines will morph
into the option to boot a kernel from microSD instead of (or in addition to)
TFTP over Ethernet.


enjoy-digital:
Thanks @gsomlo.

Issue #394, opened 2020-02-20
RFC: Split LiteX CPU cores into their own Python modules


mithro:
With the increasing number of CPU cores supported by LiteX, cloning the LiteX repository means cloning a large number of submodules which can get very big.

I think we should convert each of the supported LiteX CPUs into their own Python packages. 

`litex/soc/cores/cpu` becomes a "namespace module" which other Python packages can provide modules under. See https://packaging.python.org/guides/packaging-namespace-packages/

The CPUs can be split into;
 - [ ] `litex-cpu-lm32`
 - [ ] `litex-cpu-mor1k`
 - [ ] `litex-cpu-vexriscv`
 - [ ] `litex-cpu-blackparrot`
 - [ ] `litex-cpu-microwatt`
 - [ ] `litex-cpu-chiselwatt`
 - [ ] `litex-cpu-rocket`
 - [ ] `litex-cpu-picorv32`
 - [ ] `litex-cpu-minerva`

This allows projects which are only using a single CPU like VexRISCV to only pay the cost of that CPU.

The setup process then becomes;
```
git clone git+ssh://github.com/enjoy-digital/litex.git
(cd litex; python setup.py install)
git clone git+ssh://github.com/enjoy-digital/litex-cpu-vexriscv.git
(cd litex-cpu-vexriscv; python setup.py install)
```

People can also submodule the repository in using the normal method of;
```
git submodule add third_party/litex git+ssh://github.com/enjoy-digital/litex.git
(cd third_party/litex; python setup.py -e install)
git submodule add third_party/litex-cpu-vexriscv git+ssh://github.com/enjoy-digital/litex-cpu-vexriscv.git
(cd third_party/litex-cpu-vexriscv; python setup.py -e install)
```

Or specify a Python requirements file using;
```
-e https://github.com/enjoy-digital/litex.git#egg=litex
-e https://github.com/enjoy-digital/litex-cpu-vexriscv.git#egg=litex-cpu-vexriscv
```

We can easily set up Travis CI so that any push to the LiteX repository causes each of the CPU packages to be tested in combination with the based LiteX.

As VexRISCV is currently the most popular CPU to use with LiteX, we could also make VexRISCV the "default" CPU that is included with LiteX by default and just move the other CPU cores into their own modules. I personally **don't** think we should preference VexRISCV like that, but it could be a possible compromise?

Thoughts?

mithro:
FYI - @xobs, @gsomlo, @enjoy-digital

[...]

PR #395, opened 2020-02-21
WIP: Improve Submodules/Setup

enjoy-digital:
As discussed in https://github.com/enjoy-digital/litex/issues/394, the fact that we are now supporting multiple CPUs and still recommend the git clone in --recursive mode for the installation has a side effect that we are now downloading lots of submodules that are potentially not going to be used and which increases LiteX's download time and disk size.

This PR reorganize the submodules in a `third_party` directory (so that we have a better naming and visibility of the submodules), changes default `git clone` behaviour to **non-recursive** and adds a `submodule-init` command that can be used to initialize all submodules or a only specific one:

Minimal LiteX install with all the cores, but no CPU support:
````
$ ./litex_setup.py --init --install
````

Minimal LiteX install with all the cores, with VexRiscv support:
````
$ ./litex_setup.py --init --install
$ ./litex_setup.py --submodule-init=litex_compiler_rt
$ ./litex_setup.py --submodule-init=litex_cpu_vexriscv
````

Full LiteX install with all the cores and support for all CPUs:
````
$ ./litex_setup.py --init --install --submodule-init=all
````

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,23 @@
 [submodule "third_party/litex_cpu_blackparrot"]
 	path = third_party/litex_cpu_blackparrot
 	url = https://github.com/enjoy-digital/black-parrot.git
+[submodule "litex_cpu_vexriscv"]
+	shallow = true
+[submodule "litex_cpu_rocket"]
+	shallow = true
+[submodule "litex_cpu_microwatt"]
+	shallow = true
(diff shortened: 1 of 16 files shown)


enjoy-digital:
Closing, as discussed in https://github.com/enjoy-digital/litex/issues/394, we'll use a better solution with Python data packages replacing the submodules.

PR #397, opened 2020-02-21
Add 'volatile' qualifier to new CSR accessors

gsomlo:
Through their use of the MMPTR() macro, the "classic"
csr_[read|write]simple() accsessors identify the MMIO
subregister with the 'volatile' qualifier.

Adjust the new, csr_[rd|wr]_uint[8|16|32|64]() accessors
to also utilize the 'volatile' qualifier. Since accesses
are implicit (a[i], where a is an 'unsigned long *'),
change 'a' to be a 'volatile unsigned long *' instead.

No difference was noticed in opcodes generated using the
gcc9 risc-v cross-compiler on x86_64 with standard LiteX
cflags (vexriscv and rocket were tested), but since
reports exist that 'volatile' matters on some combinations
of compilers and targets, add the 'volatile' qualifier just
to be on the safe side.

Signed-off-by: Gabriel Somlo <gsomlo@gmail.com

### DIFF
--- a/litex/soc/software/include/hw/common.h
+++ b/litex/soc/software/include/hw/common.h
@@ -70,7 +70,7 @@ static inline int num_subregs(int csr_bytes)
 }
 
 /* Read a CSR of size 'csr_bytes' located at address 'a'. */
-static inline uint64_t _csr_rd(unsigned long *a, int csr_bytes)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks!

PR #399, opened 2020-02-23
Converting LiteX to use Python modules.

mithro:
This is the start of converting LiteX to use the `litex-data-XXX` python modules. 

It **hasn't** been tested and probably **doesn't** work yet, but wanted you to see the direction it is going.


mithro:
@enjoy-digital - I can now compiler an arty bitstream using this pull request.

mithro:
@enjoy-digital - I still need to fix the `litex-data-misc-tapcfg` module before I can merge....

mithro:
@xobs - can you check this works with your lxbuildenv.py stuff?

litex/data/find.py
```diff
@@ -0,0 +1,15 @@
+def find_data(data_type, data_name):
+    imp = "from litex.data.{} import {} as dm".format(data_type, data_name)
+    try:
+        l = {}
+        exec(imp, {}, l)
+        dm = l['dm']
+        return dm.data_location
+    except ImportError as e:
+        raise ImportError("""\
+litex-data-{dt}-{dn} module not install! Unable to use {dn} {dt}.
```
xobs: should this be "module not installED?

mithro:
@enjoy-digital / @xobs - I finally got back to this pull request today. I've rebased it onto the latest master.

enjoy-digital:
@mithro: Thanks. Do you still have things you want to do on this or do you think it's ready? I could start doing some tests with it, see if everything is behaving correctly and report the results here if you think it's ready.

litex/data/find.py
mithro: Fixed.

mithro:
@enjoy-digital - Could you do some tests and see if it is working for you?

mithro:
Ping?

enjoy-digital:
@mithro: sorry, i've been able to do a quick test, i'm able to install it and build targets with it, but i see two things:

1. When installing LiteX, we get this:
````
Searching for litex-data-software-compiler_rt
Reading https://pypi.org/simple/litex-data-software-compiler_rt/
Reading https://pypi.org/simple/litex-data-software-compiler-rt/
Couldn't find index page for 'litex-data-software-compiler_rt' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.org/simple/
No local packages or working download links found for litex-data-software-compiler_rt
error: Could not find suitable distribution for Requirement.parse('litex-data-software-compiler_rt')
````
But that's just probably the package needs to be uploaded to pypi.

2. LiteX simulator is not working due to an issue with tapcfg: the #include "tapcfg.h" is failing.

i was just a quick test, i could spend more time on it next week.

**Comment by enjoy-digital** [comment]
@mithro:  i'm not sure to understand why the paths in the packages need to be that complex: this is confusing when looking at the package since it's difficult to know where the data is really located and this also create a dependency with the way LiteX organize its data, which is what we are trying to avoid on others topics.

If we take lm32 for example, the path is `litex/data/cpu/lm32/verilog`, couldn't it be just `data`  and in LiteX this data would made available in `litex/data/cpu_lm32/`? Is it possible with the python packages?

**Comment by mithro** [comment]
We could use `litex/data/cpu_lm32/verilog` rather than `litex/data/cpu/lm32/verilog`? 

The data can't just be in directory directly as the `__init__.py` file needs to be in a separate directory to where the data is merged too.

I think having separate data directories for each of the "types" of data is potentially useful if data modules grow to understand their contents more. (IE The CPU data modules could potentially bit split via variant or similar in the future?).

**Comment by mithro** [comment]
FYI - I've been working on making travis push the litex-data modules to PyPi. Currently just using the test.pypi.org index.

**Comment by enjoy-digital** [comment]
@mithro: ok thanks. Then we can keep your initial structure, but just remove the `verilog`. (similarly for the `verilog` or `sources` used by the others CPU) This was useful when the data was in the same directory as the core.py but this is not longer necessary here since the data are in their own directory already. This allow keeping your initial structure but simplifies the path a bit.

**Comment by mithro** [comment]
@enjoy-digital - I'm not sure I understand what change you are suggesting?

**Comment by mithro** [comment]
@enjoy-digital - What command are you running to test the litex simulator?

**Comment by enjoy-digital** [comment]
@mithro: instead of using `litex/data/cpu/lm32/verilog` we can just use `litex/data/cpu/lm32`  and more generally remove the `verilog` or `sources` of the paths.

For LiteX simulator, once LiteX is installed just run `litex_sim` or just execute this script: https://github.com/enjoy-digital/litex/blob/master/litex/tools/litex_sim.py

**Comment by mithro** [comment]
@enjoy-digital We need the `__init__.py` for the data module to be in a separate directory to the data contents. The [`litex/data/cpu/lm32/__init__.py`](https://github.com/litex-hub/litex-data-cpu-lm32/blob/master/litex/data/cpu/lm32/__init__.py) verse [`litex/data/cpu/lm32/verilog`](https://github.com/litex-hub/litex-data-cpu-lm32/tree/master/litex/data/cpu/lm32/verilog).

**Comment by mithro** [comment]
@enjoy-digital -- That also helps with the linguist data classification -> https://github.com/litex-hub/litex-data-cpu-lm32/blob/master/.gitattributes

**Comment by mithro** [comment]
I guess we could change the `litex/data/cpu/lm32/__init__.py` file to `litex/data/cpu/lm32.py` and try put the data in a `litex/data/cpu/lm32` directory -- but I'm unsure if Python will like that?

**Comment by enjoy-digital** [comment]
OK i see, but just to understand, do we need to have the same path in the Python package than in LiteX? If not, we could simplify the path in the package. If that's not possible, we could at least use a name that is more generic and work for all kind of files, for example `submodule`, `sources`, etc...

**Comment by mithro** [comment]
The data files need to be "under the module". I originally started with the generic name "contents" but switched to using verilog / vhdl / system_verilog / software names to be more descriptive about what the data was.

**Comment by mithro** [comment]
After discussion with @enjoy-digital we changed some things up.

The data containing Python modules will no longer be inside the LiteX name space. Instead they will be standalone and name `pythondata-XXX-YYY`. This means that;
 - We don't need to worry about namespace packages.
 - The data files are much closer to the root of the repository.
 - The packages can be reused by other Python systems.

I have updated the modules -> https://github.com/litex-hub?q=pythondata-&type=&language=

They have been uploaded from Travis-CI to the test.pypi.org repository at https://test.pypi.org/search/?q=pythondata-

**Comment by mithro** [comment]
@xobs and @enjoy-digital - The pull request should be ready to test again.

Only thing which I haven't had the chance to look into and fix is the tapcfg / sim example.

**Comment by xobs** [comment]
I finally got around to trying it, and I like what I see.

When I checked out the branch from @mithro and tried building, I received the following error:

```
INFO:SoC:IRQ Handler (up to 32 Locations).
IRQ Locations: (2)
- timer0 : 2
- usb    : 3
INFO:SoC:--------------------------------------------------------------------------------
  File "D:\Code\Fomu\foboot\hw\deps\litex\litex\__init__.py", line 18, in get_data_mod
    exec(imp, {}, l)
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'pythondata_software_compiler_rt'

During handling of the above exception, another exception occurred:

  File ".\foboot-bitstream.py", line 476, in <module>
    main()
  File ".\foboot-bitstream.py", line 450, in main
    vns = builder.build()
  File "D:\Code\Fomu\foboot\hw\deps\litex\litex\soc\integration\builder.py", line 195, in build
    self._generate_includes()
  File "D:\Code\Fomu\foboot\hw\deps\litex\litex\soc\integration\builder.py", line 106, in _generate_includes
    get_data_mod("software", "compiler_rt").data_location)
  File "D:\Code\Fomu\foboot\hw\deps\litex\litex\__init__.py", line 28, in get_data_mod
    """.format(dt=data_type, dn=data_name, e=e))
ImportError: pythondata-software-compiler_rt module not installed! Unable to use compiler_rt software.
No module named 'pythondata_software_compiler_rt'

You can install this by running;
 pip install git+https://github.com/litex-hub/pythondata-software-compiler_rt.git
$
```

This is a good error because it gives a hint as to what the problem is.  My solution was not to do `pip install`, but rather `cd deps; git submodule add https://github.com/litex-hub/pythondata-software-compiler_rt.git; cd ..`.

That is the only change I had to make, which resulted in a working build.

mithro:
I didn't get to looking at the tagcfg issue today. Will try and look at it over the weekend.

mithro:
This change is now dependent on https://github.com/enjoy-digital/litex/pull/464

enjoy-digital:
@mithro: thanks. Do you still think #464 is WIP or could it be merged?

mithro:
Does #464 work for you? I didn't have high confidence it would?

enjoy-digital:
@mithro: #464 seems fine yes, we don't have large dependencies on build/sim, so we could still do some changes after if needed.

enjoy-digital:
@mithro, @xobs: we can probably merge now if this is ok for you? Users will probably have to re-install LiteX with the litex_setup.py script or update their installation script, but if it's not now it will be later..., so if you think this is the right time, we could merge.

xobs:
Agreed.  Let's merge.

xobs:
As this is a big change for existing users, is it possible to add a note to the README?  Would it make sense to tag this release as well, so we know exactly when this change took place?

enjoy-digital:
Thanks @xobs, yes i was thinking about that too. I will create a first release with this and a note in the README.

enjoy-digital:
I just created a first release and added a note for the switch to Python modules in the README with https://github.com/enjoy-digital/litex/commit/56aa7897df99d7ad68ea537ab096c3abdc683666.

PR #400, opened 2020-02-24
Fix ECP5PLL VCO frequency range

Xiretza:
See [ECP5 and ECP5-5G Family Data Sheet](https://www.latticesemi.com/view_document?document_id=50461), section 3.19 "sysCLOCK PLL Timing".


enjoy-digital:
Thanks!

PR #401, opened 2020-02-25
software: enable link time optimization (LTO)

mateusz-holenko:
It turned out that after merging https://github.com/enjoy-digital/litex/commit/ff2775c264cc452a05119dd895e25b559867dac2 to LiteX, `-flto` works fine with `-Os` and does not require `-lgcc` anymore. That's why we are going back to the simplest solution, originally proposed by @mithro and discussed in: https://github.com/enjoy-digital/litex/pull/253.

Note: enabling LTO in LiteX requires enabling it in linux-on-litex-vexriscv as well. We have a patch prepared and will create a PR shortly.


enjoy-digital:
Thanks, merged. I disabled it for LM32 with https://github.com/enjoy-digital/litex/commit/e2aebb427ece793537e64f28cc3193704e7be782 since old versions of GCC easily available don't support it.

PR #402, opened 2020-02-25
tools: litex_gen: fix missing UART pins

kowalewskijan:
These changes fix following error:
`raise ConstraintError("Resource not found: {}:{}".format(name, number))
litex.build.generic_platform.ConstraintError: Resource not found: serial:None`

When calling:
`python3 litex_gen.py --with-uart --no-compile-gateware --no-compile-software`


enjoy-digital:
Thanks!

PR #403, opened 2020-02-28
Add option to implement UART without FIFOs

kowalewskijan:
These changes introduce optional UART implementation without FIFOs and enables this option in litex_gen tool.


enjoy-digital:
Thanks. To minimize code changes and duplication, its has been implemented differently using a FIFO of minimal depth. A `uart-fifo-depth` parameter has been added `litex_gen` can now be called with `--uart-fifo-deph=1` to minimize FIFO depth. `--uart-fifo-deph=0` is not working for RX, but i'll look at that.

PR #405, opened 2020-03-02
add riscv-sifive-elf triple

sajattack:
The AUR (Arch User Repository) [package](https://aur.archlinux.org/packages/riscv-sifive-elf-gcc/) for the sifive toolchain uses the triple `riscv-sifive-elf`. This PR adds this triple to the relevant cpu core.py files.


enjoy-digital:
Thanks.

PR #408, opened 2020-03-04
targets/nexys4ddr: fix sdcard clocker initialization

gsomlo:
Following the new clocker from https://github.com/enjoy-digital/litesdcard/commit/2d4230c5e439f114d4fcb0c64c828d996fb5cb20 this is my attempt to get the nexys4ddr to successfully build using `--with-sdcard` once again. Based on my experiments, one would now have to go with at least
```
sdclk 20
sdinit
sdtest 8
```
from the LiteX bios command line. Anything below 20 and initialization fails for me (rocket cpu at `sys_clk_freq=67e6` -- the maximum that will pass timing for rocket when `--with-ethernet` is also selected).

@enjoy-digital PTAL at the new `add_period_constraint()` statements in particular, as I'm not sure `period_ns(self.sys_clk_freq))` is the right value for the constraints (in fact I'm not sure I understand how that part works at all, so if you can explain or point me to where I can figure it out more easily, I'd appreciate that too :) )

### DIFF
--- a/litex/boards/targets/nexys4ddr.py
+++ b/litex/boards/targets/nexys4ddr.py
@@ -34,11 +34,8 @@ class _CRG(Module):
         self.clock_domains.cd_sys2x_dqs = ClockDomain(reset_less=True)
         self.clock_domains.cd_clk200    = ClockDomain()
         self.clock_domains.cd_eth       = ClockDomain()
-        self.clock_domains.cd_sdcard    = ClockDomain(reset_less=True)
 
         # # #
-
-        self.sd_clk_freq = int(100e6)
 
         self.submodules.pll = pll = S7MMCM(speedgrade=-1)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks. The PR is fine and we can use the `add_period_constraint()` you added.  Initially, the SD Card clock is initialized to 10MHz, but we change it dynamically to increase it after initialization has been done so we need to the constraints to the maximum frequency we are going to use. Here using `sys_clk_freq` seems fine.

It's possible to the S7MMCM is configuring the MMCM multipliers/dividers differently than what we were doing previously in SDClockerS7. So the current update we are doing on the dividers is probably not correct, which would explain why it only works with `sdclk 20` We should probably reuse similar code than what is used by Antmicro for the Linux driver in the BIOS. (I will look at that).

PR #410, opened 2020-03-05
platform/netv2: add proper I2C pins for HDMI IN0

piotr-binkowski:
This PR adds `SDA_PU` and `SDA_PD` pins to NeTV2 platform and inverts `SCL` and `SDA` pins. With this and a small modification to LiteVideo it is possible to get EDID information from the board.


enjoy-digital:
Thanks, merged.

Issue #346, opened 2020-01-15
Support for multiple DRAMs in SoCSDRAM


kgugala:
We've been working on adding Litex support for Pano Logic G2 board (https://github.com/tomverbeure/panologic-g2) we've added basic support in Litex-buildenv (https://github.com/timvideos/litex-buildenv/pull/223).

The board has 2 DDRAM chips connected to separate banks. To handle this we should instantiate two DDRAM controllers, but current SoCSDRAM implementation assumes there is single DDRAM PHY in the system.

CC @mithro

mithro:
@kgugala You should just be able to use the non-`SoCSDRAM` and just instantiate the block twice?

@enjoy-digital Might have a suggestion on how to make the 2 x DDR chips driven via a single DDR controller?

enjoy-digital:
@mithro: i've been thinking for some time about having a single SoC class that could cover SoCCore/SoCSDRAM needs and that would be more flexible and would remove this kind of limitations. I'll prototype something.

enjoy-digital:
When we have multiple DRAM chips connected to the board using separate command lines we have the choice to:
1. Have 2 PHYs and a single Controller, DFI commands will just be replicated on each PHY and the only change for the Controller and system is that the data-width/capacity/bandwidth is doubled.
2. Have a Controller per PHY that will be mapped at a different location and each Controller/PHY will be fully independent from the other.

Both solutions can have advantages/disadvantages depending the use case.

To support 1. we could create a wrapper that will encapsulate the 2 PHYs and create a single DFI interface from the 2 DFIs of each PHY.

To support 2. we could use the new SoC class that has a `add_sdram` method that we could use to add the 2 Controllers, but this would still need some changes to the software to do the initialization of the 2 Controllers. The new `add_sdram ` method is currently prototyped on the nexys4ddr target:
Nexys4DDR targets: https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/nexys4ddr.py#L74

enjoy-digital:
@mglb submitted https://github.com/enjoy-digital/litex/pull/411 and https://github.com/enjoy-digital/litedram/pull/157 just after i wrote my previous comment. I'm a bit uncomfortable with this, because i realize how much work it is, but i'm not sure we have a real use case for it. From my previous comment, on the Pano Logic G2 we clearly want to implement solution 1. while it is doing so implementing 2.

[...]

PR #411, opened 2020-03-05
Support for multiple SDRAM PHYs in single SoC

mglb:
Requires https://github.com/enjoy-digital/litedram/pull/157

Summary:
* Multiple SDRAM PHYs can be added with `LiteXSoC.add_sdram(...)`:
  * Each SDRAM can be added as part of `main_ram` or as a fully independent region (pass `main_ram = False` to `add_sdram()`).
  * Each SDRAM has a separate L2 cache.
* Generated headers contain additional code:
  * An array of structs with information about each SDRAM has been added to `sdram_phy.h`. Each entry contains properties and pointers to functions used by BIOS code.
  * An array with SDRAM regions and L2 cache sizes has been added to `mem.h`.
  * `CONFIG_L2_SIZE` has been replaced by` CONFIG_L2`. The former was not removed for compatibility with older code.
  * New headers are backward compatible - old code still works.
* Software/BIOS:
  * All SDRAMs are initialized and tested on boot.
  * `memtest` optionally takes 2 arguments: base address and size in bytes.
  * New command for listing SDRAM regions: `meminfo`.
  * New command for choosing the SDRAM on which low-level functions (like `sdrhw` or` sdrrdbuf`) operate - `currentsdramphy <phy_number>`.
  * `flush_l2_cache` flushes all L2 caches.
* Support for multiple SDRAMs in `litex_sim` with `--add-sdram key=val [key=val ...]` command line argument:
  * Available keys: `module`, `data-width`, `address`, `max-size`, `region-name`, `verbosity`, `init`.
  * Can be used with `--with-sdram` (and other old sdram-related arguments) or without it.

Known limitations:
* Only SDRAMPHYModel and S6HalfRateDDRPHY are supported

Todo:
* Tests/CI
* Some cleanup

I'm testing the code with litex_sim and on Pano Logic G2 (litex-buildenv).

Example litex_sim command:
```
litex_sim \
    --add-sdram module=IS42S16320  address=0x10000000 \
    --add-sdram module=MT48LC16M16 address=0x18000000 \
    --add-sdram module=MT48LC16M16 address=0x1c000000 \
```

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -128,12 +128,16 @@ class Builder:
             cpu_interface.get_git_header()
         )
 
-        if hasattr(self.soc, "sdram"):
+        if hasattr(self.soc, "sdrams") and len(self.soc.sdrams) > 0:
             from litedram.init import get_sdram_phy_c_header
+            sdram_settings = [
+                (sdram.controller.settings.phy,
+                 sdram.controller.settings.timing,
+                 name)
+                for name,sdram in self.soc.sdrams.items()
+            ]
             write_to_file(os.path.join(self.generated_dir, "sdram_phy.h"),
-                get_sdram_phy_c_header(
-                    self.soc.sdram.controller.settings.phy,
-                    self.soc.sdram.controller.settings.timing))
+                          get_sdram_phy_c_header(sdram_settings))
 
     def _generate_csr_map(self, csr_json=None, csr_csv=None):
         if csr_json is not None:
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -117,6 +117,27 @@ def get_mem_header(regions):
     for name, region in regions.items():
         r += "#define {name}_BASE 0x{base:08x}L\n#define {name}_SIZE 0x{size:08x}\n\n".format(
             name=name.upper(), base=region.origin, size=region.length)
+
+    r += "\n"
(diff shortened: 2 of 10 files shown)


enjoy-digital:
Thanks for this. This is a huge work that enhances the capabilities, but i'm also wondering if it's does not complexify too much the code for the actual use-case it's trying to cover. I suggest we discuss in https://github.com/enjoy-digital/litex/issues/346 before going further.

mithro:
Hi @mglb, I believe this PR can be reworked to be significantly smaller and less invasive, however it is unlikely I'll get a chance to provide proper suggestions until early next week.

mglb:
@mithro like, first commit (f0ada2ef7edba04bb5db52aaac6992fb7c3354de) + https://github.com/enjoy-digital/litedram/pull/157/commits/069d89c849d883837832dff0924951795b511fe1 + https://github.com/enjoy-digital/litedram/pull/157/commits/a9add8e23dce68c1c2c87d2b75ce642397e545a5? All other commits add support for multiple PHYs in BIOS or make it possible to write generic software with support for arbitrary number of PHYs.

enjoy-digital:
Closing, we'll continue the discussion in: https://github.com/enjoy-digital/litex/issues/346. In the future I would appreciate being able to discuss these features before too much time is spent in the implementation, since here it was probably too early for the project and could probably have been worked around for the Pano Logic G2.

PR #416, opened 2020-03-06
Add SVD export capability to Builder (csr_svd parameter) and targets …

enjoy-digital:
…(--csr-svd argument) and fix svd regression.

This allows generating SVD export files during the build as we are already doing for .csv or .json.

Use with Builder:
builder = Builder(soc, csr_svd="csr.svd")

Use with target:
./arty.py --csr-svd=csr.svd

### DIFF
--- a/litex/soc/doc/__init__.py
+++ b/litex/soc/doc/__init__.py
@@ -32,7 +32,7 @@ def generate_svd(soc, buildpath, filename=None, name="soc", **kwargs):
         filename = name + ".svd"
     kwargs["name"] = name
     with open(buildpath + "/" + filename, "w", encoding="utf-8") as svd:
-        svd.write(export.get_svd(soc, **kwargs))
+        svd.write(export.get_csr_svd(soc, **kwargs))
 
 
 def generate_docs(soc, base_dir,
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -152,12 +152,12 @@ class Builder:
             os.makedirs(svd_dir, exist_ok=True)
             write_to_file(self.csr_svd, export.get_csr_svd(self.soc))
 
-    def _prepare_software(self):
+    def _prepare_rom_software(self):
         for name, src_dir in self.software_packages:
             dst_dir = os.path.join(self.software_dir, name)
             os.makedirs(dst_dir, exist_ok=True)
 
-    def _generate_software(self, compile_bios=True):
+    def _generate_rom_software(self, compile_bios=True):
          for name, src_dir in self.software_packages:
             if name == "bios" and not compile_bios:
                 pass
@@ -167,7 +167,7 @@ class Builder:
                 if self.compile_software:
                     subprocess.check_call(["make", "-C", dst_dir, "-f", makefile])
 
-    def _initialize_rom(self):
+    def _initialize_rom_software(self):
         bios_file = os.path.join(self.software_dir, "bios", "bios.bin")
         bios_data = soc_core.get_mem_data(bios_file, self.soc.cpu.endianness)
         self.soc.initialize_rom(bios_data)
@@ -182,11 +182,11 @@ class Builder:
         self._generate_includes()
         self._generate_csr_map()
         if self.soc.cpu_type is not None:
-            self._prepare_software()
-            self._generate_software(not self.soc.integrated_rom_initialized)
+            self._prepare_rom_software()
+            self._generate_rom_software(not self.soc.integrated_rom_initialized)
             if self.soc.integrated_rom_size and self.compile_software:
                 if not self.soc.integrated_rom_initialized:
-                    self._initialize_rom()
+                    self._initialize_rom_software()
 
         if "run" not in kwargs:
             kwargs["run"] = self.compile_gateware
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -14,7 +14,7 @@ import struct
 import shutil
 
 from litex.build.tools import write_to_file
-from litex.soc.integration import cpu_interface, soc_core
+from litex.soc.integration import export, soc_core
 
 __all__ = ["soc_software_packages", "soc_directory",
            "Builder", "builder_args", "builder_argdict"]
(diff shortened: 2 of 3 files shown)


bunnie:
I might not be doing this right, but...I tried doing the build like this:

```
    platform = Platform(dutio)
    soc = Dut(platform)

    builder = Builder(soc, output_dir="./run", csr_svd='../../target/soc.svd')
    BiosHelper(soc)

    builder.software_packages = [
        ("bios", os.path.abspath(os.path.join(os.path.dirname(__file__), "test")))
    ]
    vns = builder.build(run=False)
    soc.do_exit(vns)
```

This is for a CI system I'm putting together now, so I'm not running gateware build, just going straight to simulation.

When I tried doing this, the soc.svd file is degenerate:
```
<?xml version="1.0" encoding="utf-8"?>

<device schemaVersion="1.1" xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="CMSIS-SVD.xsd" >
    <vendor>betrusted.io</vendor>
    <name>SIMULATION</name>
    <description><![CDATA[simulation core framework]]></description>

    <addressUnitBits>8</addressUnitBits>
    <width>32</width>
    <size>32</size>
    <access>read-write</access>
    <resetValue>0x00000000</resetValue>
    <resetMask>0xFFFFFFFF</resetMask>

    <peripherals>
    </peripherals>
</device>
```

The order of events needs to be:

1. make the builder
2. build the BIOS
3. call builder.build()

If I change the order events to

1. make the builder
2. call builder.build()
3. build the BIOS

I get the following error:

`FileNotFoundError: [Errno 2] No such file or directory: '/home/bunnie/code/betrusted-soc/deps/gateware/sim/memlcd/run/software/bios/bios.bin'`

So it seems adding this parameter doesn't quite solve the use case I was worried about.

That being said, I'm OK doing a two-step build, which I'm doing right now successfully:

https://github.com/betrusted-io/gateware/blob/81098604b58ddf5ccae0ef81bacdd3417e2a4468/sim/template/dut.py#L111

This is more of a "nice to have" feature, it doesn't break anything for me to be missing it.

bunnie:
Also, for my mainline build, I can't seem to do both a csv and an svd file:

```
    builder = Builder(soc, output_dir="build", csr_svd="build/software/soc.svd", csr_csv="build/software/csr.csv", compile_software=compile_software, compile_gateware=compile_gateware)
```
This will fail with

```
  File "/home/bunnie/code/betrusted-soc/betrusted-soc.py", line 1245, in <module>
    ret = main()
  File "/home/bunnie/code/betrusted-soc/betrusted-soc.py", line 1206, in main
    vns = builder.build()
  File "/home/bunnie/code/betrusted-soc/deps/litex/litex/soc/integration/builder.py", line 190, in build
    self._generate_csr_map()
  File "/home/bunnie/code/betrusted-soc/deps/litex/litex/soc/integration/builder.py", line 147, in _generate_csr_map
    os.makedirs(self.csr_dir, exist_ok=True)
AttributeError: 'Builder' object has no attribute 'csr_dir'
```

Whereas if I just specify a csr_svd= target, the build does succeed.

In this case, the BIOS is a degenerate C stub at location 0x0 that jumps to our loader in SPI area, so we don't need to do a two-step build because the entry point is based on a "native" Litex Makefile flow.

bunnie:
Just some background, we want both because:

* litescope needs csr.csv, as well as I think wishbone-tool
* Our Rust build.rs flow wants the soc.svd

Generating our own code APIs from the SVD file helps to ensure stability in our codebases, allowing Litex firmware to continue its evolution in a Linux-directed path without us holding it back. But we still need the csr.csv file to interoperate with the other LiteX suite of tools.

However, my legacy flow (like NeTV2) will be reliant on the C Litex stuff, but...hard to say when if ever I will get a chance to bring that up to HEAD for Litex, I think it may forever sit on an older version of Litex.

enjoy-digital:
I just fixed the `csr_dir` issue and force-pushed. I'm testing by going to boards/targets and doing:
````
./arty.py --csr-csv=csr.csv --csr-svd=csr.svd
````
I will now see if we can fix the others points.

enjoy-digital:
With https://github.com/enjoy-digital/litex/pull/416/commits/69ffafd81d779deceb1f81089514b15fa26d285d we now generate the .json/.csv/.svd before compiling the software.

xobs:
With the changes, I'm able to build, and I can verify that the `soc.svd` file appears before the build starts.

enjoy-digital:
@xobs: thanks for the feedback.
@bunnie: for your usecase, how about having your customized Builder for Betrusted?:

````python3
class BetrustedBuilder(Builder):
    def _prepare_rom_software(self):
    	pass
    def _generate_rom_software(self, compile_bios=True):
    	BiosHelper(self.soc)
    def _initialize_rom_software(self):
        bios_file = ???
        bios_data = soc_core.get_mem_data(bios_file, self.soc.cpu.endianness)
        self.soc.initialize_rom(bios_data)

platform = Platform(dutio)
soc      = Dut(platform)

builder = BetrusterBuilder(soc, output_dir="./run", csr_svd='../../target/soc.svd')
builder.build()
````
`_prepare_rom_software`, `_generate_rom_software` and `_initialize_rom_software`  will now be called after the `.svd` has been generated, so you could compile your software in these calls and avoid the two-step build.

bunnie:
I /think/ I follow what you're suggesting. Generally, that flow is useful for people who want to build the BIOS using a non-C language, so perhaps naming it ExternalBiosBuilder? ForeignBiosBuilder? ImportedBiosBuilder? AlternativeBuilder?

Anyways, i think it looks like you're proposing a flow where we can pass a `BiosHelper()` method to an alternative builder, and the `BiosHelper` is invoked after the SVD is created but before the `bios.bin` file is pulled in for initialization. 

Maybe the `BiosHelper()` can return the path to the `bios.bin` or even return an object that is the binary itself...so the full method signature would return either the path to the bios file or a byte array that is the bios itself, and it would take as a common argument the path to the SVD file. 

Something like that could work! However, I would re-iterate that I'm quite OK with rebuilding the SoC twice; even though it's not elegant, the whole build process is so fast that if the first iteration didn't print anything I would never notice it. 

So I'd say my request is pretty low priority but just something to keep in view in case the number of Rust+Litex users grows.

enjoy-digital:
@xobs, @bunnie Thanks for the feedback, i'm now merging since all points should be covered.

PR #419, opened 2020-03-09
software/bios: fixup for Ultrascale SDRAM debug

gsomlo:
Keep CSR accesses independent of csr_data_width and csr_alignment.

@enjoy-digital Please test before applying -- I'm _guessing_ the assumption was that  alignment is 32 and csr-data-width is 8, but I'm not 100% sure :)


enjoy-digital:
Thanks @gsomlo, i've been able to test it on hardware and it's working fine.

PR #421, opened 2020-03-10
add fractional division options to clk0 config on PLL

bunnie:
S7 MMCMs allow fractional divider on clock 0. Add a fallback
to try fractional values on clock 0 if a solution can't be found.

This is necessary for e.g. generating both a 100MHz and 48MHz
clock from a 12MHz source with margin=0


enjoy-digital:
Thanks, that will be very useful. I'll look at that soon (and will also make sure we only enalbe it for the MMCM).

Issue #420, opened 2020-03-09
LiteXSoC.add_ethernet() vs. mem_map["ethmac"]


gsomlo:
For better or worse (and I'm happy to stop doing it, if it was a bad idea to begin with) I've been hardcoding `mem_map["ethmac"]` to `0x3000_0000` in https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/rocket/core.py#L81. The "classic" (inheritance-based) `EthernetSoC` uses `mem_map.update()` to set `mem_map["ethmac"]` to a default (`0xb000_000`) unless already set (see https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/versa_ecp5.py#L103 or https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/trellisboard.py#L109). That works as expected, and the hard-coded `0x3000_0000` from `rocket/core.py` stays in place.

This changed recently for nexys4ddr (https://github.com/enjoy-digital/litex/blob/master/litex/boards/targets/nexys4ddr.py#L81) which now uses the new `LiteXSoC.add_ethernet()`, which in turn grabs an automatically selected `SoCRegion` in https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L1045.

So, on nexys4ddr I get `memory_region,ethmac,0x10008000,8192,io` instead of `0x30000000` as hardcoded in `rocket/core.py`.

I'd like to switch all targets to using `LiteXSoC.add_ethernet()` (we _should_ factor out Ethernet instantiation rather than have the same code replicated for each target), but first the question is whether we want to support hard-coding region base addresses, and, if so, patch `LiteXSoC.add_ethernet()` to check if `mem_map["ethmac"]` exists, and, if yes, use it (i.e., as the `origin` parameter to `SoCRegion` (https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L57).

enjoy-digital:
Just for info, the idea with the new SoC class is to:
1) Allow the user to defined the Memory Map and use it when specified.
2) Let the SoC determines the Memory Map if no specified or partially.
That was already possible for the CSRs and the idea is to extends is to the Memory Map.

I think the default in LiteX for the Ethernet MAC mapping should be to determine it automatically since:
- in most of the cases, the BIOS is compiled for a specific SoC.
- we are able to pass device tree to Zephyr and Linux.
[...]

PR #423, opened 2020-03-11
integration/soc: add_ethernet: honor self.map["ethmac"], if present

gsomlo:
Can't (yet) stop setting `self.map["ethmac"]` in Rocket, because doing so would break the build on targets _other_ than nexys4ddr (which don't yet use `LiteXSoC.add_ethernet()`).

But at least for now they're consistent across targets.

This (partially) fixes https://github.com/enjoy-digital/litex/issues/420.


enjoy-digital:
Thanks.

PR #424, opened 2020-03-12
Add --mr-memory-x parameter to generate memory regions memory.x file

esden:
This file is used by rust embedded target pac crates.

It is essentially a linker file but has some specific entries needed by the rust runtime. So we are mostly reusing the linker script generator for this and adding a few of our own entries.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -5,6 +5,7 @@
 # This file is Copyright (c) 2016-2017 Tim 'mithro' Ansell <mithro@mithis.com>
 # This file is Copyright (c) 2018 William D. Jones <thor0505@comcast.net>
 # This file is Copyright (c) 2020 Xiretza <xiretza@xiretza.xyz>
+# This file is Copyright (c) 2020 Piotr Esden-Tempski <piotr@esden.net>
 # License: BSD
 
 
@@ -46,7 +47,8 @@ class Builder:
         compile_gateware        = True,
         csr_json                = None,
         csr_csv                 = None,
-        csr_svd                 = None):
+        csr_svd                 = None,
+        mr_memory_x             = None):
         self.soc = soc
 
         # From Python doc: makedirs() will become confused if the path
@@ -62,6 +64,7 @@ class Builder:
         self.csr_csv  = csr_csv
         self.csr_json = csr_json
         self.csr_svd  = csr_svd
+        self.mr_memory_x = mr_memory_x
 
         self.software_packages = []
         for name in soc_software_packages:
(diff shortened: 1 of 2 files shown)


mithro:
@xobs - As you are one of the other people doing Rust stuff, could you review this?

xobs:
I think it's a good first step, and definitely helps us build Rust binaries more easily.

If I may bikeshed a bit: My concern with this has to do with how region names are identified.  In the patch, text is defined to be in a region called `spiflash`, but what if you use a Rust binary as the boot ROM and you want the text section to go into a region called `rom`?  Or what if you're writing for a large FPGA with external RAM, and you want the data section to go into `main_ram` instead of `sram`?

esden:
@xobs that is a good question. But I think we should address them when someone is faced with those specific scenarios, and add the necessary config/modification parameters to generate the memory.x file as needed. It is always easier to see what needs to be adapted when faced with a specific scenario. I am aware that the solution I submitted is pretty limited to a specific situation. In the future we could add a parameter to pass what section should be used for text, or even make it fully configurable by passing a dictionary with the REGION_ALIAS list.

xobs:
Sounds good to me.  We can make those changes when it actually comes time to start writing the bios in Rust.

enjoy-digital:
Thanks @esden. You approach is similar to what is generally done in the project: provide a minimal implementation for the actual need, but allow it to evolves in the future to cover more cases, so i'm also perfectly fine with that. About the name of the parameter, what about renaming `mr_memory_x` to` rust_memory_x`? This would make it clear that this parameter is Rust specific (and i don't think memory.x files are used in others language?)

xobs:
I think `memory.x` is a common term in certain areas, and has been adopted by Rust.  For example, [synopsis uses it in their ARC chips](https://embarc.org/toolchain/baremetal/linker.html), as does [GNU for the 68HC11](https://www.gnu.org/software/m68hc11/m68hc11_faq.html), and I see mostly error reports about [MSP430 using it](https://forum.43oh.com/topic/3406-compiler-error-cannot-open-linker-script-file-memoryx-no-such-file-or-directory/).

enjoy-digital:
@xobs: thanks, then what about just `memory_x`  instead  of `mr_memory_x`? This way this is consistent with others parameters:  `csr_json` basename will generally be `csr.json`,  `csr_csv` will generally be `csr.csv`, etc... and  `memory_x` basename will generally be `memory.x`. (Sorry for being more careful on this, but this is what is going to break if we change it later).

enjoy-digital:
That's merged with just the renaming of the parameter with https://github.com/enjoy-digital/litex/commit/a27385a79c080c40cb80b0d39611a48821e10736. Thanks.

PR #425, opened 2020-03-13
Make CSR documentation diagrams, with more than 8 bits, be split into multiple lanes.

esden:
In cases when each CSR bit has a name and we use CSR with more than 8
bits, the register diagram quickly becomes crowded and hard to read.

With this patch we split the register into multiple lanes of 8 bits
each.

The motivation for this patch is motivated by the [iCEBESOC documentation](https://icebreaker-fpga.github.io/icebreaker-litex-examples/leds.html#leds-out).

I might be missing some details here. Let me know if there is a better way of doing this. We might want to add a member variable to the CSR that will allow us to fine tune if we want the register to be split into 4 lanes or not. Maybe even a way to select if we want it to be split into 16bit lanes instead of 8bit.

@xobs your input is very much appreciated. :)

### DIFF
--- a/litex/soc/doc/csr.py
+++ b/litex/soc/doc/csr.py
@@ -235,7 +235,9 @@ class DocumentedCSRRegion:
         print("", file=stream)
         print("        {", file=stream)
         print("            \"reg\": [", file=stream)
+        multilane = False
         if len(reg.fields) > 0:
+            min_field_size = self.csr_data_width
             bit_offset = 0
             for field in reg.fields:
                 field_name = field.name
@@ -254,6 +256,9 @@ class DocumentedCSRRegion:
                     term=""
                 print("                {\"name\": \"" + field_name + "\",  " + type_str + attr_str + "\"bits\": " + str(field.size) + "}" + term, file=stream)
                 bit_offset = field.offset + field.size
+                min_field_size = min(min_field_size, field.size)
+                if min_field_size < 8:
+                    multilane = True
             if bit_offset != self.busword:
                 print("                {\"bits\": " + str(self.busword - bit_offset) + "}", file=stream)
         else:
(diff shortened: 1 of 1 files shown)


xobs:
I think you might want `self.busword // 8` to avoid non-integer widths.

But the bigger issue is what you get when you split large registers.  For example, in Betrusted, we have a timer that's a variable width, and currently is 64 bits wide.  Litex will split this into two 32-bit CSRs.  With `lanes=4`, the output ends up looking like this:

image

We definitely need more logic there to (1) parse the name into something more reasonable, and (2) split it up again into each individual lane.

There's also the less-urgent problem of very sparsely-populated registers:

image

but I think the first image is the more concerning one.

xobs:
One additional wrinkle: We currently display the default value below the register:

image

How should that look when a register is split across 4 lanes?

esden:
Ok, I fixed the division issue.

If I understand you correctly you would prefer if the registers had names:
* `time[31:24]`
* `time[23:16]`
* `time[15:8]`
* `time[7:0]`

I think there are two approaches to it. And we might want to have both as an option. 
1. We parse the name and generate ranges with the different names to have the correct ones.
2. a) We check how long the ranges are in the register and decide based on that how many lanes we generate. (the automatic logic for that can become pretty brittle)
  b) We check a csr doc parameter that tells us how long to make the lanes. (less brittle but not automatic)

Another wrinkle to add. We might want to combine the empty lanes for bits 31:8 into one lane, and keep the lower 8 in one lane as those are the only ones with "active ranges"... 

So I can see a lot of yak shaving opportunities here. I am not sure what we should do here. I would prefer that we don't make it too complicated and grow it into a large project. :D

xobs:
Yes, if we split it up that way I do think we should change the names so they appear that way.

We already have `split_field()` which maybe could be repurposed to do this without too much effort: https://github.com/enjoy-digital/litex/blob/master/litex/soc/doc/csr.py#L200-L229

It's still an open question where to put the `default` values, but that's kind of been an open question for a while.

esden:
Ok, I have added a little bit more logic to decide when a register should be split into multilane display. I think that we only need to split the display into miltilane when we we use fields and at least one field is less than 8 bit wide. This seems to be the moment when we start running out of space for the field names. I have created a bunch of examples. (Sidenote: I think it might be a good idea to create a project that generates documentation exercising the documentation generator, so that we can check and improve the documentation generation as we continue adding more logic)

This change does not address the init value issues or field overflow in cases when people add `[m:n]` to their field names. But this should at least cover a big range of situations more gracefully.

Let me know what you think. Following are some examples that I came up with:

**tiny fields in a 32bit register**
Screen Shot 2020-03-12 at 10 38 47 PM

**large field of the same size as the register*
Screen Shot 2020-03-13 at 2 40 02 PM

**multiple fields but all >= 8bit**
Screen Shot 2020-03-13 at 2 40 07 PM

**multiple fields but at least one < 8bit**
Screen Shot 2020-03-13 at 2 40 12 PM

**a register with no fields**
Screen Shot 2020-03-13 at 2 40 20 PM

xobs:
I like it, and it seems like a good, quick compromise, and less error-prone than re-parsing the name.

Of course, there will be problems with 9+-bit fields, but we can work on those later.

@bunnie and @enjoy-digital are the two other big users of this -- what do you think about this approach?

bunnie:
I think this looks great. We can deal with 9-bit fields later, and they are kind of rare anyways. I guess there is also the problem of, I put an enable bit in bit 0, and then stuck 8 bits on top of it? But that always rendered funky. Either way, big improvement. 👍

enjoy-digital:
Thanks @esden, it also seems good for me and we could improve later the use cases that are not yet covered.

PR #426, opened 2020-03-13
Updating the vendored wavedrom js files.

esden:
This update fixes a glitch where multi lane register diagrams used wrong colors. (I am sure there are many other updates and improvements too :D)


xobs:
Thanks for the update

PR #427, opened 2020-03-13
cores/clock: simplify Fractional Divide support on S7MMCM.

enjoy-digital:
Specific clkoutn_divide_range can now be provided by specialized XilinxClocking classes.
When provided, the specific range will be used. Floats are also now supported in the
range definition/iteration.

@bunnie: can you check this is ok and working for your use case?

### DIFF
--- a/litex/soc/cores/clock.py
+++ b/litex/soc/cores/clock.py
@@ -4,6 +4,7 @@
 
 """Clock Abstraction Modules"""
 
+import math
 import logging
 
 from migen import *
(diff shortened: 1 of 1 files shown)


bunnie:
it came out with the correct numbers when i checked out this change and used it. I'm having trouble following exactly all the magic in your code but from the little I can follow, it looks like this should work. thank you!

enjoy-digital:
Thanks for the review/check. With the code, there is  a default `clkout_divide_range` that be be overridden for specific CLKOUTs. Here for the S7MMCM, we use a specific `clkout0_divide_range` for CLKOUT0. The other change is here to allow iterating on range of integers or floats. I initialized a `test_clock` in LiteX (https://github.com/enjoy-digital/litex/commit/eb9f54b2bca60c416f929acb40c7c5f7db1335d8) that for now just covers very basic cases but i'll try to improve it.

PR #430, opened 2020-03-18
bios/sdcard: provide sdclk_set_clk() stub for clocker-less targets

gsomlo:
Targets which lack an adjustable clocker will not expose the required
registers. Provide a stub sdclk_set_clk() routine for those situations.


enjoy-digital:
Thanks, merged.

PR #431, opened 2020-03-19
litex_sim: add support for hybrid mac

piotr-binkowski:
This PR adds basic support for `hybrid` MAC mode enjoy-digital/liteeth#36, it might need some cleanup and probably UDP/IP core should be modified to also support this configuration so that it won't be necessary to manually instantiate all the cores (ARP, IP, etc.)


enjoy-digital:
Thanks a lot. I'm merging and will do eventual cleanup/changes very soon.

PR #433, opened 2020-03-20
Support SPI-mode SDCard booting on Litex+Rocket (64bit) configuration

gsomlo:
specifically targeted at (and successfully tested on) nexys4ddr, but should work on other boards as well

### DIFF
--- a/litex/boards/platforms/nexys4ddr.py
+++ b/litex/boards/platforms/nexys4ddr.py
@@ -57,6 +57,15 @@ _io = [
         IOStandard("LVCMOS33"),
     ),
 
+    ("spisdcard", 0,
+        Subsignal("rst",  Pins("E2")),
+        Subsignal("clk",  Pins("B1")),
+        Subsignal("mosi", Pins("C1"), Misc("PULLUP True")),
+        Subsignal("cs_n", Pins("D2"), Misc("PULLUP True")),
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks, i've been able to test it with Linux-on-LiteX-Vexriscv.

rob-ng15:
Just out of curiosity, are there specific unit8t, uint16t, uint32t types in
the Litex environment, would I be better off using those throughout my spi
sd card bitbanging code?

My question is regards the partition table and directory structures, as the
members in those structures are a predefined size.

If so, I'll happily provide a patch for more portability.

I'm pleased to see my patch being adapted and used by others.

Rob.

gsomlo:
> Just out of curiosity, are there specific unit8t, uint16t, uint32t types in the Litex environment, would I be better off using those throughout my spi sd card bitbanging code? My question is regards the partition table and directory structures, as the members in those structures are a predefined size. If so, I'll happily provide a patch for more portability.

https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/include/base/stdint.h#L11

Thanks,
--Gabriel

rob-ng15:
OK thanks.

I'll update and test. Got an airport run to do, so may be a couple of
hours.

rob-ng15:
Have made a patch, tested on de10nano.

https://github.com/enjoy-digital/litex/pull/434

PR #434, opened 2020-03-20
Use <stdint.h> to provide structure sizes

rob-ng15:
Point raised that my use of unsigned int/long/short could potentially raise issues on various bit length setups.

Compiled and tested on de10nano.


gsomlo:
tested OK with 64-bit Rocket/Litex on nexys4ddr and trellisboard -- thanks!

enjoy-digital:
Thanks @rob-ng15, @gsomlo, that's merged.

rob-ng15:
No worries, glad that @gsomlo pointed out this potential issue

Rob.

PR #435, opened 2020-03-20
soc/core/spi: add Clk Divider CSR (defaults to sys_clk_freq/spi_clk_f…

enjoy-digital:
…req).

Can be configured by calling spisdcard_clk_divider_write(x).

### DIFF
--- a/litex/soc/cores/spi.py
+++ b/litex/soc/cores/spi.py
@@ -25,14 +25,15 @@ class SPIMaster(Module, AutoCSR):
         self.pads       = pads
         self.data_width = data_width
 
-        self.start    = Signal()
-        self.length   = Signal(8)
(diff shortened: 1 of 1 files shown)


rob-ng15:
Just cooking dinner. So will test tomorrow.

Rob.

rob-ng15:
Tested.

After a bit of experimentation about when to reclock. Put the code into
readmbr, but before the file allocation table is read, as that is the first
big block transfer.

Have gone on the DE10nano for 12.5MHZ as played safe, code aims for 16MHZ
with a maximum of half the processor speed.

Pull request is #436.

Rob.

enjoy-digital:
Thanks for the test/feedback. I'm merging this and will merge #436 right after.

PR #436, opened 2020-03-21
Reclock spi sdcard access after initialisation

rob-ng15:
Depends upon https://github.com/enjoy-digital/litex/pull/435

After initialising the card, reclock the card, aiming for ~16MHz (divider is rounded up, as slower speed is safer), but a maximum of half of the processor speed.

Tested with the card being clocked to 12.5MHz on de10nano


enjoy-digital:
Thanks! merged.

PR #440, opened 2020-03-26
[WIP] Add initial version of the new SPI XIP core (simulation only)

piotr-binkowski:
This PR adds initial version of the new SPI XIP core, it currently consists of 2 parts:
- `LiteSPICore` that is responsible for handling the Wishbone bus, controlling the `cs_n` pin and deciding when end the SPI transfer and send new command + address
- `LiteSPIPHYModel` is a simulation model that can be used instead of the real PHY in `litex_sim`

Those two parts communicate via pair of Litex streams.
At this point the `LiteSPICore` is still missing a CSR for configuring the delay between ending a transfer and starting a new one during which `cs_n` stays high.

### DIFF
--- a/litex/soc/cores/spi_xip/common.py
+++ b/litex/soc/cores/spi_xip/common.py
@@ -1,0 +1,8 @@
+spi_phy_ctl_layout = [
+    ("addr", 32),
+    ("cmd", 1),
+]
+
+spi_phy_data_layout = [
+    ("data", 32),
+]
--- a/litex/soc/cores/spi_xip/core.py
+++ b/litex/soc/cores/spi_xip/core.py
@@ -1,0 +1,77 @@
+from migen import *
+from migen.genlib.fsm import FSM, NextState
+
+from litex.soc.cores.spi_xip.common import *
+from litex.soc.interconnect import wishbone, stream
+from litex.gen.common import reverse_bytes
+
+
+class LiteSPICore(Module):
+    def __init__(self, endianness="big"):
+        self.source = source = stream.Endpoint(spi_phy_ctl_layout)
+        self.sink   = sink   = stream.Endpoint(spi_phy_data_layout)
+        self.bus    = bus    = wishbone.Interface()
+        self.cs_n   = cs_n   = Signal()
+        curr_addr   = Signal(32)
+        bus_read    = Signal()
+        cs_cnt      = Signal(16)
+        cs_val      = Signal(16)
+
+        self.submodules.fsm = fsm = FSM(reset_state="IDLE")
+
+        self.comb += [
+            bus_read.eq(bus.cyc & bus.stb & ~bus.we),
+            bus.dat_r.eq(sink.data if endianness == "big" else reverse_bytes(sink.data)),
+        ]
+
+
+        # TODO: make this configurable via CSR
+        self.comb += cs_val.eq(10000)
+
+        fsm.act("IDLE",
+            cs_n.eq(1),
+            If(bus_read,
+                NextState("CS_DELAY"),
+            )
+        )
+        fsm.act("CMD",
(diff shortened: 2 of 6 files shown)


piotr-binkowski:
@enjoy-digital @mithro please take a look

litex/soc/cores/spi_xip/core.py
```diff
@@ -0,0 +1,79 @@
...
+        bus_read = Signal()
+
+        self.comb += bus_read.eq(bus.cyc & bus.stb & ~bus.we)
+
+        if endianness == "big":
+            self.comb += bus.dat_r.eq(phy.source.data)
+        else:
+            for i in range(4):
+                self.comb += bus.dat_r[i*8:(i+1)*8].eq(phy.source.data[(3-i)*8:(4-i)*8])
+
```
enjoy-digital: There is a` reverse_bytes` function in `litex.gen.common` that you could probably use here:
````
self.comb += bus.dat_r.eq(phy.source.data if endianness == "big" else reverse_bytes(phy.source.data))
````

**Comment by enjoy-digital** [comment]
@piotr-binkowski: thanks the initial work. This is fine for me and using streams for the interfaces or in the core is a good idea since it will probably make it a lot more flexible/modular.

One of the issue with the current SPI Flash core is that it's doing too much thing in a very compact code which make it difficult to understand. It also lacks simulations, which make it difficult to adapt/change, so this new SPI XIP core should probably have unit-tests in `litex/test` to ease adding new features and simplify maintenance in the future.

File: litex/soc/cores/spi_xip/core.py (hunk 1 above)
Comment (piotr-binkowski) [reply litex/soc/cores/spi_xip/core.py]: Thanks, I was not aware of that function

**Comment by mithro** [comment]
@enjoy-digital - I'm wondering if it makes sense to create a new LiteSPI module? I have a feeling that while the module should remain small from a FPGA resource usage there will be a lot of code to support the various SPI modes and get the best possible performance?

File: litex/soc/cores/spi_xip/common.py [hunk 2]
```diff
@@ -0,0 +1,8 @@
+spi_phy_ctl_layout = [
```
mithro: Please add comments here.

mithro:
@piotr-binkowski It might be worth linking to my [Google Doc here](https://docs.google.com/document/d/1JZ7pU7roOaJwsv_LlXDKpWrLMSp254-tuof5KIFcC-I/edit).

enjoy-digital:
@mithro: yes this could be an external module, it would probably ease development and reuse, since from what i understand it's only for SPI Flash chips, what about LiteSPIFlash? Do you want me to create an initial skeleton? in enjoy-digital, litex-hub?

mithro:
@enjoy-digital - I think we are just concentrating on SPI Flash for now but will expand to other parts in the future?

enjoy-digital:
@mithro: ok then in this case LiteSPI makes sense yes.

mithro:
@enjoy-digital -- Do you want to create a skeleton for litespi under litex-hub?

enjoy-digital:
@mithro: yes i'll do that on next monday.

enjoy-digital:
@mithro, @piotr-binkowski: i just created an initial LiteSPI skeleton in https://github.com/litex-hub/litespi. I tried to have a structure close to what is used for LiteDRAM, but that's just to have a skeleton, feel free to change. We could probably move these PRs to it.

litex/soc/cores/spi_xip/common.py
piotr-binkowski: I've added some comments to the version in litex-hub/litespi#2

piotr-binkowski:
I think we can close this and continue in litex-hub/litespi#2

PR #441, opened 2020-03-27
SPI SDCard fixes and features

gsomlo:
- ensure parameters representing a host address are `unsigned long` (rather than `uint32_t`) to work on 64-bit CPUs
- factor out `busy_wait()` so we can use the same one throughout the bios source tree
- use `busy_wait()` in SPI SDCard initialization loop to avoid running down the retry counter before card is ready (on some boards, e.g. trellisboard)
- finally, add `spisdcardboot()` to the bios `boot_sequence()`

### DIFF
--- a/litex/soc/software/bios/main.c
+++ b/litex/soc/software/bios/main.c
@@ -593,6 +593,9 @@ static void boot_sequence(void)
 #ifdef ROM_BOOT_ADDRESS
 		romboot();
 #endif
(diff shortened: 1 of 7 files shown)


enjoy-digital:
Thanks! merged.

pdp7:
@gsomlo what board are you using?  I'd like to try it out, too.  thanks

PR #442, opened 2020-03-27
[WIP] Add generic PHY for the new SPI XIP core

piotr-binkowski:
This adds initial version of a generic PHY for the new SPI XIP core. It was tested on the NeTV2 board with single/dual/quad reads (address only on mosi for now).

Currently missing or untested features are:
- [ ] support for configuring core parameters based on selected chip
- [ ] support for sending address on 2/4 data pins (untested)
- [ ] support for octal mode chips
- [ ] 32-bit addressing (untested)
- [ ] DDR transfers
- [ ] option for software based access to the chip (configure registers, erase/program)

This depends on #440

### DIFF
--- a/litex/boards/platforms/netv2.py
+++ b/litex/boards/platforms/netv2.py
@@ -23,7 +23,7 @@ _io = [
         Subsignal("cs_n", Pins("T19")),
         Subsignal("mosi", Pins("P22")),
         Subsignal("miso", Pins("R22")),
-        Subsignal("vpp", Pins("P21")),
+        Subsignal("wp",   Pins("P21")),
         Subsignal("hold", Pins("R21")),
         IOStandard("LVCMOS33")
     ),
--- a/litex/soc/cores/spi_xip/phy/generic.py
+++ b/litex/soc/cores/spi_xip/phy/generic.py
@@ -1,0 +1,122 @@
+from migen import *
+from migen.genlib.fsm import FSM, NextState
+
+from litex.soc.cores.spi_xip.common import *
+from litex.soc.cores.spi_xip.clkgen import *
+
+from litex.soc.interconnect import stream
+
+cmd_oe_mask = 0b0001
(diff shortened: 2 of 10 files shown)


litex/boards/platforms/netv2.py
```diff
@@ -23,7 +23,7 @@
         Subsignal("cs_n", Pins("T19")),
         Subsignal("mosi", Pins("P22")),
         Subsignal("miso", Pins("R22")),
-        Subsignal("vpp", Pins("P21")),
+        Subsignal("wp",   Pins("P21")),
```
mithro: Can you submit this change to other locations that have netv2.py boards?

litex/soc/cores/spi_xip/phy/generic.py
```diff
@@ -0,0 +1,122 @@
...
+soft_oe_mask = 0b0001
+addr_oe_mask = {
+    1: 0b0001,
+    2: 0b0011,
+    4: 0b1111,
+}
+
+def GetConfig(flash=None):
+    if flash is None:
+        return (24, 8, 1, 1, 1, 0x0b)
```
mithro: Magic numbers - what do these mean? Should it be returning a named tuple or some other similar structure?

test/test_spi_xip.py
```diff
@@ -0,0 +1,68 @@
...
+            dut.addr_ok = 0
+            dut.cmd_ok = 0
+
+            yield dut.sink.valid.eq(0)
+            yield dut.source.ready.eq(1)
+
+            while (yield dut.source.valid) == 0:
+                yield
+
+            if (yield dut.source.addr) == (0xcafe<<2): # address cmd
```
mithro: Make address command a named value.

piotr-binkowski:
I've added some more comments/docstrings to the version in litex-hub/litespi#8 so I think that we can close this PR and continue work on this there

PR #446, opened 2020-03-30
Packet cleanup

jersey99:
Increasing the Ethernet data path width to 512 bit led me to this problem :)

The tx side is tested in both Simulation and Hardware. I haven't implemented the Rx side yet :(

But do have a look!


jersey99:
Beware of the pointers to other litex repos that I modified for testing. You will have to revert that commit if you decide to merge. I am looking for a clean way to set those pointers without making a commit.

enjoy-digital:
Thanks, curious to know how many Gbps you are currently testing to have a 512-bit data path?? :)
Have you been able to reproduce the issue with https://github.com/enjoy-digital/litex/blob/master/test/test_packet.py by adding a dw=512 test case? 

I'll have a closer look at the code. BTW, sorry for letting https://github.com/enjoy-digital/liteeth/pull/21 behind, it's a large PR which make it more difficult to merge, but i should have some help in the next weeks.

jersey99:
> Thanks, curious to know how many Gbps you are currently testing to have a 512-bit data path?? :)
This is for just the UDP/IP/MAC crossbar to generate packets for the Xilinx CMAC.

> Have you been able to reproduce the issue with https://github.com/enjoy-digital/litex/blob/master/test/test_packet.py by adding a dw=512 test case?

I will have a look, sorry didn't notice.

> I'll have a closer look at the code. BTW, sorry for letting [enjoy-digital/liteeth#21](https://github.com/enjoy-digital/liteeth/pull/21) behind, it's a large PR which make it more difficult to merge, but i should have some help in the next weeks.

Awesome!

enjoy-digital:
@jersey99: to help me understand the issue, can you explain what was the issue you observed?

jersey99:
Sorry, I didn't describe it earlier.

if the entire packet header was smaller than a single `dw` the state machine would fail. Previously, all the packet headers (specifically ethernet) that I had handled with this code were at least greater in size than the `dw` itself.

Hope that clarifies the issue a little better.

enjoy-digital:
@jersey99: ok i see now, thanks. I did a quick test with `test_packet`, previous tests are passing, but when adding 256-bit and 512-bit tests:
````python3
    def test_256bit_loopback(self):
        self.loopback_test(dw=256)

    def test_512bit_loopback(self):
        self.loopback_test(dw=512)
````
 The data seems shifted by one cycle, i'll have a closer look, but if you also want to have a look, i don't mind :)

**Comment by jersey99** [comment]
I added the test and watched it fail, didn't have time to look into it. I promise to look into it early this week, if that's ok?

**Comment by enjoy-digital** [comment]
Sure, no hurry on my side, thanks.

File: litex/soc/interconnect/packet.py [hunk 1]
```diff
@@ -337,13 +345,14 @@ def __init__(self, sink_description, source_description, header):
...
         )
         self.sync += If(sink.valid & sink.ready, sink_d.eq(sink))
+        print(header_words, header_leftover)
         fsm.act("UNALIGNED-DATA-COPY",
             source.valid.eq(sink.valid | sink_d.last),
             source.last.eq(sink.last | sink_d.last),
             sink.ready.eq(source.ready),
             source.data.eq(sink_d.data[header_leftover*8:]),
             source.data[min((bytes_per_clk-header_leftover)*8, data_width-1):].eq(sink.data),
-            If(fsm_from_idle,
```
jersey99, line 346: @enjoy-digital fsm_from_idle is a misleading name. As this is active here even when the FSM transitions from HEADER-RECEIVE.

jersey99:
@enjoy-digital I wonder if all branches are failing this travis check. Do you mind having a quick look? The test_packet does pass.

jersey99:
@enjoy-digital I merged upstream into the branch and looks like tests are passing. I was concerned that the tests we have were insufficient, I no longer think so. Here are my thoughts:

This PR tries to address the case where header.length is < dw.

1. The tests both unit and HW tests indicate that the case where header.length > dw isn't affected. No regression.
2. The unit tests we have seem to catch the case where header.length < dw (Note 512 bit dw loopback has 31 bytes/ 248 bits header.length).
3. And the Rx side is implemented.

Let me know what you think!

enjoy-digital:
We can probably close this, the changes required have been integrated in LiteEth.

PR #447, opened 2020-03-31
Add initial support for the new LiteSPI core

piotr-binkowski:
This PR adds support for LiteSPI to litex_sim and to NeTV2 target

### DIFF
--- a/litex/boards/targets/netv2.py
+++ b/litex/boards/targets/netv2.py
@@ -13,11 +13,15 @@ from litex.soc.cores.clock import *
 from litex.soc.integration.soc_core import *
 from litex.soc.integration.soc_sdram import *
 from litex.soc.integration.builder import *
+from litex.soc.integration.soc import *
 
 from litedram.modules import K4B2G1646F
 from litedram.phy import s7ddrphy
 
 from liteeth.phy.rmii import LiteEthPHYRMII
+
+from litespi import LiteSPI
+from litespi.phy.generic import LiteSPIPHY
 
 # CRG ----------------------------------------------------------------------------------------------
 
@@ -46,7 +50,7 @@ class _CRG(Module):
 # BaseSoC ------------------------------------------------------------------------------------------
 
 class BaseSoC(SoCCore):
-    def __init__(self, sys_clk_freq=int(100e6), with_ethernet=False, **kwargs):
+    def __init__(self, sys_clk_freq=int(100e6), with_ethernet=False, with_spi_xip=False, **kwargs):
         platform = netv2.Platform()
 
         # SoCCore ----------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks, is it working in simulation and on hardware?

piotr-binkowski:
Yes, it works both in simulation and on hardware but after litex-hub/litespi#9 gets merged the NeTV2 target will need an update to still work correctly.

litex/boards/targets/netv2.py
```diff
@@ -72,6 +79,17 @@ def __init__(self, sys_clk_freq=int(100e6), with_ethernet=False, **kwargs):
                 l2_cache_reverse        = True
             )
 
+        with_spi=True
```
enjoy-digital, line 82: Can you rename to `with_spi_xip` and add this to the BaseSoC parameters?

litex/boards/targets/netv2.py
```diff
@@ -72,6 +79,17 @@ def __init__(self, sys_clk_freq=int(100e6), with_ethernet=False, **kwargs):
...
+
+        if with_spi:
+            spi_size = 1024*1024*8
+            spipads = platform.request("spiflash4x")
+            #spipads = platform.request("flash")
+            self.submodules.spiphy = LiteSPIPHY(spipads)
+            self.submodules.spictl = LiteSPI(phy=self.spiphy, endianness=self.cpu.endianness)
+            self.add_memory_region("spixip", self.mem_map["spixip"], spi_size, type="io")
```
enjoy-digital, line 90: It's now possible to let the SoC automatically determines the mapping of the added periperal when added like this:
````python3 
spi_xip_region = SoCRegion(origin=self.mem_map.get("spixip",  None), size=spi_size, cached=False)
self.bus.add_slave(name="spixip", slave=self.spictl.bus, region=spi_xip_region)
````
Current code is not updated everywhere, but this is the recommended way for new peripherals. Can you change to that? Also, if you don't need static mapping (which should the case since the bios is using the csr.h and Zephyr/Linux are using Device Tree) you could remove the `spixip` definition in the `mem_map`.

File: litex/tools/litex_sim.py [hunk 3]
```diff
@@ -177,6 +182,14 @@ def __init__(self,
         # CRG --------------------------------------------------------------------------------------
         self.submodules.crg = CRG(platform.request("sys_clk"))
 
+        # SPI XIP ----------------------------------------------------------------------------------
+        if with_spi_xip:
+            xip_size = kwargs["spi_xip_size"]
+            self.submodules.spiphy = LiteSPIPHYModel(xip_size, init=kwargs["spi_flash_init"])
+            self.submodules.spi = LiteSPI(phy=self.spiphy, endianness=self.cpu.endianness)
+            self.add_memory_region("spixip", self.mem_map["spixip"], xip_size, type="io")
```
enjoy-digital, line 190: Same as previous comment in NeTV2 target.

litex/tools/litex_sim.py
```diff
@@ -279,6 +292,8 @@ def main():
     parser.add_argument("--threads",              default=1,               help="Set number of threads (default=1)")
     parser.add_argument("--rom-init",             default=None,            help="rom_init file")
     parser.add_argument("--ram-init",             default=None,            help="ram_init file")
+    parser.add_argument("--with-spi-xip",         action="store_true",     help="Enable SPI XIP support")
+    parser.add_argument("--spi-init",             default=None,            help="spi_init file")
```
enjoy-digital, line 296: We should avoid mixing `spi-xip`/`spi`, can you use `--spi-xip-init`?

litex/tools/litex_sim.py
```diff
@@ -311,6 +326,9 @@ def main():
     soc_kwargs["uart_name"] = "sim"
     if args.rom_init:
         soc_kwargs["integrated_rom_init"] = get_mem_data(args.rom_init, cpu_endianness)
+    if args.with_spi_xip:
+        soc_kwargs["spi_xip_size"] = 8*1024*1024
+        soc_kwargs["spi_flash_init"] = get_mem_data(args.spi_init, "big") if args.spi_init is not None else None
```
enjoy-digital, line 331: Same as previous comment, should we use "spi_xip_init" instead?

piotr-binkowski:
@enjoy-digital I've updated the code with changes you mentioned

enjoy-digital:
Thanks! That's merged.

PR #448, opened 2020-04-05
Fix timing constraints

kessam:
A proposed fix to issue https://github.com/enjoy-digital/litedram/issues/176


enjoy-digital:
I've been able to reproduce the issue by removing the `quiet` and test your fix. Thanks!

PR #450, opened 2020-04-06
litex_setup: Use subprocess so failures are noticed.

mithro:
os.system doesn't report if any of the commands fail. This means that if something goes wrong it happily reports success making it hard to debug issues.


enjoy-digital:
Thanks, merged.

PR #452, opened 2020-04-07
Add GCC downloading via litex_setup.py

### LINKED ISSUE #449: add note to quick start guide about ~/.local/bin path
The quick start guide suggests adding the `--user` switch to `litex_setup.py` which I think installs into `$HOME/.local`. It took me a while to figure out the `lxsim` executable ended up in `$HOME/.local/bin` which was not already part of my `$PATH`

mithro:
The downloading got a little more complicated then I would have liked.

It also tells user to update path at the end now (fixes #449).

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -85,6 +85,10 @@ def sifive_riscv_download():
             assert not os.path.exists(dst), dst
             os.symlink(src, dst)
 
+if os.environ.get('TRAVIS', '') == 'true':
+    # Ignore `ssl.SSLCertVerificationError` on CI.
+    import ssl
+    ssl._create_default_https_context = ssl._create_unverified_context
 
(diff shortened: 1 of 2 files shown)


litex_setup.py
```diff
@@ -29,18 +31,69 @@
...
+        url = base_url+fn
+        print("Downloading", url, "to", fn)
+        urllib.request.urlretrieve(url, fn)
+    else:
+        print("Using existing file", fn)
+
+    print("Extracting", fn)
+    shutil.unpack_archive(fn)
+
+    if "--user" in sys.argv[1:] and not is_windows:
```
mithro, line 65: This code symlinks the risc-v compiler into your `~/.local/bin` directory. Now that I have added this, I think it could be a mistake to do this and it does add a lot of extra code here.

PR #458, opened 2020-04-09
Add riscv64-none-elf triple

david-sawatzke:
It's (one) of the embedded riscv toolchains in nixos: https://github.com/NixOS/nixpkgs/pull/65410, there's no other already supported toolchain available (afaik). 

Inspired by #405.


david-sawatzke:
I've tested it with the "vexriscv" cpu and the bios works, so I guess it does

mithro:
@david-sawatzke - IIRC Generally it doesn't fail until you try to print a floating point number (as that causes gcc to pull in the floating point emulation stuff). Fact that the BIOS works seems good enough test for now.

PR #459, opened 2020-04-09
Two small Travis-CI related patches

mithro:
* Use the `litex_setup.py` from the checked out code (rather than always downloading the head version).
 * Run the Windows build but allow it to fail (it won't cause Travis to report failure if it fails).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,16 @@ jobs:
     - os: osx
       osx_image: xcode10.2 # Includes Python 3.7
       language: shell
-#    - os: windows
-#      language: shell
-#      before_install:
-#        - choco info python3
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks. The changes to download the head version was intended to have almost similar .travis.yml script between LiteX and the cores which simplify maintenance. (litex_setup.py is not available in the cores).

enjoy-digital:
In fact using the head version will not work when changes are made to litex_setup.py and Travis-CI is running on a branch or PR, so let's keep your solution.

mithro:
@enjoy-digital Modifying `litex_setup.py` is exactly how I discovered the problem :-)

PR #462, opened 2020-04-12
Add support for ecp5 12k device in trellis.py 

ironsteel:
It's a minor change specific to trellis, so we can skip it if we decide that it's not that important. 

Update the help description for --device option for ulx3s

I've tested the changes on my ulx3s-12k board and it works.


enjoy-digital:
Thanks, that's merged.

PR #464, opened 2020-04-12
Improve the litex_sim Makefiles

mithro:
This pull request fixes the way litex_sim Makefiles work (and was discovered as part of #399).

The fixes are;
 - [x] All compiled object files are in the `soc_simsoc_litex_sim/gateware` directory. Previously a bunch of `.o` and `.so` files ended up in the source directory.
 - [x] The correct `variables.mak` file is used. Previously there was a file committed to the source tree that was used.
 - [x] Reduces complexity of the sim modules Makefiles.


mithro:
I would like to litex_sim to be tested on Travis, at the moment I'll probably only do compiling the simulation as I can't see an easy way to test the actual simulation side of things.

enjoy-digital:
Thanks this seems fine for me. We are in fact using running the simulation with Travis-CI on LiteDRAM to test the DRAM controller with the DFI model and the different SDRAM Modules. We could do something similar in LiteX with a simulation of a simple SoC.

mithro:
@enjoy-digital -- If this works for you, then we can merge it and work on further improvements later.

enjoy-digital:
Thanks, merged.

PR #474, opened 2020-04-26
Terminal: add history and auto completion

fjullien:
This serie adds history and auto-completion to the BIOS terminal.
The BIOS size increase is very small (about 3K of ROM and 1K of RAM).
This serie needs some testing as I wasn't able to test all available commands.


mithro:
@fjullien - This seems like a cool feature! However, as the BIOS is frequently embedded in blockram in the gateware an increase of 3K ROM and 1K RAM will be too big for many targets (specially for targets like the iCE40 which only have tiny amounts of space). If this feature can be optionally enabled then we should most certainly merge it.

You might be able to reduce the size quite a bit by using link time optimization (LTO). @mateusz-holenko was working on getting that re-enabled in LiteX. I'm sure he would love help working on that.

litex/soc/software/bios/Makefile
```diff
@@ -10,7 +10,23 @@ ifdef TFTP_SERVER_PORT
 CFLAGS += -DTFTP_SERVER_PORT=$(TFTP_SERVER_PORT)
 endif
 
-OBJECTS=isr.o sdram.o sdcard.o main.o boot-helper-$(CPU).o boot.o
+OBJECTS= \
```
mithro, line 13: Can you indent this list?

litex/soc/software/bios/commands/cmd_bios.c
```diff
@@ -0,0 +1,148 @@
...
+struct command_struct cmd_crc =
+{
+	.func = crc,
+	.name = "crc",
+	.help = "Compute CRC32 of a part of the address space",
+};
+
+define_command(cmd_crc);
+
+/**************************************
```
mithro, line 116: The formatting of this comment header kind of looks weird.

It would be nice if it was a doxygen style comment. (Same for other comments of this form.)

litex/soc/software/bios/commands/cmd_bios.c
```diff
@@ -0,0 +1,148 @@
...
+ *            Flush cache             *
+ **************************************
+ */
+static void flush_cpu_dcache_handler(int nb_params, char **params)
+{
+	flush_cpu_dcache();
+}
+
+struct command_struct cmd_flush_dcache =
```
mithro, line 125: Can this be a const structure (same of other similar structures).

litex/soc/software/bios/commands/cmd_boot.c
```diff
@@ -0,0 +1,107 @@
...
+
+struct command_struct cmd_serialboot =
+{
+	.func = serialboot_handler,
+	.name = "serialboot",
+	.help = "Boot via SFL",
+};
+
+define_command(cmd_serialboot);
+
```
mithro, line 67: I was wondering if the handler / struct / define_command call should all be one big macro (or maybe even autogenerated using a Python script)? Don't know if that is a good idea or not?

litex/soc/software/bios/commands/cmd_boot.c
```diff
@@ -0,0 +1,107 @@
...
+/**************************************
+ *          Ethernet boot             *
+ **************************************
```
mithro, line 70: I do like that this kind of structure also leads the way towards `liteeth` and `litedram` providing the BIOS features as compile time plugins rather than it all being in stuck in the litex repo.

litex/soc/software/bios/complete.c
```diff
@@ -0,0 +1,174 @@
+// This file is Copyright (c) 2020 Franck Jullien <franck.jullien@gmail.com>
+//
+//     Largely inspired/copied from U-boot and Barebox projects wich are:
+//         Sascha Hauer, Pengutronix, <s.hauer@pengutronix.de>
+
+// License: BSD
```
mithro, line 6: Since this is mostly there, could you use an SPDX identifier tag here?

litex/soc/software/bios/commands/cmd_dram.c
```diff
@@ -0,0 +1,309 @@
...
+}
+
+struct command_struct cmd_sdrhw =
+{
+	.func = sdrhw_handler,
+	.name = "sdrhw",
+	.help = "Gives SDRAM control to HW",
+};
+
+define_command(cmd_sdrhw);
```
fjullien, line 77: Reading this now makes me think I could save space if .func would point directly to sdrhw()

litex/soc/software/bios/Makefile
fjullien: ok

litex/soc/software/bios/commands/cmd_bios.c
fjullien: I'll try to find something. Or, you could give me an example.

litex/soc/software/bios/commands/cmd_bios.c
fjullien: Yes

litex/soc/software/bios/commands/cmd_boot.c
fjullien: A big macro could save us a lot of line code. Especially if make .func point directly to functions that don't require parameters.

litex/soc/software/bios/complete.c
fjullien: ok

litex/soc/software/bios/commands/cmd_boot.c
mithro: I'm not a *huge* fan of macro abuse, but it does seem like a lot of code could be saved.

The fact the bios is already part of a Python based build system might mean it is better to use a Python script to generate code rather than making the C preprocessor do the work?

fjullien: > I'm not a huge fan of macro abuse
Neither am I.

> a Python script to generate code

From my point of view, doing some auto-generated C code is like doing a big macro.
So, I would vote for a C macro here.

fjullien:
> If this feature can be optionally enabled then we should most certainly merge it.

I'll work on this.

fjullien:
> @fjullien - This seems like a cool feature! However, as the BIOS is frequently embedded in blockram in the gateware an increase of 3K ROM and 1K RAM will be too big for many targets (specially for targets like the iCE40 which only have tiny amounts of space). If this feature can be optionally enabled then we should most certainly merge it.
> 
> You might be able to reduce the size quite a bit by using link time optimization (LTO). @mateusz-holenko was working on getting that re-enabled in LiteX. I'm sure he would love help working on that.

fjullien:
>  If this feature can be optionally enabled then we should most certainly merge it.

 I'll work on this.

Here is numbers:
```
Original:

ROM usage: 18.07KiB     (56.47%)
RAM usage: 0.30KiB      (7.42%)

Complete + hist:

ROM usage: 21.73KiB     (67.90%)
RAM usage: 2.23KiB      (55.66%)

----- Every new case feature is additive -----

Without complete:

ROM usage: 20.58KiB     (64.31%)
RAM usage: 0.94KiB      (23.44%)

Without history:

ROM usage: 20.08KiB     (62.74%)
RAM usage: 0.30KiB      (7.42%)

struct member .func point to function directly:

ROM usage: 19.77KiB     (61.79%)
RAM usage: 0.30KiB      (7.42%)

Terminal does no handle special keys:

ROM usage: 19.00KiB     (59.38%)
RAM usage: 0.30KiB      (7.42%)

Use old readline (readstr) conditionally but still use new parser and cmd struct:

ROM usage: 18.64KiB     (58.25%)
RAM usage: 0.30KiB      (7.42%)
```

What degree of granularity do we want in the configuration ?
We can easily have all.

I need to do some testing before I submit a v2 of this PR.

mithro:
@xobs and @mateusz-holenko probably have a better understanding of what sizes we should be targetting.

enjoy-digital:
Thanks @fjullien for this great contribution!  Having auto-complete and history will be very useful. We should indeed make this configurable. I'm not sure we need too much granularity, maybe just being able enable auto-complete and history separately.

@mithro, @mateusz-holenko, @xobs: that would indeed be useful to define the constraints on the different targets for the BIOS ROM/RAM usage to avoid being too restrictive on the additional features. Since the BIOS can be used on iCE40 to Ultrascale+ FPGAs, even if we try to be efficient on all targets, "big" and "small" are still a bit relative to the FPGA.

I'm not sure for example to be well aware of the constraints on very small FPGAs, since from my understanding, on iCE40 the available ROM memory is already too small to fit the current BIOS and when the SoC is using the LiteX BIOS it is always stored in SPI Flash and executed from there. So in this case the constraint seems more related to the RAM and the ROM size is not really an issue? (it could be if we wanted to maximize speed in the BIOS, but i don't think it's the case).

The  BIOS can also be used for different use cases: as the main bootloader, as a first stage bootloader before a proper bootloader (like Barebox for example), just to initialize some peripherals with no user interaction (LiteDRAM generated in standalone for example), etc...  so a next step would maybe to have flags to enable/disable features more easily.

litex/soc/software/bios/commands/cmd_boot.c
enjoy-digital: @mithro: i'm not sure we should abuse too much of Python/C code mixing. It's probably fine to have auto-generated access functions or headers from Python, but not sure this is good idea for this part of the code, so i would also vote for C macros for this part (at least for now, we could change it in the future if we see if would really simplify things).

xobs:
The BIOS seems to be useful for the same reasons that a PC BIOS is useful: Initializing RAM and loading the main program.  However, even PC BIOSes nowadays have full GUI environments.

I believe we talked about this in an earlier patch, but U-Boot has a concept of the `SPL` whose job it is to initialize memory and then load the rest of U-Boot.  It's designed to be relatively small, whereas U-Boot itself is relatively big.

Overall I find the BIOS has four uses:

1. Initialize DDR
2. Act as a rudimentary development environment (with peek and poke)
3. Load programs, either off SPI or over serial
4. Provide that "A-ha!" moment when people first port litex to their platform and see the banner

Once you have a payload such as linux or zephyr then items (2)-(4) become less useful.  This patch helps with item (2).

On ice40 we skip (1) because RAM doesn't need to be initialized, and (2)-(4) are not as useful on a shipping product, which is why I very much appreciate being able to remove the BIOS.

So ultimately I think this is a good thing, because it definitely helps with (2) and to a lesser extent (4).

mateusz-holenko:
@fjullien Size reduction we where able to obtain aftert enabling LTO is presented in https://github.com/enjoy-digital/litex/pull/253. We had problems with getting rid of `-lgcc` flag and got it working at some point with https://github.com/enjoy-digital/litex/pull/401.

Unfortunatelly it turned out that the solution was not versatile enough and didn't work in all cases, so it was reverted.

litex/soc/software/bios/Makefile
fjullien: done

litex/soc/software/bios/commands/cmd_bios.c
fjullien: done

litex/soc/software/bios/commands/cmd_bios.c
fjullien: done

litex/soc/software/bios/commands/cmd_boot.c
fjullien: I created a macro doing this:
https://github.com/fjullien/litex/blob/term_hist_auto_compl/litex/soc/software/bios/command.h#L36

fjullien:
I added an optional parameter to Builder in order to pass options to the BIOS.
This way, we can configure what termial feature we want. For example:
`builder_kwargs["bios_options"] = ['TERM_MINI']`
So we have:

1.   `TERM_MINI` -> the smallest one
2.   `TERM_NO_COMPLETE` standard terminal with no command completion
2.   `TERM_NO_HIST` standard terminal with no history

enjoy-digital:
Thanks for the update @fjullien, we'll review it soon.

enjoy-digital:
I think all @mithro's comments have been addressed and i've been able to test it successfully in simulation and on hardware, so think we can merge now. We'll maybe have to discuss the default settings depending the use cases/targets, but otherwise this is really cleaning up  and improving the BIOS, thanks for this work @fjullien!

Issue #471, opened 2020-04-23
Add SERV support


enjoy-digital:
SERV is a bit-serial  RISC-V core that could be useful in some designs to replace a complex FSM by a CPU while still minimizing resource usage. This would be a perfect fit for LiteDRAM standalone core where the CPU is only doing the DRAM initialization.

Initial SERV integration has been pushed to https://github.com/enjoy-digital/litex/tree/serv but needs to be finished.

The LiteX SoC with  SERV CPU can be simulated with: `litex_sim --cpu-type=serv` and waveform can be generated by adding `--trace` to the command. The waveform will then be located in `soc_simsoc_litex_sim/gateware/dut.vcd`.

For bring up, the UART is configured in polling mode with `self.add_constant("UART_POLLING")`. This allows doing the initial tests without interrupt support.

gregdavill:
So I just setup a project here using SERV. So I wish I'd seen this a few hours ago.

One thing I hit when bringing up my project was the wishbone address bus. LiteX uses a word addressed bus, Serv a byte addressed one, which provides the lowest 2 bits. So I had to drop these on both instruction and data buses.
```python
i_addr = Signal(32)
d_addr = Signal(32)

self.comb += [
    ibus.sel.eq(0xF),
    ibus.stb.eq(ibus.cyc),
    dbus.stb.eq(dbus.cyc),
    ibus.adr.eq(i_addr[2:]),
    dbus.adr.eq(d_addr[2:]),
]

self.cpu_params = dict(
...
        o_o_ibus_adr      = i_addr,
...
        o_o_dbus_adr      = d_addr,
    )
```

When I get a chance I'll take a look at your initial integration.

enjoy-digital:
@gregdavill: thanks for the feedback. I just fixed these points in the initial integration https://github.com/enjoy-digital/litex/commit/2efd939d06ba7903892fc5a742838b49ab1728fb.
There is still probably something wrong somewhere since the simulation is still not booting, i will have a look in the next days. Is your project available somewhere to have a look?

gregdavill:
I think we need to use `serv_rf_top` here:
```diff
- self.specials += Instance("serv_top", **self.cpu_params)
+ self.specials += Instance("serv_rf_top", **self.cpu_params)
```
https://github.com/enjoy-digital/litex/compare/serv#diff-9f586f8936c1ea0b2741be8735346161R91

[...]

PR #475, opened 2020-04-27
build/trellis: add verilog_read -defer option to yosys script

gregdavill:
This enables Yosys to parse verilog files which contain potential invalid default parameters. 

As a verilog coding style, parameters are left with a value of 0 to force/ensure all instantiated modules fill these parameters with correct values.

This fixes an issue encountered when integrating the serv CPU. https://github.com/enjoy-digital/litex/issues/471#issuecomment-619569830

```
 -defer
        only read the abstract syntax tree and defer actual compilation
        to a later 'hierarchy' command. Useful in cases where the default
        parameters of modules yield invalid or not synthesizable code.
```


enjoy-digital:
Thanks for investigating on this.

PR #477, opened 2020-04-27
diamond: fix include paths

shuffle2:
include paths given via tcl script need semicolon separators and forward slash as directory separator (even on windows)


enjoy-digital:
Thanks.

PR #478, opened 2020-04-28
Extended SPI flash support

mateusz-holenko:
This adds support for memory mapped writes
in the same configuration as memory mapped reads
are currently supported.

It also adds support for accessing registers
and erasing sectors in non-bitbanged single SPI
mode.

Finally it fixes building spi_flash module in non-bitbanged mode.

### DIFF
--- a/litex/soc/cores/spi_flash.py
+++ b/litex/soc/cores/spi_flash.py
@@ -2,6 +2,7 @@
 # This file is Copyright (c) 2014-2018 Florent Kermarrec <florent@enjoy-digital.fr>
 # This file is Copyright (c) 2013-2014 Robert Jordens <jordens@gmail.com>
 # This file is Copyright (c) 2015-2014 Sebastien Bourdeauducq <sb@m-labs.hk>
+# This file is Copyright (c) 2020 Antmicro <www.antmicro.com>
 
 # License: BSD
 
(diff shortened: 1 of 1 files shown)


mateusz-holenko:
@enjoy-digital What do you think about this PR?

enjoy-digital:
@mateusz-holenko: thanks for the PR. Even if it create some duplication, it would probably be easier to merge if this was integrated as a separate Module keeping the old one un-modified (the Module would just then a different name). We could try to remove duplication later (or never if LiteSPI replace all the usecases advantageously). Also can you indicate how to test it and/or provide a minimal software for it?

DurandA:
I am sorry to hijack the conversation, but I am a bit confused about the status of flash support in LiteX. Should we use `litex.soc.cores.spi_flash` until [LiteSPI](https://github.com/litex-hub/litespi) is ready?

Also, is it advisable to add SPI flash support to existing targets? There aren't so many targets that use `SpiFlash` in [litex-boards](https://github.com/litex-hub/litex-boards) whilst most of them (if not all?) use some kind of SPI flash memory.

enjoy-digital:
@DurandA: yes you can use the `spi_flash` core while `LiteSPI` is still being developed. We could indeed integrate the SPIFlash in `litex-boards`, i'll look at that in the next days.

enjoy-digital:
@mateusz-holenko: i've merged the PR and added back the old version and renamed the new one to `SpiFlashQuadReadWrite`. I'd like to do some tests with it, can you indicate how to use it? Also, do you still need the bitbang interface with in this version?

mateusz-holenko:
@enjoy-digital Sorry for not responding, I am involved in many issues lately.

Thanks for merging our version of the module. I guess renaming it and keeping the old one is better from the backward compatibility point of view.

As for the software I know for sure we were testing it on our WIP versions of Linux and Zephyr drivers, but I'll ask for details and get back to you. 

I don't think we need the bitbanged interface anymore (but I'll ask to make sure), but we probably kept it in order to support the previous use cases.

enjoy-digital:
> @enjoy-digital Sorry for not responding, I am involved in many issues lately.

Sure no problem, that's what i was guessing. Having the informations i ask will allow me to have a closer look and do eventual simplifications, but no hurry.

mateusz-holenko:
@enjoy-digital I asked and we were testing it on Zephyr/Linux drivers only. They are currently still in progress, but we hope to publish them soon.

As for the bitbang interface - isn't it used by LiteX BIOS to `flashboot`? We kept the interface compatible so that this feature works with the new version of SPI Flash core.

Issue #479, opened 2020-04-28
litex broken under python3.5


bunnie:
The use of f-string formatting in on this line breaks litex under python 3.5.2:

https://github.com/enjoy-digital/litex/blob/56aa7897df99d7ad68ea537ab096c3abdc683666/litex/build/lattice/programmer.py#L35

This is from commit 351551a041f737b7b0ac08132a5b356fa8be468b 

It looks like f-strings were introduced in python 3.6+

Was it the intent to break environments running with python 3.5? That is, what is official minimum recommended level for python for litex to work. 

If the officially supported python is a level higher than python3.5 I'll upgrade my system and re-do the build scripts to check for this compatibility issue. However, I'd rather not be constantly chasing the upgrade curve on python 3 to incorporate the latest features for no reason. Especially considering I don't even use this feature, but it's a build-breaker because of the syntax error it causes.

mithro:
Python 3.5 was released in 2015-09-13 and the current Python version is 3.8 -- If you are upgrading, jump straight to 3.8

Currently tested versions are 3.6, 3.7 and 3.8 -> https://travis-ci.com/github/enjoy-digital/litex

bunnie:
So is it official policy that python 3.6>= is required for Litex?

I don't think "old" is a reason on its own to deprecate support. Ideally it would be a reason like, "this feature is crucial to rock solid code" or "there was a security problem and the upstream maintainers aren't fixing it so we've deprecated this version (e.g. like Python2)".

Example: NeTV2 was started in 2016 and is now in maintenance mode. At that time, Python 3.5 was only a year old. I still have to maintain and build the NeTV2 base, and I'd like a reproducible build environment. 

Upgrading Python on my development instance may also introduce bugs into NeTV2 and should strictly require a full regression. This is especially true because we had problems with a sensitivity to the ordering of dictionary keys back then, and those orderings change with the Python version. 

So while it's an easy matter to upgrade Python for a project in active development, it's a big problem for projects in maintenance mode because of the fragility of Python code. Thus I would prefer we deprecate a Python version only if there's a really good reason to do so...

mithro:
@bunnie I'm not sure I understand? For projects not in active development why would LiteX be changing?

FYI - It is always recommended you use isolated environments + installs (not based on your system) for each project you have -- this is specially for maintenance mode projects.

bunnie:
Litex isn't changing. The version of Python3 on the system is. 

[...]

PR #481, opened 2020-04-28
propose patch to not break litex for python 3.5

bunnie:
fstrings in this module breaks Litex for python3.5

If you could please take this patch it would save me a lot of headache updating python on my system and potentially upsetting other projects in maintenance....thank you!


enjoy-digital:
Thanks, we can apply it, as discussed in https://github.com/enjoy-digital/litex/issues/479, LiteX will need to live and evolve and at some point we'll probably want to use features introduced after Python3.5. So we even if with this patch LiteX still support Python3.5, i think it would be good to also prepare things on your side and eventually use VM or isolated environment to help maintenance if the it's too complicated/risky to use upstream LiteX. We'll also try to do regular release of the project (every 3 month probably) and could make sure Python requirements do not changes between release and that things are better planned.

PR #482, opened 2020-04-28
SPIMaster: Added support for different CPOL/CPHA

ilya-epifanov:


enjoy-digital:
Thanks, it seems fine for CPOL, but i don't think it's correct for CPHA, because in the case of CPOL=1, CPHA=1, CPOL ^ CPHA = 0 so the behavior will be similar to CPOL=0, CPHA=0.

ilya-epifanov:
@enjoy-digital yep, it's WIP. should be easy to code, but I have to verify it with an LA.

sergachev:
Note: there is an implementation of SPI Master in MiSoC which supports this and is naturally easy to use with LiteX: https://github.com/m-labs/misoc/blob/master/misoc/cores/spi.py

enjoy-digital:
@ilya-epifanov: i'm closing since some work would still need to be done to get it working and as @sergachev others alternative core could be used, but please create another PR if you have a working implementation.

PR #483, opened 2020-04-28
Lattice OpenOCD JTAG programmer: removed erase flag and made progress output less noisy

ilya-epifanov:
I don't know what I was thinking when I exposed the `erase` flag. There's no reason to set it to `False`


enjoy-digital:
Thanks, merged.

PR #487, opened 2020-04-29
litex/soc/integration: Add SDK output directory

John-K:
This PR adds an SDK output directory and moves includes and binary artifacts into it.

In discussion with @mubes on 1BitSquared discord, we realized that in order for software developers to build programs for our SoCs they would have to have litex setup and at least build the design with `--no-compile-gateware`. It would be much more user friendly if we could ship an SDK along with the bitstream to facilitate firmware development.

To this end, this PR adds a `--sdk-dir` argument which defaults to `soc_<soctype>_<board>/sdk`, adds `install` targets to the litex library makefiles, and modifies `soc/integration/builder.py` to copy headers and binary artifacts to the SDK location as well as adding that directory to `variables.mak`

I'd welcome feedback on this PR, it's a quick (and a bit dirty) implementation of the idea and would love some help to refine it.

Example end-user Makefile:
```
SDK_DIR=../sdk

include $(SDK_DIR)/include/generated/variables.mak
include $(SDK_DIR)/common.mak

OBJECTS=main.o

all: firmware.bin

# pull in dependency info for *existing* .o files
-include $(OBJECTS:.o=.d)

%.bin: %.elf
        $(OBJCOPY) -O binary $< $@
        chmod -x $@

firmware.elf: $(OBJECTS)
        $(LD) $(LDFLAGS) \
                -T linker.ld \
                -N -o $@ \
                $(SDK_DIR)/lib/crt0-$(CPU)-ctr.o \
                $(OBJECTS) \
                -lbase-nofloat -lcompiler_rt
        chmod -x $@

main.o: main.c
        $(compile)

%.o: %.c
        $(compile)

%.o: %.S
        $(assemble)

clean:
        $(RM) $(OBJECTS) $(OBJECTS:.o=.d) firmware.elf firmware.bin .*~ *~

.PHONY: all main.o clean load
```
Example end-user linker script:
```
INCLUDE generated/output_format.ld
ENTRY(_start)

__DYNAMIC = 0;

INCLUDE generated/regions.ld

SECTIONS
{
        .text :
        {
                _ftext = .;
                *(.text .stub .text.* .gnu.linkonce.t.*)
                _etext = .;
        } > main_ram

        .rodata :
        {
                . = ALIGN(4);
                _frodata = .;
                *(.rodata .rodata.* .gnu.linkonce.r.*)
                *(.rodata1)
                _erodata = .;
        } > main_ram

        .data :
        {
                . = ALIGN(4);
                _fdata = .;
                *(.data .data.* .gnu.linkonce.d.*)
                *(.data1)
                _gp = ALIGN(16);
                *(.sdata .sdata.* .gnu.linkonce.s.*)
                _edata = .;
        } > main_ram

        .bss :
        {
                . = ALIGN(4);
                _fbss = .;
                *(.dynsbss)
                *(.sbss .sbss.* .gnu.linkonce.sb.*)
                *(.scommon)
                *(.dynbss)
                *(.bss .bss.* .gnu.linkonce.b.*)
                *(COMMON)
                . = ALIGN(4);
                _ebss = .;
                _end = .;
        } > sram
}

PROVIDE(_fstack = ORIGIN(sram) + LENGTH(sram) - 4);
```

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -6,10 +6,12 @@
 # This file is Copyright (c) 2018 William D. Jones <thor0505@comcast.net>
 # This file is Copyright (c) 2020 Xiretza <xiretza@xiretza.xyz>
 # This file is Copyright (c) 2020 Piotr Esden-Tempski <piotr@esden.net>
+# This file is Copyright (c) 2020 John Kelley <john@kelley.ca>
 # License: BSD
 
 
 import os
+import re
 import subprocess
 import struct
 import shutil
@@ -41,6 +43,7 @@ class Builder:
         output_dir       = None,
         gateware_dir     = None,
         software_dir     = None,
+        sdk_dir          = None,
         include_dir      = None,
         generated_dir    = None,
         compile_software = True,
@@ -56,7 +59,8 @@ class Builder:
         self.output_dir    = os.path.abspath(output_dir    or "soc_{}_{}".format(soc.__class__.__name__.lower(), soc.platform.name))
         self.gateware_dir  = os.path.abspath(gateware_dir  or os.path.join(self.output_dir,   "gateware"))
         self.software_dir  = os.path.abspath(software_dir  or os.path.join(self.output_dir,   "software"))
-        self.include_dir   = os.path.abspath(include_dir   or os.path.join(self.software_dir, "include"))
+        self.sdk_dir       = os.path.abspath(sdk_dir       or os.path.join(self.software_dir, "sdk"))
+        self.include_dir   = os.path.abspath(include_dir   or os.path.join(self.sdk_dir, "include"))
         self.generated_dir = os.path.abspath(generated_dir or os.path.join(self.include_dir,  "generated"))
 
         self.compile_software = compile_software
@@ -103,6 +107,8 @@ class Builder:
             define("SOC_DIRECTORY", soc_directory)
             variables_contents.append("export BUILDINC_DIRECTORY\n")
             define("BUILDINC_DIRECTORY", self.include_dir)
+            variables_contents.append("export SDK_DIRECTORY\n")
+            define("SDK_DIRECTORY", self.sdk_dir)
             for name, src_dir in self.software_packages:
                 define(name.upper() + "_DIRECTORY", src_dir)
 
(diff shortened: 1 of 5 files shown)


John-K:
It would appear that this only works correctly with Python 3.8 at the moment due to the `dirs_exist_ok` argument to `shutil.copytree`

mithro:
@John-K - The approach we have been using for litex-buildenv is to just ship the whole build output directory. See https://github.com/timvideos/HDMI2USB-firmware-prebuilt/tree/master/archive/master/v0.0.4-584-g2011661 for example.

John-K:
@mithro that doesn’t seem to work in my case - firmware source code still needs headers from within Litex source tree to compile

mithro:
@John-K - It might be worth looking at how FuPy does it. @ewenmcneill has a bunch of experience here.

If a person is developing their own firmware, they are probably better of using a *real* C standard library rather than the hacked together one that the BIOS currently uses. (In fact we are slowly trying to kill that library and move the BIOS to a more normal C standard library too.)

ewenmcneill:
> The goal here is to produce a release artifact that doesn’t depend on Litex or any of the tooling and enables end user to develop their own firmware for the SoC

I'd also be in favour of having a way to do this: ie have a way of releasing something relatively small as a companion to a pre-built gateware bitstream, that would allow "software tools only" development.

Currently I suspect FuPy (FPGA MicroPython)  probably only builds in the context of litex-buildenv, and then only after building the gateware at the same time.   I'd *wanted* a way that there could be a "user mode" for just building MicroPython alone (see, eg, https://github.com/fupy/micropython/issues/52, which I closed without anything being done after I learned enough about litex to realise just now non-trivial it was).  But the last hope of that disappeared when the CSR interface of litex changed a few months back in non-backwards compatible ways (so we [removed the cached copies](https://github.com/fupy/micropython/commit/e23216bf29912cfe6dad22c9a5f42451625234e7)).

The [existing way that FuPy builds from *within* litex-buildenv](https://github.com/timvideos/litex-buildenv/blob/master/scripts/build-micropython.sh) is an ugly kludge of symlinks and compile flags to expose the "just built minutes ago" litex artefacts to the MicroPython build environment, via  build directory.  Which I suspect isn't usable outside litex-buildenv, even if on were to, eg, throw it into a tarball.  

It'd be a *lot* easier if there was a clear equivalent of, eg, "linux-kernel-headers" packaging in most distros, that given a "binary kernel" (ie, binary bitstream) and that header files, one could build related things.  (It'd also be easier if, eg, the memory location/CSR locations didn't shift around from build to build as much; IIRC some of the Fomu things end up pinning locations of everything to be able to, eg provide end user documentation referencing specific bus locations for reset, metadata, etc.)

Ewen

PS: I'm currently a day or two behind on reading Discord, so I've not seen the referenced discussion there (but have seen most of the earlier history that I assume is behind that); I'm unlikely to have time to catch up reading there until at least tonight.

mubes:
I didn't manage to get to this today but will look at it tomorrow with any luck. I think everyone is in agreement that something should be done. The exact detail of what form it has and how it's packaged can be discussed once there's something underway. My personal preference is for one version-able and archive-able file the represents an internally consistent state (documentation, libraries, bitfile, headers and information about how it was built) because those could be reasonably safely referenced from other projects, but that detail is for later once there's an exemplar.

enjoy-digital:
Thanks @John-K, i agree we should improve things. I'm currently currently doing some cleanup/refactoring that should simplify things and i just like to finish this before being able to generate a SDK. I'll think about this while doing the changes and we could discuss more.

enjoy-digital:
We can probably close this, the idea are interesting but I'd like to avoid duplicating things for now and would just recommend using the generated files as SDK. Things will probably evolve but we haven't been able to focus on this for now.

PR #491, opened 2020-05-01
software: spisdcard: cosmetic: avoid filling screen with cluster numbers

gsomlo:
When boot.bin is large, we fill several screens with cluster numbers. Instead, let's use `\r` and keep overwriting the same line until we're done loading clusters.


enjoy-digital:
Thanks, merged.

PR #494, opened 2020-05-04
diamond: quiet warning about missing clkin freq for EHXPLLL

shuffle2:
FREQUENCY_PIN_CLKI should be given in mhz


enjoy-digital:
Thanks!

PR #500, opened 2020-05-07
Fixups

mubes:
A number of small changes to fix compiler warnings etc with gcc 9.2.


litex/soc/interconnect/wishbonebridge.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, phy, clk_freq):
         ]
 
         fsm = ResetInserter()(FSM(reset_state="IDLE"))
-        timer = WaitTimer(clk_freq//10)
+        timer = WaitTimer(int(clk_freq//10))
```
enjoy-digital, line 70: `clk_freq` is supposed to already be an` int` in the SoC.

litex/soc/software/libbase/strtod.c
```diff
@@ -107,7 +107,7 @@ double strtod(const char *str, char **endptr)
   switch (*p)
     {
     case '-':
-      negative = 1; /* Fall through to increment position */
+      negative = 1; /* FALLTHRU */ /* to increment position */
```
enjoy-digital, line 110: Are you sure you wanted this to be part of the PR?

litex/soc/software/libbase/strtod.c
```diff
@@ -166,7 +166,7 @@ double strtod(const char *str, char **endptr)
       switch(*++p)
         {
         case '-':
-          negative = 1;   /* Fall through to increment pos */
+          negative = 1;   /* FALLTHRU */ /* to increment pos */
```
enjoy-digital, line 169: Are you sure you wanted this to be part of the PR?

litex/soc/software/libbase/vsnprintf.c
```diff
@@ -262,6 +262,7 @@ int vsnprintf(char *buf, size_t size, const char *fmt, va_list args)
 
 			case 'X':
 				flags |= PRINTF_LARGE;
+                                /* FALLTHRU */
```
enjoy-digital, line 265: Are you sure you wanted this to be part of the PR?

litex/soc/interconnect/wishbonebridge.py
mubes: Probably so, but for the case that someone accidentally passes a float in (that would've been me), this avoids a hiccup and it has no downside as far as I can see.

litex/soc/software/libbase/strtod.c
mubes: Yes, the text /* FALLTHRU */ is compiler magic to avoid a fall through code warning.

litex/soc/software/libbase/strtod.c
mubes: See previous response.

litex/soc/software/libbase/vsnprintf.c
mubes: See previous response...avoids the alternative of someone switching off the warning from the command line, which would defeat its purpose.

mubes: I don't like these decorators, they look untidy in the code, but the warnings are worse.

litex/soc/interconnect/wishbonebridge.py
enjoy-digital: OK, we can add it yes, but using a non int `clk_freq` will probably case issues in others part the SoC. But we should do this to allow Modules in the future to avoid converting `sys_clk_freq` to int in the top level of the SoC.

litex/soc/software/libbase/strtod.c
enjoy-digital: OK thanks.

enjoy-digital:
Thanks for the answers and fixes, merged.

PR #502, opened 2020-05-09
diamond: project generation improvements

shuffle2:
Mostly timing closure related

### DIFF
--- a/litex/build/lattice/diamond.py
+++ b/litex/build/lattice/diamond.py
@@ -104,6 +104,9 @@ def _build_tcl(device, sources, vincpaths, build_name):
     if _produces_jedec(device):
         tcl.append("prj_run Export -impl impl -task Jedecgen")
 
+    # Cleanly close the project
+    tcl.append("prj_project close")
+
     tools.write_to_file(build_name + ".tcl", "\n".join(tcl))
 
 # Script -------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks! Do you know if there is a way to pass the proper timing constraints to Diamond for the synthesis with Synplify? (i see you are constrainting them to 200MHz). Otherwise, can you remove the f-strings refactoring (and use) from the PR? We'll probably move to f-strings at some point, but we we'll do the adaption on all builld backends at the same time.

shuffle2:
The synthesis tool (either Synplify or LSE) is provided constraints in sdc/fdc file (Synplify) or ldc file (LSE). The synth tool then propagates the constraints to the lpf so you don't need to duplicate the timing constraint listings.

During synthesis, any clock without constraint will default to 200MHz target. This has always been the case for all clocks with litex/migen, as no sdc/fdc/ldc file is created.
Post-synthesis, lpf file is used as constraints.

So far I haven't noticed a reason to deal with generating sdc/fdc/ldc file - the flow seems capable of achieving good results with user constraints provided only post-synthesis. Maybe it would be good for a future commit. It would also be nice to allow choosing between synplify/lse.

shuffle2:
I've removed the f-strings

enjoy-digital:
Thanks for the info and removing f-strings. I'll look closer at it very soon.

enjoy-digital:
Merged with minor changes: https://github.com/enjoy-digital/litex/commit/1e610600f6a088941700d6736e0621010527ef64, thanks!

PR #503, opened 2020-05-09
BIOS boot firmware from SPI with address offset

rprinz08:
Hello,

for a project [(https://github.com/rprinz08/arty-s7-multinic)](https://github.com/rprinz08/arty-s7-multinic)
I needed the FPGA bitstream and the firmware in the SPI flash. The standard LiteX BIOS tries to 
load the firmware from flash address 0 which is already used by the FPGA bitstream. I did not find
a standard LiteX way to do this so I came up with this patch.

This patch allows the specification of an address offset in the Python SoC definition (e.g.
`add_constant("FLASH_BOOT_OFFSET", 0x800000)`) from which the BIOS tries to load (or
run when no RAM available) the firmware. So for the 16MB flash in this project the first half
is used by the FPGA bitstream and the second half by the firmware.

br
Richard

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -366,6 +366,11 @@ void netboot(void)
 #endif
 
 #ifdef FLASH_BOOT_ADDRESS
+#ifdef FLASH_BOOT_OFFSET
+unsigned int flash_boot_address = FLASH_BOOT_ADDRESS + FLASH_BOOT_OFFSET;
+#else
+unsigned int flash_boot_address = FLASH_BOOT_ADDRESS;
+#endif
 
 /* On systems with exernal SDRAM we copy out of the SPI flash into the SDRAM
    before running, as it is faster.  If we have no SDRAM then we have to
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Hello @rprinz08,

thanks for the PR, in fact `FLASH_BOOT_ADDRESS` is already expected to be the address from which the binary will be copied to RAM so i'm not sure we need to add another `FLASH_BOOT_OFFSET`. For example in your project,  instead of:

````python3
self.add_constant("FLASH_BOOT_ADDRESS", self.mem_map["spiflash"])
````

you should be able to just do:
````python3
self.add_constant("FLASH_BOOT_ADDRESS", self.mem_map["spiflash"] + 0x800000)
````
Does it make sense?

The part of the PR where `memcpy` is replaced with `copy_image_from_flash_to_ram` would still be useful, do you want to rework your PR with just this? (i could also do the changes).

Regards,

Florent

enjoy-digital:
Thanks @rprinz08, that's merged, with just minor modifications: https://github.com/enjoy-digital/litex/commit/5ea3bae03609e472e7e4ab90d29cfdb94653798d.

PR #504, opened 2020-05-10
integration/soc: fix add_adapter for slaves

sergachev:
I have a SoC with 32-bit wishbone bus; it has one Module with wishbone.Interface(data_width=8) connected by add_wb_slave(). It is silently broken (compiles but does not work) with the current implementation of add_adapter() because it does not respect master and slave argument order for wishbone.Converter() it uses. This commit fixes the problem by adding an argument is_master to add_adapter() and swapping arguments to wishbone.Converter() accordingly.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -282,7 +282,7 @@ class SoCBusHandler(Module):
         return is_io
 
     # Add Master/Slave -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for catching it (adaptation on slaves was probably not tested yet). This is merged with a minor change: https://github.com/enjoy-digital/litex/commit/59d88a880ca2313cd3b8f875abcf29297209a224.

PR #505, opened 2020-05-11
Enable 1x mode on SPI flash

### LINKED ISSUE #501: Memory-mapped ECP5SPIFlash
While trying to add SPI flash support to the [_ecp5_evn_](https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/ecp5_evn.py), I noticed that the [`ECP5SPIFlash`](https://github.com/enjoy-digital/litex/blob/fbbbdf03b583a5571069fb747170e6749ad118e6/litex/soc/cores/spi_flash.py#L372) is not memory-mapped.

Does it mean no BIOS support to boot images from flash?

Is using SPI flash from ECP5 just (compared to "normal" `SpiFlash`) a matter of driving the clock from the _USRMCLK_ hardware block or is there more to it?

DurandA:
Fix #501


enjoy-digital:
Thanks!

Issue #391, opened 2020-02-20
Black Parrot cleanups


enjoy-digital:
- [ ] Fix submodule commit issue (we can't do a checkout of https://github.com/black-parrot/pre-alpha-release/tree/8aa6b6259308105872e19675c1cd5aee22283913) 
- [ ] Cleanup core.py.
- [ ] Cleanup integration.
- [x] Avoid cloning too much submodules (only https://github.com/black-parrot/pre-alpha-release and https://github.com/black-parrot/basejump_stl should be needed).

@scanakci: Could you help me with the first point? I'm not able to checkout the code you used for validation and would like to start from that. Can you update to a new version of the CPU or provide a commit that we can use for the checkout?

scanakci:
@enjoy-digital thanks for letting me know. I am currently using another commit since that version is not supporting some features required for Linux boot. Once I make it work, I will let you know.

enjoy-digital:
OK thanks, i'll wait for that then. I just did a first cleanup pass of `core.py`: https://github.com/enjoy-digital/litex/commit/53ee9a5e059b60d59c17d9d2d8a1249796695e73.

scanakci:
I made a PR which includes a version of BP that supports Linux and simulates BIOS .
https://github.com/litex-hub/pythondata-cpu-blackparrot/pull/1

Since I am using a commit from a developer branch of BP, I did not put as a submodule for now. I will add those submodules later when I make my BP version upstream.

I will make another PR for LiteX repo changes sometimes this weekend.

scanakci:
I will update the PR based on @mithro's suggestion.

scanakci:
I updated the BP version that boots up Linux with this [PR](https://github.com/enjoy-digital/black-parrot/pull/1). The recent PR](https://github.com/enjoy-digital/litex/pull/506) in LiteX repo updates the README and core.py in addition to a minor change in vivado command for properly generating bitstream for system verilog.

scanakci:
Can you please merge [PR](https://github.com/enjoy-digital/black-parrot/pull/1) for BlackParrot repo as well unless there is an issue?

enjoy-digital:
[...]

PR #506, opened 2020-05-12
Update README and core.py for Blackparrot and change vivado command for systemverilog

scanakci:
The current vivado command did not work for Blackparrot and I replaced it with a working tcl command sequence. Genesys.py generates bitstream without an issue now. Also, I cleaned README and core.py for BlackParrot.


enjoy-digital:
Thanks, merged. I'll test it on simulation and hardware soon and will see if we can close #391. Can you give more information about the issue with `read_verilog -sv` with Vivado/BlackParrot?

scanakci:
Thank you for merging. The issue mainly originates from how BlackParrot project names files. Although BP is written in system verilog, the file extensions are in "*.v" . When I use the default read_verilog -sv format, vivado got confused and were popping errors such as:

ERROR: [Synth 8-1031] bp_common_pkg is not declared [/home/scanakci/Research_sado/litex/pythondata-cpu-blackparrot/pythondata_cpu_blackparrot/system_verilog/bp_me/src/include/v/bp_cce_pkg.v:12]

Reading as a verilog file and setting the file to a systemverilog fixed the issue. Since I am not very familiar with tcl commands and vivado, I could not reason why directly reading with -sv caused trouble.

PR #509, opened 2020-05-12
RFC: improve critical path timing due to wishbone bridge

bunnie:
As a design gets larger (lots of CSRs) the critical path
to decode addresses drives the clock rate. The current
wishbone bridge implementation puts a combinational path from
the UART PHY, setting an address, adding an offset to it,
then propagating that through the rest of wishbone bus.
This adds an extra few logic levels to what is the slowest
and most routed path on the SoC.

This patch proposes adding a wait state to the wishbone bridge,
between the RECEIVE_DATA, and the WRITE_DATA phases. This
will slow down the latency of the bridge by one clock cycle,
but this also breaks the combinational path by allowing us to
register the results of the address and data before passing
it on to the rest of the bus.

I think this is OK because the wishbone bridge is often
used via e.g. UART or Ethernet, and an extra 10ns or so
to process a packet should be negligible impact, but there
may be other use cases I am unaware of, hence the RFC.

Also, I am just eyeballing the state machine to insert the
delay. This version does work on real hardware, but there
may be other subtle paths in the wishbone bridge state
machine that are broken by inserting this sync delay,
also hence the RFC.

However, if you can accept this patch, I think it will
improve timing closure and compile times on large/congested
designs that also include the debug bridge.


enjoy-digital:
Thanks @bunnie for the PR, i was planning to refresh/rewrite this core. Maybe i'll do that now then, to ease adding your findings/improvements and avoid breaking them later.

enjoy-digital:
@bunnie can you test with https://github.com/enjoy-digital/litex/commit/873d95e517b3e6074c3b07502c4a5f05aef09bf4? The Wishbone address offset is now computed synchronously and this should improve the timings without impacting too much resource usage. (While doing the changes i was testing iCEBreaker and noticed a ~20 LUTs increase).

bunnie:
👍 looks great. The slack histogram is even better, and the wishbone bridge seems to be working fine. Thank you for looking at this!

enjoy-digital:
Thanks for the feedback, i'm closing this since this should be good now. Please re-open if you have timings issue with the new version in the future.

PR #510, opened 2020-05-12
Colorlight usb

mubes:
Trivial PR to add bootaddr option for ecppack on ecp5 to allow multi-image booting.


enjoy-digital:
Thanks, it seems the part where `bootaddr` is passed to `ecppack` is missing.

mubes:
Well, that was daft!  On testing I'd overridden this with my own 
template anyway without realising, so it worked perfectly for me :-/

All fixed now. Good catch.

enjoy-digital:
Thanks for the update, merged.

PR #511, opened 2020-05-12
Improve standalone cores

ozbenh:
These are some improvements for the generation of standalone cores without CPUs such
as LiteDRAM

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -781,7 +781,13 @@ class SoC(Module):
         for n, (origin, size) in enumerate(self.cpu.io_regions.items()):
             self.bus.add_region("io{}".format(n), SoCIORegion(origin=origin, size=size, cached=False))
         self.mem_map.update(self.cpu.mem_map) # FIXME
-        self.csr.update_alignment(self.cpu.data_width)
+
+        # We don't want the CSR alignemnt reduced from 64-bit to 32-bit on
+        # a standalone system with a 64-bit WB and no CPU.
+        # Should we instead only update alignment if the CPU is *bigger*
+        # than the CSR ?
+        if name != "None":
+            self.csr.update_alignment(self.cpu.data_width)
         # Add Bus Masters/CSR/IRQs
         if not isinstance(self.cpu, cpu.CPUNone):
             if reset_address is None:
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -824,7 +824,14 @@ class SoC(Module):
         # SoC Bus Interconnect ---------------------------------------------------------------------
         bus_masters = self.bus.masters.values()
         bus_slaves  = [(self.bus.regions[n].decoder(self.bus), s) for n, s in self.bus.slaves.items()]
-        if len(bus_masters) and len(bus_slaves):
+        # One master and one slave, use a point to point interconnect, this is useful for
+        # generating standalone components such as LiteDRAM whose external control
+        # interface is a wishbone.
+        if len(bus_masters) == 1 and len(bus_slaves) == 1:
+            self.submodules.bus_interconnect = wishbone.InterconnectPointToPoint(
+                master = list(bus_masters)[0],
+                slave  = list(self.bus.slaves.values())[0])
+        elif len(bus_masters) and len(bus_slaves):
             self.submodules.bus_interconnect = wishbone.InterconnectShared(
                 masters        = bus_masters,
                 slaves         = bus_slaves,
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks, it's merged with some changes: https://github.com/enjoy-digital/litex/commit/3a6dd95d6f89c12960a29f58bcfd8e5fbb161275. I don't think it's necessarily a good idea to use a `PointToPoint` Interconnect or adapt the Wishbone address width to the CSR  Bus since these changes are specific to the use case and will break others, it's probably better to let the synthesis tool do the optimization here automatically.

ozbenh:
On Tue, 2020-05-12 at 07:28 -0700, enjoy-digital wrote:
> Thanks, it's merged with some changes: 3a6dd95. I don't think it's
> necessarily a good idea to use a PointToPoint Interconnect or adapt
> the Wishbone address width to the CSR Bus since these changes are
> specific to the use case and will break others, it's probably better
> to let the synthesis tool do the optimization here automatically.

Except it doesn't work that way unfortunately, at least with the stuff
that's in git right now.

The main problem is that since LiteDRAM creates the exported wishbone
interface with an address_size matching the CSR address size, we lose
all the top bits. But the SharedInterconnect uses them for address
decoding, so all CSR accesses fail (I think Vivado even optimizes out
the entire CSR bus as it realizes that the address decode can never
match, the resulting non-working bitfile is ... small).

The "obvious" fixes are problematic:

 - Setting CSR base to 0 means all the csr.h definitions have the wrong
address, unless I fold that base address into the accessors themseslves
but that leads to other problems if I try to use more than one csr-
using module such as litdram and liteeth. In fact, even then, I'm not
sure if that will help, what is the result of comparing unwired bits of
a vector to 0 ?

 - I could make LiteDRAM create a full size address bus instead of one
that only has the bits necessary for the CSR bus. The problem here then
becomes two folds:

    * This does create extra hardware and wires, and the Arty is really
struggling with all those big busses going around. We get the extra
comparators of the address decode in the timing path.

    * The way we are changing our interconnect in Microwatt is that the
"IO" bus where all the dumb slaves like the LiteDRAM CSRs live will
only have a subset of the address bits going down to it. IE. We have a
"master" address decoder that handles the fast path (core -> bram/dram)
combinationally, and an IO leg that has pipeline buffers, and goes to
IOs like CSRs, UART, etc.... This helps a LOT with timing. However I
have only about 24 bits of address going down that bus (again to
relieve pressure on the Vivado router), so having the address decoder
based on the full 32-bit address inside LiteDRAM breaks.

Generally speaking, when creating a standalone IO core like this, we
really don't need any fancy interconnect with address decode in it.
Well it still has enough decoding for the actual CSRs but that's on the
CSR bus itself.

So I see two ways out of this quagmire:

One which Im not sure how to do yet since I don't have much experience
with that part of LiteX, would be to be able to specify separate
"csr_base" addresses for the generated HW vs generated code. That would
allow to leave the HW one at 0 and add an offset to csr.h. I would
*also* have to extend I think the exported wishbone to the full 32-bits
as to be able to match addresses properly even against 0 unless I can
also reduce the size of the intermediary wishbone.

We would still get the extra HW for the SharedInterconnect & address
decoder, but with some hope, the synth tool would get rid of most of it
(fingers crossed) since it would just unwired compare bits to 0... but
maybe not. Unclear. Unless we still have to route the whole bus up and
force the bits to 0 "upstream".

The other approach would be to keep the option of generating P2P
interconnect. But instead of doing it implicitly based on #masters &
#slaves being both 1, make it explicit via a parameter that
litedram_gen can set. That or make the implicit case even more tight:
#masters=1,#slaves=1,cpu=None for example. Out of curiosity, what use
cases are broken here ?

I've attached a LiteX patch the does the second approach (explicit
parameter) along with the corresponding LiteDRAM patch. What do you
think ?

Cheers,
Ben.

ozbenh:
On Tue, 2020-05-12 at 07:28 -0700, enjoy-digital wrote:
> Thanks, it's merged with some changes: 3a6dd95. I don't think it's
> necessarily a good idea to use a PointToPoint Interconnect or adapt
> the Wishbone address width to the CSR Bus since these changes are
> specific to the use case and will break others, it's probably better
> to let the synthesis tool do the optimization here automatically.


Another small nit. You changed my updating of self.mem_map["csr"]
to instead do this:

        self.add_csr_bridge(self.mem_map["csr"] if csr_base is None else csr_base)

This works, however the log printed when generating still has the
0x82000000 base all over rather than the specified one. Not a huge deal
but can be confusing esp. when debugging.

Cheers,
Ben.

PR #513, opened 2020-05-12
Bios linker edits to prevent inappropriate optimisation

mubes:
Small change to prevent LTO eating sections we don't want disappearing in our target image. 

I'm not comfortable that I understand the full detail of the software ecosystem here yet, but this
change should be conservative anyway as there aren't too many circumstances where you'd want
your bootstrap or ISRs disappearing.


enjoy-digital:
Thanks.

Issue #145, opened 2019-01-23
crt0 doesn't copy data section into RAM?


mithro:
There seems to be some confusion around if the data section should be copied to RAM.

If the program is executing from a read only location (ROM, spiflash, etc) - then the crt0 needs to copy the data section into RAM.
If the program is executing from a read/write location (SRAM, DDR, etc) - then the crt0 doesn't need to do anything (as the data section can be modified in place).

Both [`picorv32`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-picorv32.S#L197-L209) and [`lm32`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-lm32.S#L115-L131) have a `EXECUTE_IN_PLACE` define which does the data section copy.

However, [`vexriscv`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-vexriscv.S) and [`or1k`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-or1k.S) and [`minerva`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/crt0-minerva.S) are missing this.

Naming this `EXECUTE_IN_PLACE` is kind of misleading, as the reason the copy needs to occur is because the place it is executing from is read only.

mithro:
@xobs @cr1901 - FYI.

cr1901:
It's missing for @vexriscv, @or1k and @minerva because I haven't implemented it yet :).

>the reason the copy needs to occur is because the place it is executing from is read only.
I think it's more subtle than that. You can in principle have a read only code section in SPI flash, but e.g. SRAM is initialized during bitstream loading. No copy is needed in that case either.

I mainly used `EXECUTE_IN_PLACE` as a shorthand for "code and data do not live in the same section of address space from the CPU's point-of-view." Open to suggestions to change it (as well as `crt0-xip.S`/`crt0-ctr.S`).

mithro:
I would suggest that we need something more like `EXECUTING_FROM_RO_REGION` and `EXECUTING_FROM_RW_REGION`?

ewenmcneill:
I randomly found this issue down a chain of links; some thoughts off the top of my head (as I remember discussing this when the first TinyFPGA BX MicroPython bringup was happening, which I think led to the existing macro name...):

[...]

PR #514, opened 2020-05-12
Rommable linker files and startup

mubes:
Not sure if this is useful, but its not something I found already....feel free to bounce it if it's not something you need, but this adds rommable code boot for vexriscv.  For any individual project folks will need to edit this file to their needs, but at least it's a working starting point (it's used in tinyuf2, for example).


enjoy-digital:
Thanks @mubes, this is indeed useful, but i'm not sure we should include it directly in LiteX (it will be difficult to maintain and if we add it, then users will also want to have it for the others CPU and for the others configurations :)). Maybe we could add a page in the Wiki explaining how to run the CPU code from ROM with a link to a useful crt0/linker for that?

mubes:
Hiya, I think that would be fine provided we put a comment in the 'regular' linker file pointing people to the rommable version.  It is however a bit unusual to not have rommable boot up code directly available in the SDK for an embedded system, most programmers would expect to find ram-run and rom-run ld and startup directly available.  

It looks like this is already partially recognised in #145 ... give me a little while and I'll see if I can align this with that, then it would just be or1k and minerva that are missing it.

enjoy-digital:
@mubes:  we can indeed improve the situation but we should make sure to avoid as much as duplication as possible and if supported try to make it available to all CPUs. We also need to add a test or/and describe how to test it for maintenance. If support is added to one CPU with a test or instruction to test, i could add support for the others CPUs.

mubes:
OK, I just pushed a small update to align this with the other CPUs that already have `EXECUTE_IN_PLACE` support. That's allowed unification of the crt, but a separate ld file is still needed to put the sections into rom.

I have tested this in tinyUF2 and its fine but I don't really know how to slot a test case into litex for this because it needs a whole build that is tested on a CPU.  The actual code for testing is straightforward;

```#include <stdio.h>

uint32_t tv=42;

void main(void)

{
  if (42!=tv)
    {
      printf("Initialisation of data segment failed\n");
      return;
    }

  tv = 16;
  if (16!=tv)
    {
      printf("Cannot write to data segment\n");
      return;
    }

  printf("Data test passed\n");
}
```
...but I'm not sure how to incorporate it into the existing framework?
DAVE

enjoy-digital:
@mubes: sorry i'm not sure i understood correctly the issue in your initial post. This is now fixed by https://github.com/enjoy-digital/litex/issues/566 and copy of data section is included in the crt0 of all CPUs.

PR #516, opened 2020-05-13
Add I2S support to Arty

mateusz-holenko:
This PR extends I2S module capabilities and adds `SoundSoC` configuration for Arty.

### DIFF
--- a/litex/soc/cores/i2s.py
+++ b/litex/soc/cores/i2s.py
@@ -1,35 +1,55 @@
 # This file is Copyright (c) 2020 bunnie <bunnie@kosagi.com>
+# This file is Copyright (c) 2020 Antmicro <www.antmicro.com>
 # License: BSD
 
 from migen.genlib.cdc import MultiReg
 
+from litex.soc.cores.clock import *
 from litex.soc.interconnect import wishbone
 from litex.soc.interconnect.csr_eventmanager import *
 
 from litex.soc.integration.doc import AutoDoc, ModuleDoc
+from enum import Enum
+import math
+class I2S_FORMAT(Enum):
+    I2S_STANDARD = 1
+    I2S_LEFT_JUSTIFIED = 2
 
-
-class S7I2SSlave(Module, AutoCSR, AutoDoc):
-    def __init__(self, pads, fifo_depth=256):
+class S7I2S(Module, AutoCSR, AutoDoc):
+    def __init__(self, pads, fifo_depth=256, master=False, concatenate_channels=True, sample_width=16, frame_format=I2S_FORMAT.I2S_LEFT_JUSTIFIED, lrck_ref_freq=100e6, lrck_freq=44100, bits_per_channel=28):
         self.intro = ModuleDoc("""Intro
 
-        I2S slave creates a slave audio interface instance. Tx and Rx interfaces are inferred based
-        upon the presence or absence of the respective pins in the "pads" argument.
+        I2S master/slave creates a master/slave audio interface instance depending on a configured master variable. 
+        Tx and Rx interfaces are inferred based upon the presence or absence of the respective pins in the "pads" argument.
+        
+        When device is configured as master you can manipulate LRCK and SCLK signals using below variables.
+        
+        - lrck_ref_freq - is a reference signal that is required to achive desired LRCK and SCLK frequencies.
+                         Have be the same as your sys_clk.
+        - lrck_freq - this variable defines requested LRCK frequency. Mind you, that based on sys_clk frequency, 
+                         configured value will be more or less acurate.
+        - bits_per_channel - defines SCLK frequency. Mind you, that based on sys_clk frequency, 
+                         the requested amount of bits per channel may vary from configured.
+        
+        When device is configured as slave I2S interface, sampling rate and framing is set by the
+        programming of the audio CODEC chip. A slave configuration defers the
+        generation of audio clocks to the CODEC, which has PLLs specialized to generate the correct
+        frequencies for audio sampling rates.
 
-        The interface is I2S-like, but note the deviation that the bits are justified left without a
-        1-bit pad after sync edges. This isn't a problem for talking to the LM49352 codec this was
-        designed for, as the bit offset is programmable, but this will not work well if are talking
-        to a CODEC without a programmable bit offset!
+        I2S core supports two formats: standard and left-justified.
+        
+        - Standard format requires a device to receive and send data with one bit offset for both channels. 
+            Left channel begins with low signal on LRCK.
+        - Left justified format requires from device to receive and send data without any bit offset for both channels.
+            Left channel begins with high signal on LRCK.
+    
+        Sample width can be any of 1 to 32 bits.
+
+        When sample_width is less than or equal to 16-bit and concatenate_channels is enabled,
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @mateusz-holenko, we will have to discuss/change a few things for this PR:
- the Arty does not have a native I2S interface, i imagine you are using a PMOD module, the pins definitions should then use the corresponding `pmod` connector.
- i'd like to limit the scope of supported features in the LiteX platforms/targets and here a custom system/application is created by assembling a PMOD on the Arty, so this should probably be maintained externally.  If you are planning to use this with Linux-On-LiteX-Vexriscv, that's probably something that we could integrate there.
- This add a second PLL for system that are not using `SoundSoC`, which is not optimal and could break others design build from BaseSoC.
- The prefered approach now is to avoid specialized SoCs (`SoundSoC`) and just use a `add_feature` method  (`add_sdram`, `add_ethernet`, etc...) for example.  A `add_i2s` method could be use that could add the PLL + I2S core.
- For these kind of PRs, i would appreciate having more context about what you are building, the hardware that is used (the PMOD module that is used here) and the steps to reproduce and test it.

enjoy-digital:
As discussed with @kgugala, the best for now would probably be to integrate the litex/soc/cores/i2s.py changes and keep the I2S integration specific to your repository (IIRC https://github.com/antmicro/litex-vexriscv-tensorflow-lite-demo), we will also probably have to ask @bunnie to review the I2S changes. While working on this, have you created some testbenchs that could be reused/integrated as tests?

mateusz-holenko:
@enjoy-digital Thanks for commetns - we are working on them. In the meantime some more details:

We developed & tested it using https://reference.digilentinc.com/reference/pmod/pmodi2s2/reference-manual connected to Arty/Artix7-35t.

As a software we used Zephyr (with our driver) running a sample I2S audio transceiver application (both currently available at: https://github.com/antmicro/zephyr/tree/i2s_zephyr_support). We connected a 1KHz sinusoidal source to the input and observed the output using audacity software.

Details about the sample app can be found at: https://github.com/antmicro/zephyr/commit/ae9b99aed18372794205a000df6f745ba5c0cf71#diff-dcd780700b17fda05b040838fdd3afa3.

We don't have any automated testbenches to reuse.

mateusz-holenko:
@enjoy-digital I updated the branch so it contains changes related to I2S module only.

@bunnie Could you take a look at our changes?

litex/soc/cores/i2s.py
```diff
@@ -413,53 +538,85 @@ def __init__(self, pads, fifo_depth=256):
...
+                    NextValue(tx_buf, Cat(tx_rd_d, offset)),
+                    tx_rden.eq(1),
                 )
             )
             txi2s.act("LEFT",
                 If(~self.tx_ctl.fields.enable,
                     NextState("IDLE")
                 ).Else(
-                    NextValue(tx_pin, tx_buf[31]),
+                    NextValue(tx_pin, tx_buf[sample_width - 1]),
```
bunnie, line 568: if concatenate_channels is true, in this case, you're reading from tx_buf[15] when sample_width is 16. This causes left data to go to the right channel and then the left channel gets silence. 

Suggest adding a new variable `sample_msb` that is set to `sample_width * 2` and using it when the indexing into tx_buf.

litex/soc/cores/i2s.py
```diff
@@ -413,53 +538,85 @@ def __init__(self, pads, fifo_depth=256):
...
+                        )
+                    )
+                )
+
             txi2s.act("RIGHT",
                 If(~self.tx_ctl.fields.enable,
                     NextState("IDLE")
                 ).Else(
-                    NextValue(tx_pin, tx_buf[31]),
+                    NextValue(tx_pin, tx_buf[sample_width-1]),
```
bunnie, line 619: same problem here of indexing into the wrong part of the buffer for the case of concatenate_channels

litex/soc/cores/i2s.py
```diff
@@ -1,36 +1,54 @@
...
+class S7I2S(Module, AutoCSR, AutoDoc):
+    def __init__(self, pads, fifo_depth=256, master=False, concatenate_channels=True, sample_width=16, frame_format=I2S_FORMAT.I2S_LEFT_JUSTIFIED, lrck_ref_freq=100e6, lrck_freq=44100, bits_per_channel=28):
         self.intro = ModuleDoc("""Intro
 
-        I2S slave creates a slave audio interface instance. Tx and Rx interfaces are inferred based
-        upon the presence or absence of the respective pins in the "pads" argument.
+        I2S master/slave creates a master/slave audio interface instance depending on a configured master variable. 
+        Tx and Rx interfaces are inferred based upon the presence or absence of the respective pins in the "pads" argument.
+        
+        When device is configured as master you can manipulate LRCK and SCLK signals using below variables.
```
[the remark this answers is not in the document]
mateusz-holenko: fixed

litex/soc/cores/i2s.py
```diff
@@ -1,36 +1,54 @@
...
+                         configured value will be more or less acurate.
+        - bits_per_channel - defines SCLK frequency. Mind you, that based on sys_clk frequency, 
+                         the requested amount of bits per channel may vary from configured.
+        
+        When device is configured as slave I2S interface, sampling rate and framing is set by the
+        programming of the audio CODEC chip. A slave configuration defers the
+        generation of audio clocks to the CODEC, which has PLLs specialized to generate the correct
+        frequencies for audio sampling rates.
+
+        I2S core supports two formats: standard and left-justified.
```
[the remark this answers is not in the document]
mateusz-holenko: fixed

litex/soc/cores/i2s.py
```diff
@@ -164,13 +221,20 @@ def __init__(self, pads, fifo_depth=256):
...
                     CSRField("empty",     size=1, description="No data available in FIFO to read"), # next flags probably never used
                     CSRField("wrcount",   size=9, description="Write count"),
                     CSRField("rdcount",   size=9, description="Read count"),
-                    CSRField("fifo_depth", size=9, description="FIFO depth as synthesized")
+                    CSRField("fifo_depth", size=9, description="FIFO depth as synthesized"),
+                    CSRField("concatenate_channels", size=1, reset=concatenate_channels, description="Receive and send both channels atomically")
+                ])
+            self.rx_conf = CSRStatus(description="Rx configuration",
+                fields=[
+                    CSRField("format", size=2, reset=frame_format.value, description="I2S sample format"),
```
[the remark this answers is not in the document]
mateusz-holenko: changed

litex/soc/cores/i2s.py
mateusz-holenko: As suggested we introduced `sample_msb` variable which is equal to `fifo_data_width - 1`.

For concatenated channels `fifo_data_width` is equal to `sample_width * 2`, so it should work as expected.

litex/soc/cores/i2s.py
mateusz-holenko: fixed, as described above

bunnie:
I have checked out your latest commit and built and tested it, it works now with our system too. thanks!

@enjoy-digital I'm fine for this to be merged.

enjoy-digital:
Thanks @mateusz-holenko, @bunnie. Merged.

PR #517, opened 2020-05-13
csr: Rework accessors

ozbenh:
Have all the new compound accessors be written in terms of the simple
ones and fix how CSR_ACCCESORS_DEFINED can be used to override the
simple ones but keep the definitions of the other ones around.

This *should* also also fix incorrect multiple accesses done
by  64-bit CPUs to 32-bit CSR busses, and make the accessors not
depend on CONFIG_CSR_ALIGNMENT being the same as sizeof(unsigned long)*8

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>
Reviewed-by: Gabriel Somlo <gsomlo@gmail.com>

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -198,10 +198,8 @@ def get_csr_header(regions, constants, with_access_functions=True):
     r += "#ifndef __GENERATED_CSR_H\n#define __GENERATED_CSR_H\n"
     if with_access_functions:
         r += "#include <stdint.h>\n"
-        r += "#ifdef CSR_ACCESSORS_DEFINED\n"
-        r += "extern void csr_write_simple(unsigned long v, unsigned long a);\n"
-        r += "extern unsigned long csr_read_simple(unsigned long a);\n"
-        r += "#else /* ! CSR_ACCESSORS_DEFINED */\n"
+        r += "#include <system.h>\n"
+        r += "#ifndef CSR_ACCESSORS_DEFINED\n"
         r += "#include <hw/common.h>\n"
         r += "#endif /* ! CSR_ACCESSORS_DEFINED */\n"
     csr_base = regions[next(iter(regions))].origin
--- a/litex/soc/software/include/hw/common.h
+++ b/litex/soc/software/include/hw/common.h
@@ -2,6 +2,7 @@
 #define __HW_COMMON_H
 
 #include <stdint.h>
+#include <system.h>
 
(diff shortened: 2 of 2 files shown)


enjoy-digital:
Thanks @ozbenh and @gsomlo for the review. @xobs: would you mind checking it is also ok for you and that it will not break your use cases?

xobs:
I'm okay with this change.  Others may want to make sure this doesn't add more indirection to their accessor functions, but this shouldn't affect anything I'm using it for.

ozbenh:
It shouldn't add indirection in the normal case as csr_*_simple() are inline.

The CSR_ACCCESORS_DEFINED case is indirect, bit that was already the case before. I'm just fixing it.

We could make it inline too by adding a way to #include something up there, maybe "system.h", and remove the extern declarations (in generated csr.h as well). But I don't see that being terribly urgent at this point.

ozbenh:
More precisely, would adding #include <system.h> at the top of hw/common.h and the generated csr.h be considered a problem ?

If yes, what about creating an empty-by default cpu/csr-overrides.h in all cpus dirs which I can then use from microwatt to set CSR_ACCESSORS_DEFINED (and if necessary CSR_BASE if that commit gets in). ? I can do a patch, I just want feedback on the approach.

enjoy-digital:
Including `<system.h>` in `hw/common.h ` should not cause issues, this would avoid having a new `.h` file for the CSR accessors. Can you do a test with this?

ozbenh:
I would have to also add it to the generated csr.h as things like sdram.c include csr.h before system.h, I'll test that later tonight hopefully

xobs:
One concern I do have with this is that it removes the `CSR_ACCESSORS_DEFINED` macro that is used to enable seamless development on a desktop system via etherbone.  This is described in https://github.com/litex-hub/wishbone-utils/tree/master/libeb-c

Does this still allow us to define `-DCSR_ACCESSORS_DEFINED` and use the generated `csr.h` file locally?

ozbenh:
Yes, it does, this is what I do with microwatt when using a generated litedram. However i did change things a bit (check my latest update).  You need to provide a system.h which will provide the
simple accessors (as inline or extern definitions), can set CSR_ACCESSORS_DEFINED (though you can still use -D for that) and can set CSR_BASE if you want to apply an offset.

xobs:
That seems to work fine for what I use it for.

ozbenh:
Thanks !

enjoy-digital:
Thanks, merged.

PR #518, opened 2020-05-13
export: add define of CSR_BASE if not already defined and use it for …

enjoy-digital:
…CSRs definitions/accesses.

This will allow more flexibility when integrating standalone cores.


enjoy-digital:
@ozbenh: thanks for the feedback, merged.

PR #521, opened 2020-05-13
Fix fw bios command

DurandA:
Count of `fw <offset> <value> [count]` was parsed incorrectly.


enjoy-digital:
Thanks, @rprinz08 already fixed it yesterday with https://github.com/enjoy-digital/litex/commit/1f55fcf4496652497f717c6407922ad5bba49d82!

PR #524, opened 2020-05-14
Spi autotuner

kowalewskijan:
Autotuner for LiteSPI in BIOS:
- Frequency autotuner
- Configuration of dummy bits

### DIFF
--- a/litex/soc/software/bios/Makefile
+++ b/litex/soc/software/bios/Makefile
@@ -24,6 +24,7 @@ OBJECTS = isr.o			\
 	  cmd_mem_access.o	\
 	  cmd_sdcard.o		\
 	  cmd_spi_flash.o	\
+	  spi.o			\
 
 ifneq "$(or $(TERM_NO_COMPLETE),$(TERM_MINI))" ""
 CFLAGS += -DTERM_NO_COMPLETE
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks, this has been merged manually and i moved the code a bit. I'm going to do some test with it and will report the results.

enjoy-digital:
I've been able to integrate and test it with: 
- https://github.com/enjoy-digital/litex/commit/80eca3000b26dda547f58125d69010ef0026dab2
- https://github.com/litex-hub/litespi/commit/a47bce51c179a69dfe2cb7ca0589c797b97dd3a9
- https://github.com/litex-hub/litespi/commit/c178db69018c0423a788c867fd74c9317af601c9
- https://github.com/litex-hub/litespi/commit/02eb1d7f9e1072526c475698b1bda468f0586cf3

Since things are still moving a bit, it's probably better to prototype the LiteSPI integration directly in LiteSPI repository. I added an example for Arty, you could add another example for the target you are using but the Arty target will have to be updated to ease testing.

The only dependency you should have now in LiteX to develop the core should be the software. I'm thinking about moving `liblitespi ` to LiteSPI directory to reduce this dependency, but i still have to think about this, so for now you can just send PR for the software once you update things in LiteSPI.

PR #526, opened 2020-05-15
Make booting from SD-Card to behave same as from SPI flash

rprinz08:
When trying to boot a firmware BOOT.BIN from SD-card using the Linux variant of the VexRiscv CPU, the BIOS tries to load a Linux Image - which is OK. But if it does not find one it gives up instead of trying to load a firmware BOOT.BIN instead (same as when booting from SPI flash).
This patch makes booting from SD-card behaving the same as from SPI flash.

br
Richard

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -512,19 +512,34 @@ void spisdcardboot(void)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, we should indeed try to harmonize the behavior of the boot options, that's a good first step.

PR #527, opened 2020-05-15
timer: count total cycles

mateusz-holenko:
This adds a set of two CSRs allowing to capture and read
the total cycles count measured from reset.

Having this will allow to implement LiteX support in Zephyr easier (as described in https://github.com/zephyrproject-rtos/zephyr/issues/23622).

### DIFF
--- a/litex/soc/cores/timer.py
+++ b/litex/soc/cores/timer.py
@@ -65,6 +65,8 @@ class Timer(Module, AutoCSR, ModuleDoc):
         self.ev.zero       = EventSourceProcess()
         self.ev.finalize()
 
+        self._total_cycles_reg = CSRStatus(64, description="""Total number of cycles since start.""")        
+        self._update_total_cycles = CSRStorage(1, description="""Update trigger for the current total cycles""")
         # # #
 
         value = Signal(width)
@@ -81,4 +83,7 @@ class Timer(Module, AutoCSR, ModuleDoc):
             ),
             If(self._update_value.re, self._value.status.eq(value))
         ]
+        total_cycles_value = Signal(width)
+        self.sync += total_cycles_value.eq(total_cycles_value + 1)
+        self.sync += If(self._update_total_cycles.re, self._total_cycles_reg.status.eq(total_cycles_value))
         self.comb += self.ev.zero.trigger.eq(value != 0)


enjoy-digital:
@mateusz-holenko: Since i'm assuming it's for VexRiscv CPU, would it be possible to use the VexRiscvTimer? https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/vexriscv/core.py#L62-L78 In your target, you just have to do `cpu.add_timer()` or we could create a zephyr variant. Adding these registers/counter has a cost and this would need to be optional for small targets if we add it, so i would like to think about it. Would using the VexRiscTimer work for you?

mateusz-holenko:
@enjoy-digital You are correct - currently we support `LiteX/VexRiscv` in Zephyr. However, we might add support for other CPUs as well - in this context having one type of timer for all boards would be better.

I agree that the proposed extension comes with a cost that might be too big for small targets.

What about making those new CSRs optional and disabled by default? That way current platforms would be untouched and we could just add a new variant for Zephyr having an extended timer?

enjoy-digital:
@mateusz-holenko: i just added and optional `uptime` counter to the SoCController with https://github.com/enjoy-digital/litex/commit/82364de57f0a3759665ae20a7408a6fef8ce9f58, you can enable it in your SoC by setting `ctrl_uptime` to `True`. Can you check this is fine for your use case?

mateusz-holenko:
@enjoy-digital If I understand correctly, your solution does exactly the same what we proposed, but puts the logic into a differnt module, right?

TBH from the Zephyr's driver perspective it's more convinient and cleaner to have it on one place. This way we could have a single timer playing a role of a systick. With the solution proposed by you we need to access two independent timers from one driver to get the same functionality.

xobs:
For what it's worth, we also have a TickTimer in Betrusted (https://github.com/betrusted-io/gateware/blob/master/gateware/ticktimer.py docs at https://ci.betrusted.io/betrusted-soc/doc/ticktimer.html) that is useful as an OS timer,  It's not designed to be quite as fast, but is designed to be long-running.

enjoy-digital:
@mateusz-holenko: i'm not sure the uptime cycles should be part of the current Timer. Both are used for different purposes (and we could have multiple timers and only one uptime counter). So i have a preference for a different Timer like @xobs suggests or a the one i added to the SoCController.

enjoy-digital:
@mateusz-holenko: i don't want to make your work harder and understand your concerns, so just implemented your solution with a `add_uptime` method with https://github.com/enjoy-digital/litex/commit/be25500e9173b0e7c0ec6e8d5f626cc7f52a9484, can you check this is ok for you? With the current behaviour, the uptime counter starts at 0 on power up (is when the board is powered up and/or the the bitstream is loaded) and will never be reseted, is is the behavior you want, or do you want this counter to be reseted on CPU reset?

mateusz-holenko:
@enjoy-digital Thanks for including the uptime counter feature in `timer.py` - it'll make fixing a Zephyr driver much easier! 

I totally understand your concern about having a single uptime timer in the system - with this solution it's also possible so I hope this is an acceptable solution for everyone.

As for the resetting, I don't think this is necessary for the Zephyr scenario to set the value back to 0 on CPU reset - the current implementation looks good enough. We'll now test it with the driver.

enjoy-digital:
Thanks @mateusz-holenko for the feedback. I'm closing this since this should now be fine, but please re-open a specific issue if you need adjustments.

PR #531, opened 2020-05-19
software/bios: fix link order to avoid undefined symbol errors

gsomlo:
move `-lbase-nofloat` (providing `busy_wait()` after `-lliteeth` and `-llitesdcard` (which *use* `busy_wait()` to avoid "undefined symbol" error from `$(LD)`.


enjoy-digital:
Thanks for catching/fixing this.

PR #536, opened 2020-05-21
Integrate micron memory model simulation

mglb:
Issues:
* The simulation is hardcoded in litex_sim; should be optional
* Clocker should be configurable to allow additional clocks which are faster than sys_clk.
* Simulation time resolution is changed to 1ps, with 125ps increments in each loop. This will cause unnecessary performance drop in simulations which don't need 800MHz clock. Should be configurable

LiteDRAM PR: https://github.com/enjoy-digital/litedram/pull/196

### DIFF
--- a/litex/build/sim/config.py
+++ b/litex/build/sim/config.py
@@ -24,7 +24,7 @@ class SimConfig():
         return new
 
     def add_clocker(self, clk):
-        self.add_module("clocker", [], clocks=clk, tickfirst=True)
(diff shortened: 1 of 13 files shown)


enjoy-digital:
Thanks for the work but i don't think this should be directly integrated in LiteX/LiteDRAM since too specific and complicated to maintain. I'm sure this work will be useful later and we this is a project someone could maintain externally.

PR #539, opened 2020-05-24
Fix UART startbit: 1 cycle later

dayjaby:
So far, the startbit for UART is 1 cycle longer than all the other UART bits. With this PR, the startbit is written one cycle later.

VCD files for both scenarios are attached.

[uart_startbit_vcd.zip](https://github.com/enjoy-digital/litex/files/4673690/uart_startbit_vcd.zip)


enjoy-digital:
Thanks @dayjaby for the report and fix! That's merged but i implemented the fix a bit differently with https://github.com/enjoy-digital/litex/commit/78a9579e0996d7c9d19916cd6aef4486ca854ebd.

PR #541, opened 2020-05-27
Add support for I2C to read SPD EEPROM

jedrzejboczar:
Requires https://github.com/litex-hub/litex-boards/pull/78 and https://github.com/enjoy-digital/litedram/pull/199.

This PR implements bitbang I2C and adds a command for reading the contents of DRAM SPD EEPROM.

We have 3 generic I2C commands (`i2creset`, `i2cwr` and `i2crd`) and one specific command for reading SPD EEPROM (`spdread`). At first I wanted to have only the `spdread` command, but after some tests on ZCU104 board it seems that we need to give the user some flexibility. For example on ZCU104 there is an I2C address translation chip, which has to be configured first to enable the channel that is connected to the SPD EEPROM I2C. This is a board specific thing, so, depending on the board, user will be able to use `spdread` directly or will have to first do some configuration using `i2cwr`.

To be able to use these commands it is enough to have an I2C peripheral (`#ifdef CSR_I2C_BASE`).
 
This is an example from ZCU104 (with https://github.com/litex-hub/litex-boards/pull/78). First we configure I2C switch chip to enable channel 7:
```
litex> spdread 1      
Error when reading SPD EEPROM

litex> i2cwr 0x74 0x80

litex> spdread 1      
Memory dump:
0x00000000  23 11 0c 03 45 21 00 08 00 60 00 03 02 03 00 00  #...E!...`......
0x00000010  00 00 06 0d f8 ff 01 00 6e 6e 6e 11 00 6e f0 0a  ........nnn..n..
0x00000020  20 08 00 05 00 f0 2b 34 28 00 78 00 14 3c 00 00   .....+4(.x..<..
0x00000030  00 00 00 00 00 00 00 00 00 00 00 00 16 36 0b 35  .............6.5
0x00000040  16 36 0b 35 00 00 16 36 0b 35 16 36 0b 35 00 00  .6.5...6.5.6.5..
0x00000050  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x00000060  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x00000070  00 00 00 00 00 00 9c b5 00 00 00 00 e7 00 60 5b  ..............`[
0x00000080  0f 01 02 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x00000090  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000a0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000b0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000c0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000d0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000e0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0x000000f0  00 00 00 00 00 00 00 00 00 00 00 00 00 00 c0 e2  ................
```

This output can be then saved in a file and loaded using the function added in https://github.com/enjoy-digital/litedram/pull/199 (but will fail after the initial parsing because only DDR3 SPD parsing has been implemented until now). I am not sure if this is the best format, but it seemed to be the easiest for the user to just copy the output and be able to save it in a file and load as it is. In `litex_sim` we can load the SPD data using the `--sdram-from-spd-dump` switch.

At this moment I2C is a part of `liblitedram` (`liblitedram/i2c.*`) and all the commands are in `LITEDRAM_CMDS`, but maybe we could move I2C to general BIOS code? 

As a side note, when doing first tests of this implementation I was using only `litex_sim` with a Verilator module that imitated an SPD EEPROM chip (and probably would work with some other EEPROMs). I doubt it would be useful to have it in Litex, but if you think it could be used for something then I can include it in a separate PR.


enjoy-digital:
Thanks @jedrzejboczar: having a sequence of simple commands to execute is indeed probably better than having a complex one that we will have to customize for the different case, so this is perfecly fine. We could provide examples of commands in the LiteX Wiki. The output format is also fine.

Can you just move the I2C functions outside of `liblitedram` as you are suggesting and we could merge it.

The SPD EEPROM emulation could also be useful (for SPD or EEPROM emulation), so if it's not taking you too much time, please create a PR for it yes.

jedrzejboczar:
Ok I'll create a PR with the changes, but will need to clean it up a little bit. 
Should I2C go to `libbase` the way UART does, or rather to `bios`?

enjoy-digital:
The software for the simple cores of `litex.cores` is in `libbase`, so you can also put I2C software there.

jedrzejboczar:
@enjoy-digital I2C is now in `libbase` and I2C commands are in `I2C_CMDS` group.

enjoy-digital:
@jedrzejboczar: Thanks for the changes, merged.

PR #543, opened 2020-05-28
litex/build/sim: add module for simulating SPD EEPROM

jedrzejboczar:
This adds a Verilator simulation module for SPD EEPROM as mentioned in https://github.com/enjoy-digital/litex/pull/541. It would probably be possible to use it to simulate some other EEPROMs with single-byte addressing. To implement I2C simulation I had to add `I2CMasterSim` module, which has separate `sda_in` and `sda_out` as I couldn't get Verilator to work properly with tristate pins.


enjoy-digital:
Thanks, this is clean and will be useful, merged.

Issue #544, opened 2020-05-29
Mor1kx support broken by #517 


enjoy-digital:
#517 changes the way accesses are done on the CSRs and broke Mor1kx support: lxsim is not responding to user commands.

Command reproduce it:
`rm -rf build && lxsim --cpu-type=mor1kx --opt-level=O0`

Works before https://github.com/enjoy-digital/litex/commit/4a5072a014c4e5c1bdee4a06fbaf1b2d34a3769d.

gsomlo:
It's caused by https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/include/hw/common.h#L43
cc @ozbenh -- MMPTR should always be `unsigned long`, i.e. `u32` on 32-bit CPUs (where CSR alignment is 32 bits) and `u64` on 64-bit CPUs (where CSR alignment is 64 bits).

`CSR_DATA_WIDTH` merely indicates how many of the bits ***within*** a MMPTR are ***populated***

I'll send a PR and link this issue shortly.

enjoy-digital:
Fixed with https://github.com/enjoy-digital/litex/pull/545.

enjoy-digital:
Fixed with https://github.com/enjoy-digital/litex/pull/545.

ozbenh:
I STRONGLY disagree. MMPTR should *not* be 64-bit if the bus width isn't 64-bit, otherwise it will cause all CSR accesses to be doubled up. Florent, this is the wrong fix.

The code was written as to take the alignment into account via CSR_OFFSET_BYTE and only do the proper access size accesses to the bus.

I don't know why Mor1kx broke and I can look into it but this is certainly not the right way to fix this.

ozbenh:
BTW. Is Mor1kx big endian ? That would explain if the CSRs follow native endian (which is also a very wrong thing to do btw, HW interfaces should have fixed well defined endianness, swapping is cheap).

mithro:
There are a couple of situations here;
 - 32bit CPU, 32bit bus and 8bit CSR width
 - 64bit CPU, 32bit bus and 8bit CSR width
 - 64bit CPU, 64bit bus and 8bit CSR width

mithro:
@ozbenh mor1kx is big endian.

mithro:
[...]

PR #545, opened 2020-05-29
csr: fix simple accessor alignment

gsomlo:
MMPTR should always follow CSR alignment, NOT CSR data width.
(the latter merely indicates how many bits within a MMPTR are
actually populated).

Fixup for commit #4a5072a. Also fixes issue #544


enjoy-digital:
Thanks!

ozbenh:
I disagree ! See my comment in #544

Issue #546, opened 2020-05-29
large, non-4-byte multiple CSRs vs "array" accessors


gsomlo:
Right now, large CSRs spread across multiple "simple CSRs" whose size is ***NOT*** a multiple of 4 bytes are distributed in a counter-intuitive way. For example, a 17 byte CSR whose hardware value is:

```
foo = 0x0102030405060708091011121314151617
```

will be represented starting with address `addr`, on designs with `csr_data_width == 8` like this:

```
addr   addr+align   ...
 |     |
 v     v
 (01), (02), (03), ..., (17)
```

(so far, so good, nothing weird happening yet).

However, with `csr_data_width == 32` we have:

```
addr   addr+align   ...
 |     |
 v     v
 (01), (02030405), (06070809), (10111213), (14151617)
```

**NOTE**: this is endianness-agnostic, as a number like `(0a0b0c0d)` represents the value stored at an appropriately aligned memory location in cpu-native endianness.

This means that the "array" accessors in `litex/soc/software/include/hw/common.h` are wrong *unless* the CSR size is a multiple of 4 bytes. They assume (currently wrongly so) that the CSR would be stored like this (in what some might assume to be a more "natural" or "intuitive" way):

```
addr         addr+align   ...
 |           |
 v           v
 (01020304), (05060708), (09101112), (13141516), (17)
```

The question is, of course, whether to fix the way CSRs are allocated for `csr_data_width > 8` (basically, implement the second, "natural/intuitive" way of distribution), or rather fix the array accessors?

I'm pretty strongly biased in favor of fixing CSR allocation and making the array accessors (retroactively) correct. Major advantage to that would be that we could keep things like this:

https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/liblitesdcard/sdcard.c#L231

from proliferating outside of `litex/soc/software/include/hw/common.h`, and prevent loss of endianness/alignment/csr_data_width agnostic access to hardware registers (currently sdcard only works on 32-bit CPUs, due to the above-referenced code fragment.

@enjoy-digital @mateusz-holenko @ozbenh -- any thoughts?

enjoy-digital:
@gsomlo: thanks for the analysis, fixing the gateware seems indeed more logical/intuitive.

enjoy-digital:
@gsomlo: can you explain your analysis a bit?
On a large register, the LSB is the CSR simple with the higher index, so the gateware should be implementing this: 
```
addr         addr+align   ...
 |           |
[...]

PR #547, opened 2020-05-30
soc/software/litesdcard: update for response register back to 128 bits

gsomlo:
The additional (17th) byte returned via the response register was ignored by software (bios and kernel), so LiteSDCard was updated to only return the (original, useful) 128 bits.

This patch updates the LiteSDCard code in the LiteX bios to only expect those 128 bits, and to do so in a manner that's portable across CSR data widths and alignments.

NOTES:
1. Depends on https://github.com/enjoy-digital/litesdcard/pull/13
2. Tested on nexys4ddr with `(mor1kx, vexriscv, rocket) x (csr_data_with \in {8, 32})`
3. Side-steps issue #546 (which I'd still like us to agree upon one way or another, but is no longer "urgent" :))

### DIFF
--- a/litex/soc/software/liblitesdcard/sdcard.c
+++ b/litex/soc/software/liblitesdcard/sdcard.c
@@ -33,7 +33,7 @@
 #define BLOCK_SIZE 512
 #define NO_RESPONSE 0xFF
 
-#define SDCARD_RESPONSE_SIZE 5
+#define SDCARD_RESPONSE_SIZE 4
 unsigned int sdcard_response[SDCARD_RESPONSE_SIZE];
 
 volatile char *SDREAD = (char*)(SDREAD_BASE);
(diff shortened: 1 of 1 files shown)


gsomlo:
@mateusz-holenko  @kamilrakoczy please also take a look and let us know what you think!

gsomlo:
@enjoy-digital I just want to drop this here so I don't forget. Regardless of whether we can stuff the sdcard response register back into 4 (instead of 4.25) bytes (and I'm fairly confident we can, based on successful tests), we should define a packed struct (or union overlaid on top of `u32_t sdcard_response[4]`) with named fields, to avoid doing stuff like e.g. this:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/liblitesdcard/sdcard.c#L549
in the long term. I can volunteer to clean that up, but want to see this PR resolved first :)

enjoy-digital:
Thanks, i'm merging and will have a closer look at the others points. @gsomlo: once things are clear in the gateware side, happy to have your help to clean that up on the software side.

mateusz-holenko:
@gsomlo This simplifies code a lot and looks good.

PR #551, opened 2020-06-01
Add Symbiflow toolchain support for Xilinx 7-series

mglb:
Toolchain support + Arty/Symbiflow platform + minimal Arty/Symbiflow target.

In order to build bitstream, you'll need the toolchain:
```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
bash miniconda.sh -b -p $PWD/miniconda
source "$PWD/miniconda/etc/profile.d/conda.sh"
hash -r
conda config --set always_yes yes --set changeps1 no
conda config --add channels conda-forge
conda config --add channels symbiflow
conda update -q conda
conda activate
conda install -c symbiflow yosys
conda install -c symbiflow yosys-plugins
conda install -c symbiflow vtr=8.0.0.rc2_3575_g253f75b6d
conda install lxml simplejson intervaltree python-constraint git pip
pip install git+https://github.com/symbiflow/fasm
wget "https://storage.googleapis.com/symbiflow-arch-defs/artifacts/prod/foss-fpga-tools/symbiflow-arch-defs/continuous/install/4/20200416-002215/symbiflow-arch-defs-install-a321d9d9.tar.xz"
tar -xf symbiflow-arch-defs-install-a321d9d9.tar.xz
export PATH=$PWD/install/bin:$PATH
```

### DIFF
--- a/litex/boards/platforms/arty_symbiflow.py
+++ b/litex/boards/platforms/arty_symbiflow.py
@@ -1,0 +1,35 @@
+# This file is Copyright (c) 2015-2019 Florent Kermarrec <florent@enjoy-digital.fr>
+# This file is Copyright (c) 2020 Antmicro <www.antmicro.com>
+# License: BSD
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import XilinxPlatform
+from litex.build.openocd import OpenOCD
+from litex.boards.platforms.arty import _io, _connectors
+
+# Platform -----------------------------------------------------------------------------------------
+
+class Platform(XilinxPlatform):
+    default_clk_name   = "clk100"
+    default_clk_period = 1e9/100e6
+
+    def __init__(self, variant="a7-35"):
+        device = {
+            "a7-35":  {"part": "xc7a35tcsg324-1", "symbiflow-device": "xc7a50t_test"},
+        }[variant]
+        XilinxPlatform.__init__(self, device["part"], _io, _connectors, toolchain="symbiflow")
+        self.toolchain.symbiflow_device = device["symbiflow-device"]
+
+    def create_programmer(self):
+        bscan_spi = "bscan_spi_xc7a100t.bit" if "xc7a100t" in self.device else "bscan_spi_xc7a35t.bit"
+        return OpenOCD("openocd_xc7_ft2232.cfg", bscan_spi)
+
+    def do_finalize(self, fragment):
+        # Prevent GenericPlatform from creating period constraint on input clock
+        pass
+
+    def add_period_constraint(self, clk, period, phase=0):
+        if clk is None: return
+        if hasattr(clk, "p"):
+            clk = clk.p
+        self.toolchain.add_period_constraint(self, clk, period, phase)
--- a/litex/boards/targets/arty_symbiflow.py
+++ b/litex/boards/targets/arty_symbiflow.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+
+# This file is Copyright (c) 2015-2019 Florent Kermarrec <florent@enjoy-digital.fr>
+# This file is Copyright (c) 2020 Antmicro <www.antmicro.com>
+# License: BSD
+
+import os
+import argparse
+
+from migen import *
+
+from litex.boards.platforms import arty_symbiflow
+from litex.build.xilinx.symbiflow import symbiflow_build_args, symbiflow_build_argdict
+
+from litex.soc.cores.clock import *
+from litex.soc.integration.soc_core import *
+from litex.soc.integration.soc_sdram import *
+from litex.soc.integration.builder import *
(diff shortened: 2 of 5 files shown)


mithro:
@mglb Can we use a conda environment.yml file to describe the toolchain configuration which works? (And it should work in the future for Mac / Windows too right)

mithro:
@mglb Why do we need a specific arty_symbiflow platform?

litex/boards/platforms/arty_symbiflow.py
```diff
@@ -0,0 +1,35 @@
...
+# License: BSD
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import XilinxPlatform
+from litex.build.openocd import OpenOCD
+from litex.boards.platforms.arty import _io, _connectors
+
+# Platform -----------------------------------------------------------------------------------------
+
+class Platform(XilinxPlatform):
```
mithro, line 12: Why do we need this platform?

litex/boards/targets/arty_symbiflow.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
mithro, line 1: Why do we need this target?

litex/build/xilinx/symbiflow.py
```diff
@@ -0,0 +1,321 @@
...
+        v_file = build_name + ".v"
+        v_output.write(v_file)
+        platform.add_source(v_file)
+
+        self._generate_makefile(
+            platform   = platform,
+            build_name = build_name
+        )
+
+        # Generate design constraints
```
mithro, line 283: Why do we need all of an `.xdc`, `.sdc` and `.pcf`?

enjoy-digital:
Thanks, i understand it will still be a bit experimental for now but we can probably reduce some duplications. I'm merging and will try to remove them.

litex/boards/targets/arty_symbiflow.py
mglb: Just for testing purposes. Original Arty target doesn't work, because Symbiflow doesn't support all Xilinx primitives (e.g. IOBUFDS) yet. It also requires buffers on input clock.

litex/build/xilinx/symbiflow.py
mglb: .pcf (list of pin locations) is required by `place` script.
.sdc (timing constraints) is optional. I'll change this so that this file will be generated and passed to the toolchain only when there are constraints set.
.xdc (IO settings) is optional. It is never empty, so it is easier to just generate and use it, even when this is not strictly needed. Is there any advantage in not using it?

mglb:
> @mglb Can we use a conda environment.yml file to describe the toolchain configuration which works? (And it should work in the future for Mac / Windows too right)

For all other toolchains (including open source ones), Litex assumes that all required tools are already installed.
This environment is generic (the script comes from https://github.com/SymbiFlow/symbiflow-examples, where Litex is not used). I think it would be better to do this as another task in the Symbiflow project instead of making it part of Litex.

> @mglb Why do we need a specific arty_symbiflow platform?

* Part "xc7a35ticsg324-1L" is not present in prjxray-db.
* "symbiflow-device" have to be specified
* Original Platform adds Vivado-specific (or at least unsupported by Symbiflow) commands
* Period constraint on input clock (clk100), set by GenericPlatform.do_finalize(), is not supported by VPR:
  ```
  Error 1:
  Type: SDC file
  File: arty_symbiflow.sdc
  Line: 3
  Message: Clock name or pattern 'clk100' does not correspond to any nets. To create a virtual clock, use the '-name' option.
  ```
  Input clock has to go through IBUF and BUFG in order to be used. BUFG's output clock signal can then be optionally constrained.

The alternative is to merge it with original Arty platform by using `if toolchain="symbiflow": ... else: ...` for everything in the class.

mglb:
@enjoy-digital thanks for merge/cleanup. Just one more thing to fix - Symbiflow needs "xc7a35tcsg324-1" part (instead of "xc7a35ticsg324-1L"):
`realpath: /src/symbiflow/install/bin/../share/prjxray/prjxray-db/artix7/xc7a35ticsg324-1L/part.json: No such file or directory`

enjoy-digital:
@mglb: thanks for the answers, i indeed merged your work and tried to avoid some duplications. To continue the discussion on the remaining points, i created https://github.com/enjoy-digital/litex/issues/554. For now we are indeed using` if toolchain=="symbiflow"` switches, but that would be good if we could avoid them as much as possible as the Symbiflow support improves. For the review i just generated the files from your work and then verified i was able to regenerate the same ones with the changes, so haven't really used to the toolchain. Would you mind creating a PR for the part change? (if possible it should be located in `symbiflow.py` since this seems to be a limitation of the toolchain).

PR #553, opened 2020-06-01
sdram: Unconditionally switch to SW control before inits

ozbenh:
This will allow the controller to default to HW control which means
the sim model can be used without specific initializations

Signed-off-by: Benjamin Herrenschmidt <benh@kernel.crashing.org>


enjoy-digital:
Thanks, merged.

enjoy-digital:
That's working, but i reverted the `sdrsw()` in sdrlevel with https://github.com/enjoy-digital/litex/commit/55723f138beeb0487fd3c118a8796fd80712f01f to avoid breaking the sdrlevel command in the BIOS.

PR #555, opened 2020-06-02
Symbiflow: do not use empty .sdc file

mglb:
Fix mentioned in:
https://github.com/enjoy-digital/litex/pull/551#discussion_r433823513


enjoy-digital:
@mglb: is the toolchain complaining with an empty `.sdc`? If not, we should probably just always generate it and pass if to the toolchain.

enjoy-digital:
Closing, we can just let LIteX always generate the empty `.sdc` and pass it to the toolchain.

PR #556, opened 2020-06-02
Symbiflow toolchain support: fix part name and clock frequency

mglb:
Response for:
https://github.com/enjoy-digital/litex/pull/551#issuecomment-637507029

* Remap part name
* Set sys_clk_freq to 60MHz


enjoy-digital:
@mglb: thanks. As discussed in https://github.com/enjoy-digital/litex/pull/555, i don't think we should make optional the `.sdc` if Symbiflow is not complaining with an empty file. If so, can you remove 0616b7d from this PR?

mglb:
Oops, sorry. Commit removed

enjoy-digital:
Thanks for the update, merged.

PR #557, opened 2020-06-02
bios: boot: Boot linux on mor1kx with external device tree and rootfs

mateusz-holenko:
This allows to boot linux on mor1kx with an external device tree file and (optional) rootfs using netboot.
If the kernel file or the device tree are not found, it falls back to the `boot.bin` file.


enjoy-digital:
Thanks, merged.

PR #559, opened 2020-06-06
liblitesdcard: convert all sources to unix style newlines (cosmetic)

gsomlo:
bring fat fs sources in sync with the convention used with every other source file in the repo.

this prevents us dealing with lots and lots of `^M`s at line ends when doing e.g. `git diff` :)


enjoy-digital:
Thanks.

PR #564, opened 2020-06-10
Microwatt updates

shenki:
This fixes building a microwatt litex soc with the current version of microwatt.

```
./litex/boards/targets/arty.py --build --cpu-type microwatt --with-ethernet --integrated-rom-size $((64 * 1024))
```


ozbenh:
Looks good to me

PR #567, opened 2020-06-15
bios/linker: Place .data in sram with initial copy in rom.

### LINKED ISSUE #566: bios: `.data` section placed in rom, wrongly ends up acting like `.rodata`
The `.data` section of the bios (initialized global variables) is *wrongly* placed in `rom` by the [linker.ld](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/linker.ld#L55) script.

As a consequence, initialized global variables such as [this one](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/liblitesdcard/spisdcard.c#L271) end up being impossible to modify, leading to buggy bios behavior (`spisdcardstatus` will stubbornly remain set to `1` on Rocket, no matter how many times one tries to write a `0` to that location!!!).

I tried `s/rom/sram/` in [linker.ld](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/bios/linker.ld#L55), but that's obviously not a good solution, since it results in a hang after the `LiteX` logo is printed at startup :) Need some help figuring out what the proper fix should look like.

Thanks!

zyp:
As per #566, `.data` section is currently located in rom, rendering it read-only. This PR moves `.data` to sram, leaving initial state in rom that needs to be copied to sram during startup. This means that every `crt0.S` needs to be updated to fix this.

* `vexriscv`: Updated and tested.
* `rocket`: Updated and tested.
* `serv`: Updated and tested.
* `blackparrot`: Identical to `rocket`, not tested.
* `cv32e40p`: Identical to `vexriscv`, not tested.
* `minerva`: Identical to `serv`, not tested.
* `picorv32`: Existing copy code updated, not tested.
* `lm32`: Existing copy code updated, not tested.
* `mor1kx`: Still needs to be updated.
* `microwatt`: Still needs to be updated. Seems to also miss `.bss` initialization.

Fixes #566.

### DIFF
--- a/litex/soc/cores/cpu/blackparrot/crt0.S
+++ b/litex/soc/cores/cpu/blackparrot/crt0.S
@@ -57,6 +57,19 @@ crt_init:
   la a0, trap_entry
   csrw mtvec, a0
 
+data_init:
+  la a0, _fdata
+  la a1, _edata
+  la a2, _fdata_rom
+data_loop:
+  beq a0,a1,data_done
+  ld a3,0(a2)
+  sd a3,0(a0)
+  add a0,a0,8
+  add a2,a2,8
+  j data_loop
+data_done:
+
 bss_init:
   la a0, _fbss
   la a1, _ebss
--- a/litex/soc/cores/cpu/cv32e40p/crt0.S
+++ b/litex/soc/cores/cpu/cv32e40p/crt0.S
@@ -94,6 +94,19 @@ crt_init:
   la a0, vector_table
   csrw mtvec, a0
 
+data_init:
+  la a0, _fdata
+  la a1, _edata
+  la a2, _fdata_rom
+data_loop:
+  beq a0,a1,data_done
+  lw a3,0(a2)
+  sw a3,0(a0)
+  add a0,a0,4
+  add a2,a2,4
+  j data_loop
+data_done:
+
 bss_init:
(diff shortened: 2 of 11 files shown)


gsomlo:
foo

I get the above when building with

```
litex/litex/boards/targets/nexys4ddr.py --csr-data-width 32 --sys-clk-freq 75e6 \
    --with-ethernet --with-spi-sdcard --cpu-type rocket --cpu-variant linux \
    --integrated-rom-size=0x10000 --build
```
It eventually boots from either ethernet or sdcard, but the part where memtest results are printed out to the terminal is all scrambled (on the nexs4ddr).

gsomlo:
works if sram is doubled to accomodate both .data and stack (current 0x1000 size is not enough for both). Suggest `s/0x1000/0x2000/` [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc_core.py#L78) and [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc_core.py#L267).

Tested on rocket (64bit) and nexys4ddr, works fine otherwise.

enjoy-digital:
Thanks @zyp, @gsomlo. Let's keep #566 open to finish the work and do the changes on the mor1kx, microwatt.

PR #568, opened 2020-06-15
build/sim/core/modules: fix compilation warnings

sergachev:
I had a problem with these modules not compiling because my default CC is clang 10 and it did not let these uninitialized return values and an empty if clause. Therefore these edits.


enjoy-digital:
Thanks, merged.

PR #574, opened 2020-06-21
Initial Edalize support

mglb:
This is a PoC of using Edalize (https://github.com/olofk/edalize) as Litex toolchain library. It is implemented for Xilinx platform/Vivado toolchain as an option. Necessary command line argument (`--use-edalize`) is added to Arty target.

In order to add Edalize, I've done some refactoring in Xilinx platform and Xilinx toolchains to remove shared code (adding constraints, creating/changing build directories, etc) from toolchain classes. Similar refactoring can be done without bigger problems in other platforms/toolchains (except Sim/Verilator, where it would need more work). Common code is currently placed in XilinxPlatform, but I think it would be more suitable to make a class like `GenericToolchain` and inherit it in all toolchains.

Edalize needs following changes to be able to fully replace Litex toolchain wrappers:
* Toolchain-independent API for setting period and false path constraints and automatic generation of constraints file(s). I've placed example implementation for Vivado in litex/build/edalize_ext/vivado.py.
* A way for passing custom toolchain-specific platform commands
* Fixes/updates for some toolchains (e.g. ISE, which doesn't work as good as Litex ISE backend)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
 install:
   # Get Migen / LiteX / Cores
   - cd ~/
-  - pip3 install requests
+  - pip3 install requests edalize
   - cp $TRAVIS_BUILD_DIR/litex_setup.py .
   - python3 litex_setup.py init install
   # Install the LiteX version being tested
--- a/litex/boards/platforms/arty.py
+++ b/litex/boards/platforms/arty.py
@@ -4,6 +4,8 @@
 
 from litex.build.generic_platform import *
 from litex.build.xilinx import XilinxPlatform
+from litex.build.xilinx.platform import xilinx_platform_args, xilinx_platform_argdict, xilinx_platform_build_argdict
+from litex.build.xilinx.vivado import vivado_build_args, vivado_build_argdict
 from litex.build.openocd import OpenOCD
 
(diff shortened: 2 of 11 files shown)


enjoy-digital:
Thanks @mglb, i'll have a closer look and study it. As discussed together with @kgugala, i'd like to think about it since toolchain support is tighly coupled with LiteX, so i'll think about it to see if it's the good direction to go or if LiteX/Edalize interaction should be done differently.

enjoy-digital:
Sorry but i don't think supporting Edalize natively in LiteX is a good idea: This could seem nice for simple designs where we don't need that much interaction with the toolchain (for example a LiteX target) but as soon as we'll add more complex timings constraints, import IPs, etc... it will break. Not because of Edalize, but just because we'll need to add specific code to pass the extra information to Edalize.

We also already have good support in LiteX for the toolchain we are really interested in, so native Edalize support would not add that much additional value for us. Edalize could however be interesting to experiment with the toolchains that are not already supported but it would probably be better to just add an export tool that would generate all the gateware build information to a `json` file (similar to whas is already done for sofware) that an eventual LiteX2Edalize tool could reuse. This would be less invasive and would also allow this information to be reused for other purposes.

Issue #341, opened 2020-01-13
Add support for memtest on non-DDR memories.


mithro:
It would be good to have support for memtest functionality on non-DDR memories including;
 - [ ] HyperRAM
 - [ ] SPI RAM (like used on the hackaday badge)
 - [ ] Static-RAM
 - [ ] Single Port RAM inside the Fomu (128 kbytes).

mithro:
The memory bandwidth testing would be great too!

enjoy-digital:
With the recent changes, generic mem_test and mem_speed commands were already available. With https://github.com/enjoy-digital/litex/commit/1a338b602a125ba07ed348b7d63db49cb73f0bd8, a mem_list command is added to list the available memory regions. This way all regions of the SoC can be easily tested (write/read accesses and speed):

List the memory regions:
```
    litex> mem_list
    Available memory regions:
    ROM       0x00000000 0x8000
    SRAM      0x01000000 0x2000
    MAIN_RAM  0x40000000 0x10000000
    CSR       0x82000000 0x10000
```

 Test 0x1000 bytes of MAIN_RAM:
   
```
    litex> mem_test 0x40000000 0x1000
    Memtest at 0x40000000 (4KiB)...
      Write: 0x40000000-0x40001000 4KiB
       Read: 0x40000000-0x40001000 4KiB
    Memtest OK
```
    
Test speed on 0x1000 bytes of MAIN_RAM:
    
```
    litex> mem_speed 0x40000000 0x1000
    Memspeed at 0x40000000 (4KiB)...
      Write speed: 352KiB/s
       Read speed: 288KiB/s
```

PR #575, opened 2020-06-24
Move memtest from liblitedram to libbase

jedrzejboczar:
Closes https://github.com/enjoy-digital/litex/issues/341.

This adds two commands: `memtest <addr> [<maxsize>]` and `memspeed <addr> <size> [<readonly>]` that are added with libbase and can be used to test any memory.

I didn't include it, but we can probably remove the "hack" https://github.com/enjoy-digital/litex/compare/master...antmicro:jboc/memtest?expand=1#diff-f29671cd8c5038ad1a5446dde0c2d4fbR27-R32, as it is probably not needed anymore?


enjoy-digital:
Thanks! Merged. We could indeed probably remove the hack, i'm going to look at this. I'll keep #341 open for now since we'll also probably want others memories to be tested at startup.

PR #576, opened 2020-06-25
Deprecate slave terminology

bunnie:
Master/slave deprecated in favor of controller/peripheral, per OSHWA recommendation.

http://oshwa.org/a-resolution-to-redefine-spi-signal-names

(post seems to be WIP, link should go live in the next couple days or so)

### DIFF
--- a/litex/soc/cores/i2s.py
+++ b/litex/soc/cores/i2s.py
@@ -16,13 +16,17 @@ class I2S_FORMAT(Enum):
     I2S_LEFT_JUSTIFIED = 2
 
 class S7I2S(Module, AutoCSR, AutoDoc):
-    def __init__(self, pads, fifo_depth=256, master=False, concatenate_channels=True, sample_width=16, frame_format=I2S_FORMAT.I2S_LEFT_JUSTIFIED, lrck_ref_freq=100e6, lrck_freq=44100, bits_per_channel=28):
+    def __init__(self, pads, fifo_depth=256, controller=False, master=False, concatenate_channels=True, sample_width=16, frame_format=I2S_FORMAT.I2S_LEFT_JUSTIFIED, lrck_ref_freq=100e6, lrck_freq=44100, bits_per_channel=28):
+        if master == True:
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks merged. LiteX sill uses Master/Slave terminology in different locations and we could think in the the future to also replace this but this will have to be thought carefully to avoid causing regressions in projects created with it.

bunnie:
Yah, I took a brief look and in particular the wishbone stuff uses a lot of master/slave terminology. My thought is to work at the problem as I maintain various blocks; I'll send over more PRs when the opportunity presents itself. 

Backward-compatibility will be important, I agree -- especially where the terminology is baked into core APIs. The good news is Python has some pretty decent tools for refactoring APIs!

enjoy-digital:
The issue when deciding to remove the master/slave terminology use in a project and keeping backward-compatibility is incompatible with the initial goal since you'll still have to use this terminology in the part of the code handling the backward-compatibility. This would still be different since the use would not be encouraged, but part of the initial issue is still there.
 
So if we go in this direction for the others parts of the project, i don't think there could be half measures and the terminology should completely be removed (so no backward-compatibility) and i'm not sure i would use controller/peripheral but more probably main/secondary (since this avoid breaking MOSI/MISO for SPI, DIR_M_TO_S/DIR_S_TO_M on streams, etc...). So i will think about it and look at what is currently used in other projects.

bunnie:
Sure thing. I'm just going with the OSHWA "official" recommendations (which ended up being to prefer to scrap the whole *I*O terminology and go with simply "data out" (SDO) and "data in" (SDI) but also allowing controller/peripheral if you want to stick with a *I*O system). I like *I*O so I went with CIPO/COPI, but it's a compromise recommendation made by a big group of very opinionated and mostly white people with varying agendas that was litigated over a matter of weeks. In other words, your basic standards committee.

That being said, I totally get the desire to stay with M*S* terminology. Suggestions on this front include "manager/subordinate" or "main/secondary". Adafruit uses "microcontroller/sensor" but that is pretty specific for IoT applications. Just let me know which direction you want to bring the Litex community toward, I'll support it!

PR #578, opened 2020-06-29
Blackparrot litex syn with master BP, upgrade to IMA

scanakci:
Some edits in core.py and in core/blackparrot to upgrade current BlackParrot to IMA architecture supported by most recent BP repo.


enjoy-digital:
Thanks, merged.

PR #580, opened 2020-06-30
Improve verilator compilation speed

Dolu1990:
by asking verilator to split the C++ model into multiple files.

(copy pasted from the SpinalHDL verilator backend : https://github.com/SpinalHDL/SpinalHDL/blob/c9bd9dd06debf1466d3d45a3d9e12b22ba87e8d6/sim/src/main/scala/spinal/sim/VerilatorBackend.scala#L401)


enjoy-digital:
Thanks! Merged.

enjoy-digital:
Just tested and can confirm it's indeed much faster with O3, thanks again!

PR #581, opened 2020-07-01
i2s: Fix the incorrect TX fifo almost empty offset

mateusz-holenko:


bunnie:
good catch. @enjoy-digital this is reviewed and looks correct.

enjoy-digital:
Thanks @mateusz-holenko, @bunnie, merged.

PR #582, opened 2020-07-05
Minor sdcard fixes

gsomlo:
some indentation fixes to make things more legible, and an error code that should be moved *outside* of an `#ifdef DEBUG`.


enjoy-digital:
Thanks!

PR #583, opened 2020-07-06
liblitesdcard/sdcard: adjust card-ready timeout

gsomlo:
Testing on nexys4ddr and rocket, approximately 12 iterations of the
timeout loop are needed to receive a "ready" response from the SDcard,
assuming a "warm" reset where the card has already been previously
initialized.

If the SDcard is ejected and re-inserted, or if the board is "cold-reset"
(e.g., reprogrammed via openocd vs. a simple push of the reset button),
it takes approximately 450 iterations before the SDCard responds with a
"ready" message.

In either case, a timeout of 10 is insufficient, so let's bump it to 1024.

This patch also adds some minor cosmetic improvements.

### DIFF
--- a/litex/soc/software/liblitesdcard/sdcard.c
+++ b/litex/soc/software/liblitesdcard/sdcard.c
@@ -448,7 +448,7 @@ void sdcard_decode_csd(void) {
 
 int sdcard_init(void) {
 	unsigned short rca;
-    uint16_t timeout;
+	uint16_t timeout;
 
 	/* initialize freq */
 	sdcard_set_clk_freq(16000000);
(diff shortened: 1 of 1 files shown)


gsomlo:
Using `busy_wait(10)` (instead of `1`) inside the timeout loop, we need 3 (warm) or 48 (cold) iterations to get a `ready` from the SDcard (on nexys4ddr, with rocket). Force-pushed with `busy_wait(10)` and `timeout=128`.

enjoy-digital:
Thanks @gsomlo for the tests and feedback.

PR #587, opened 2020-07-10
mor1kx: Do not generate the ror instruction

mateusz-holenko:
The mor1kx core does not support `l.ror` instruction
by default, but gcc/clang flags allowed the
compiler to generate it.


enjoy-digital:
Thanks, merged.

PR #588, opened 2020-07-11
trellis: Add option to select SPI mode.

oskirby:
This allows passing the ``--spimode`` option to ecppack, and gives a significant speedup when booting large bitstreams on ECP5 boards that support dual or quad SPI operation. This was tested using a Logicbone with an 85F part and it reduced the boot time to get to the Litex bios from 12s to about 3s.


enjoy-digital:
Thanks! That's merged, i just fixed a typo withhttps://github.com/enjoy-digital/litex/commit/ae3c78f6d16dcbb6250ea61acaddd1a4c1f876e1.

PR #592, opened 2020-07-15
symbiflow: changed toolchain command names in Makefile

acomodi:

This PR changes the symbiflow toolchain commands to have the `symbiflow` prefix.

This is related to a change in the toolchain wrappers name, so to avoid conflicts between the symbiflow related binaries and the system-related commands (e.g. `route` causes conflicts with the Unix homonym command).


enjoy-digital:
Thanks, merged.

Issue #590, opened 2020-07-13
Add AXI Lite components/interconnect equivalent to the existing Wishbone ones


enjoy-digital:
The main bus of the SoC is for now limited to Wishbone which is simple and easy but starts being a bottleneck on several systems: 
- Most of the newly added CPU (Zynq, BlackParrot, Rocket, etc...) use native AXI interfaces and the bridges to Wishbone are limiting efficiency/performance.
- Most of the cores are now supporting AXI which has become the de facto standard for performance and interoperability (Ours like LiteDRAM, LitePCIe, etc... and external cores),

Supporting an AXI-Lite bus as the main bus of the SoC would improve the overall performance and interoperability (as described in more details in https://github.com/enjoy-digital/litex/issues/372) and would be a good first step before supporting AXI-Full.

With the current AXI definition and modules in https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/axi.py we are already able make the SoC interoperate with internal/external AXI components while still using Wishbone as the main bus.

We should progressively add the AXI-Lite components/interconnect equivalent to the existing Wishbone ones:
- [ ] AXILiteSRAM.
- [ ] AXILite2CSR.
- [ ] AXILiteConverter.
- [ ] AXILiteInterconnectPointToPoint.
- [ ] AXILiteArbiter.
- [ ] AXILiteDecoder.
- [ ] AXILiteInterconnectShared.
- [ ] AXILiteCrossbar.

In a first time, the components that could be easily tested on real SoCs while still using a Wishbone bus as the main bus should be added: AXILiteSRAM, AXILite2CSR, AXILiteConverter. To allow them to be used in a SoC,  the `add_adapter` function of the `SoCBusHandler` would need to be adapted to enable bridges between AXILite <--> Wishbone using the existing bridging components.

Once these components have been validated, the interconnect should be created and tested with unit-tests, then integrated in the SoC Builder and tested on a simple SoC: CPU + ROM + SRAM + CSRBridge + UART and then on real-world designs where the only change from the user would be to specify the `bus_standard` to "axi-lite".

xobs:
How difficult would it be to add bridge equivalents to all of these? I'd like for the existing `wishbone-bridge` and `wishbone-tool` stuff to continue to work, and it would be nice if the protocol would be similar.

enjoy-digital:
[...]

PR #593, opened 2020-07-15
Add AXILite components: AXILiteSRAM and AXILite2CSR

jedrzejboczar:
This is a part of https://github.com/enjoy-digital/litex/issues/590.

This PR adds a bridge from AXILite to CSR bus and an SRAM with AXILite interface.
Methods for adding RAM (`add_ram` and `add_rom`) accept additional `bus` parameter, and will use either `wishbone.SRAM` or `axi.AXILiteSRAM` depending on the bus type, defaulting to wishbone. The `add_adapter` method will perform conversion from AXILite to wishbone and do the data width conversion. For now `AXILiteConverter` is a dummy implementation that can connect only interfaces with the same width, and I am working on adding a proper converter.

I added tests for `AXILite2CSR` and `AXILiteSRAM` and I was also able to successfully test them in `litex_sim`.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -743,19 +743,28 @@ class SoC(Module):
         setattr(self.submodules, name, SoCController(**kwargs))
         self.csr.add(name, use_loc_if_exists=True)
 
-    def add_ram(self, name, origin, size, contents=[], mode="rw"):
-        ram_bus = wishbone.Interface(data_width=self.bus.data_width)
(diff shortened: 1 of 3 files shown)


jedrzejboczar:
I guess I added the `bus` argument mostly for test/development purpose, to be able to force using `AXILiteSRAM` while we still have Wishbone as the main bus. Eventually we would just use the SRAM corresponding to the main bus, so this argument actually makes no sense. I'll add a commit to remove the `bus` argument and will just add the logic for choosing between AXILite/Wishbone SRAM when we have the other components implemented.

enjoy-digital:
Thanks for the update, that's merged.

PR #594, opened 2020-07-16
Add AXILiteDownConverter

jedrzejboczar:
Part of https://github.com/enjoy-digital/litex/issues/590.

This implements the down-converter for AXILite along with unit tests. It could probably be simplified a bit, and some transitions could be optimized not to waste cycles (1 in `IDLE` and probably 1 in `*-RESPONSE-MASTER`).

The up-converter has not been implemented yet, but maybe I should move on to the interconnect, as we don't have wishbone up-converter too, and then we could add the AXILite up-converter if needed? 

The `axi.py ` file gets longer and with more components from https://github.com/enjoy-digital/litex/issues/590 it may grow quite large, so I could probably split it into `axi.py`/`axi_lite.py`? The same with tests.

### DIFF
--- a/litex/soc/interconnect/axi.py
+++ b/litex/soc/interconnect/axi.py
@@ -696,6 +696,7 @@ class AXILiteDownConverter(Module):
         last_was_read = Signal()
         aw_ready      = Signal()
         w_ready       = Signal()
+        resp          = Signal.like(master.b.resp)
 
         # Slave address counter
         master_align = log2_int(master.data_width//8)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, looks good. For now i have a preference to keep all AXI modules in a single file, but if the file becomes too large, we could split in `axi/axi_lite/axi_stream` at some point yes. We can merge your current PR and if you find way to improve things later you could create another PR (i will also do a closer review at some point and will also try to see if we can improve things).

You can move the interconnect yes, if things are clear on your side to create and up-converter (Wishbone & AXI) feel free to also look at that.

PR #598, opened 2020-07-20
interconnect/csr_bus: fix paged access warning

sergachev:
A recent edit broke it - making a minor fix.


enjoy-digital:
Thanks, it seems i was also doing some C this day :)

PR #600, opened 2020-07-22
Implement AXI Lite interconnect

jedrzejboczar:
This would close https://github.com/enjoy-digital/litex/issues/590.

This PR adds all the AXI Lite interconnect modules, similar to the wishbone ones: point-to-point interconnect, arbiter, decoder, shared interconnect, crossbar, timeout. It also exposes the `bus_standard` `SoC` argument in the standard SoC arguments.

The tests of AXI Lite converter and interconnect have been moved to `test/axi_lite.py`. The most complex interconnect tests cover a scenario with 3 masters and 3 slaves for 64 transactions per master, for both shared interconnect and a crossbar, with randomized transaction timings. There is also a test of the timeout module in shared interconnect.

The interconnect works separately for read and write channels, so we could add some more specific tests for this, but the stress tests seemed to work correctly and looked good when examining in gtkwave, however I am not entirely sure if there are no corner cases that could fail. I will also rewrite the down converter to separate write and read channels, as the current implementation doesn't yet convert in parallel, although is easy to separate.

I was also able to successfully run `litex_sim` with `--bus-standard axi-lite` (cpu, csr, main ram busses converted from Wishbone to AXI Lite) and run correct memtests, and also test timeout detection by writing to not-mapped regions. But I haven't yet done any tests on hardware.


enjoy-digital:
Thanks for the great work @jedrzejboczar! I did a first review and also got it working with `litex_sim`  and the simulation in Linux-on-LiteX-VexRiscv. The current tests are also enough for now and we could add more specific/complex tests in the future. So we can merge this initial version, thanks.

PR #602, opened 2020-07-24
doc: socdoc: document `sphinx_extensions` parameter

xobs:
This adds documentation for `sphinx_extensions` which can be used to add
additional features to output.


enjoy-digital:
Thanks, merged.

xobs:
I have one more patch on top of this to append extra config. I'll merge that additional patch once it passes build.

PR #605, opened 2020-07-26
Merge sequential reads for the UART litex_server backend

cklarhorst:
The UART backend writes [read identifier, num_reads, addr] for every read request.
Etherbone packets are able to include multiple read requests.
Therefore, it is beneficial to merge sequential read requests to reduce writes (and possible latency overhead).

Benchmark:
A typical litescope fetch script with the following signals [ddrphy.dfi,cpu.ibus.cyc,cpu.ibus.stb] results in 1 read for the
data_valid register and 24 sequential reads for the scope data per timestamp.
Fetching data for a capture length of 512 over a 921600 baud UART (arty board) took:
```
205s (current master branch)
 18s (with this merge function)
```

The proposed merger only merges read requests from one etherbone packet at a time and doesn't change the read order.
I didn't observe a speedup for UART over litex_server and gdb over the wishbone-script over linux_server. I think they only do single read requests. Although it would be pretty cool if gdb would issue larger reads for example when examining memory content.
I could only test the UART communication therefore the merger is only active for this backend.

### DIFF
--- a/litex/tools/litex_server.py
+++ b/litex/tools/litex_server.py
@@ -15,6 +15,22 @@ import threading
 from litex.tools.remote.etherbone import EtherbonePacket, EtherboneRecord, EtherboneWrites
 from litex.tools.remote.etherbone import EtherboneIPC
 
+# Merges sequential reads:
+# input:  list of addresses
+# output: list of (start_addr, read_size)
+# example: [0x0, 0x4, 0x10, 0x14] -> [(0x0,2), (0x10,2)]
+def _read_merger(addrs):
+    addr_start = addrs[0]
+    num_reads  = 1
+    for addr in addrs[1:]:
+        if addr_start+4*num_reads != addr:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for looking at this @cklarhorst! Optimizing this was indeed planned, but with a very low priority, so never happened... Great if this improves Litescope uploads by 10X over UART, this will be very useful! Before applying it, i'll just do some tests and check on the Etherbone code, IIRC the maximum transactions per packet has been reduced to 4 to reduce resource usage, so possible we'll also have to limit this to 4 for UDP.

enjoy-digital:
@cklarhorst it's merged and indeed improving LiteScope upload speed over UART significantly, thanks! I just did a bit of refactoring with https://github.com/enjoy-digital/litex/commit/cc8440549fb3bb4a3885423162353fb14f4f0853. I will also do some tests with CommUDP soon.

PR #606, opened 2020-07-28
litex_setup.py add pythondata-cpu-vexriscv-smp

Dolu1990:


Dolu1990:
Seem i need to update some eggs first XD

PR #607, opened 2020-07-28
soc/cores/cpu/vexriscv_smp integration

Dolu1990:
This (as far i know) break nothing, as it has no interations with the single core vexriscv integration.


enjoy-digital:
Thanks! merged.

PR #611, opened 2020-07-30
soc/interconnect/axi: add AXILite -> AXI converter

jedrzejboczar:
The tests use `AXI2Wishbone` and `AXI2AXILite` and SRAMs to verify that `AXILite2AXI` works, which isn't the best, but was the most straightforward.


enjoy-digital:
Thanks! merged. (cc @gsomlo you should probably use this for the DMA port of Rocket).

PR #615, opened 2020-08-01
Add openFPGALoader programmer

pepijndevos:
Currently the ULX3S board uses the ujprog programmer, which is specific to this board, and does not have build system that works on my machine.

[OpenFPGALoader](https://github.com/trabucayre/openFPGALoader) is a well-written en maintained program that supports many FPGA boards, and this PR adds a simple class for it.

In a separate PR I'll update the ULX3S target to use this class.


enjoy-digital:
Thanks, merged.

PR #617, opened 2020-08-01
RFC: enable DMA with Rocket

gsomlo:
What the patch does:
  - create a wrapper to convert Wishbone2AXI
  - activate slave AXI port on Rocket, convert to wishbone, expose as `dma_bus`
  - fix `dma_bus` routing by covering entire 4Gbyte region from 0 to `0x1_0000_0000`
      - since rocket has a mmio uncached connection to the "main" wishbone system bus connecting to all slave peripherals, and
         a memory point-to-point link connecting it directly to the DRAM, we need to route any and all dma master accesses through
         the rocket chip so that devices can be reached at addresses below `0x8000_0000` (mmio w.r.t. rocket's memory map), and
         above `0x8000_0000` (RAM as far as rocket-based litex soc is concerned).
  - the `dma_bus` fix requires an extra bit to be added to the address width of the wishbone interconnect i.e. the `adr_width` of the
   entire WB `Interface()`.

The result is that sdcard boot successfully copies `boot.bin` to DRAM (to address `0x8000_0000`), but unfortunately still hangs at `Liftoff!`. I *did* a `mr 0x80000000 0x100` and compared it with a `hexdump -C boot.bin | head`, and found at least the first several hundred bytes matching. Not entirely sure about the origin of a potential error in the dma copy process to DRAM that would cause booting BBL and the 64-bit linux kernel to fail.

The command line to build the bitstream is:

```
litex/litex/boards/targets/nexys4ddr.py --csr-data-width 32 \
        --with-ethernet --with-sdcard \
        --cpu-type rocket --cpu-variant linux \
        --integrated-rom-size=0x10000 --build
```

***EDIT***: added a patch to slow down LiteSDCard clock to 25MHz -- as noted by @enjoy-digital, it appears Rocket's internal DMA routing can't handle DMA writes from LiteSDCard at the full 50MHz rate.

### DIFF
--- a/litex/soc/interconnect/wishbone.py
+++ b/litex/soc/interconnect/wishbone.py
@@ -35,7 +35,7 @@ _layout = [
 
 
 class Interface(Record):
-    def __init__(self, data_width=32, adr_width=30):
+    def __init__(self, data_width=32, adr_width=31):
         self.data_width = data_width
         self.adr_width  = adr_width
         Record.__init__(self, set_layout_parameters(_layout,
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -93,7 +93,7 @@ class RocketRV64(CPU):
         flags += "-D__rocket__ "
         return flags
 
-    def __init__(self, platform, variant="standard", use_memory_bus=True):
+    def __init__(self, platform, variant="standard"):
         self.platform  = platform
         self.variant   = variant
 
@@ -102,18 +102,16 @@ class RocketRV64(CPU):
 
         mem_dw, mmio_dw = AXI_DATA_WIDTHS[self.variant]
 
-        self.mem_axi   =  mem_axi = axi.AXIInterface(data_width= mem_dw, address_width=32, id_width=4)
+        self.mem_axi   =  mem_axi = axi.AXIInterface(data_width=mem_dw,  address_width=32, id_width=4)
         self.mmio_axi  = mmio_axi = axi.AXIInterface(data_width=mmio_dw, address_width=32, id_width=4)
+        self.l2fb_axi  = l2fb_axi = axi.AXIInterface(data_width=mmio_dw, address_width=32, id_width=4)
 
-        self.mem_wb    = mem_wb  = wishbone.Interface(data_width=mem_dw,  adr_width=32-log2_int(mem_dw//8))
         self.mmio_wb   = mmio_wb = wishbone.Interface(data_width=mmio_dw, adr_width=32-log2_int(mmio_dw//8))
+        self.l2fb_wb   = l2fb_wb = wishbone.Interface(data_width=mmio_dw, adr_width=32-log2_int(mmio_dw//8))
 
+        self.memory_buses = [mem_axi]
         self.periph_buses = [mmio_wb]
-        self.memory_buses = []
-        if use_memory_bus:
-            self.memory_buses.append(self.mem_axi)
-        else:
-            self.periph_buses.append(self.mem_wb)
+        self.dma_bus      =  l2fb_wb
 
         # # #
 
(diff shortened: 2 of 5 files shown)


litex/soc/interconnect/wishbone.py
```diff
@@ -35,7 +35,7 @@
 
 
 class Interface(Record):
-    def __init__(self, data_width=32, adr_width=30):
+    def __init__(self, data_width=32, adr_width=31):
```
enjoy-digital, line 38: @gsomlo: i think this is breaking tests and CI, could we keep 30 as default for now?

gsomlo: I don't think we can have a memory region larger than 2GB (`0x8000_0000`) (e.g., [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L848)) without increasing the `addr_width` by one bit, to `31`.

We can hold off on applying this PR until you have a chance to
  1. decide whether bumping to `31` is, in fact, OK to do
  2. update CI to stop complaining about it :)

But I think the change is necessary to cover the entire 4GB address space with a single memory region.
**Unless** we can somehow cover a single slave port with *two* memory regions (each of them 2GB in size)? Not sure if that's possible, but I'm certainly of the opinion that bumping `adr_width` would be cleaner...

gsomlo:
@enjoy-digital: re-worked the patch series to also address CI complaints re. relationship between address width of AXI_Lite and WB (in the axi_lite <-> wishbone converter modules) -- see new commit [7a20ced](https://github.com/enjoy-digital/litex/pull/617/commits/7a20ced1f4fad2af93857fce708e81f4364687a6). CI appears happy now.

gsomlo:
@enjoy-digital: now I'm really confused. With this:
```
diff --git a/litex/soc/interconnect/axi.py b/litex/soc/interconnect/axi.py
index ec25bbae..f1b57ed5 100644
--- a/litex/soc/interconnect/axi.py
+++ b/litex/soc/interconnect/axi.py
@@ -496,7 +496,7 @@ class AXILite2Wishbone(Module):
     def __init__(self, axi_lite, wishbone, base_address=0x00000000):
         wishbone_adr_shift = log2_int(axi_lite.data_width//8)
         assert axi_lite.data_width    == len(wishbone.dat_r)
-        assert axi_lite.address_width == len(wishbone.adr) + wishbone_adr_shift
+        assert axi_lite.address_width == len(wishbone.adr) + wishbone_adr_shift - 1
 
         _data         = Signal(axi_lite.data_width)
         _r_addr       = Signal(axi_lite.address_width)
@@ -580,7 +580,7 @@ class Wishbone2AXILite(Module):
     def __init__(self, wishbone, axi_lite, base_address=0x00000000):
         wishbone_adr_shift = log2_int(axi_lite.data_width//8)
         assert axi_lite.data_width    == len(wishbone.dat_r)
-        assert axi_lite.address_width == len(wishbone.adr) + wishbone_adr_shift
+        assert axi_lite.address_width == len(wishbone.adr) + wishbone_adr_shift - 1
 
         _cmd_done  = Signal()
         _data_done = Signal()
```
CI was happy, but the assertions failed on build (rocket *and* vexriscv). Without the above changes, builds work fine (both vexriscv and rocket), but CI fails....

gsomlo:
The CI problem is caused by these two lines:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/axi.py#L499
and
https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/axi.py#L583
both of which look like this:
```
```
Now, for some reason, when I build rocket (even after having bumped wishbone adr_width to 31), I have:
```
axi_lite.address_width == 32
len(wishbone.adr) == 29
wishbone_adr_shift == 3
```
The one thing I'm not quite sure about is why `len(wishbone.adr) == 29`. I think the assertion passes by coincidence, and I still need to figure out why.

When CI runs its tests, the values are:
```
axi_lite.address_width == 32
len(wishbone.adr) == 31
wishbone_adr_shift == 2
```
... which seems perfectly reasonable, given that I set the WB address width to 31 myself, and the shift is 2 (for 32 bit data width, just like 3 above was OK for 64 bit data width). The question remains why I get `len(wishbone.adr) == 29` with Rocket, and how to modify the assertion to actually check for the correct relationship between those values in a generally applicable way (to both 32- and 64-bit architetures).

I'll keep investigating tomorrow.

enjoy-digital:
Thanks @gsomlo for the PR and investigation. I fixed the CI tests with https://github.com/enjoy-digital/litex/commit/a5d0a340c3b19c8c1eee7b2967db36199585461c, sorry for the trouble with this, we'll probably have to improve handling of `address_width` in the future on the `buses` to ease things.

PR #619, opened 2020-08-04
Allow to define multiple simulation clocks

jedrzejboczar:
Currently in `litex_sim` it is not possible to easily create clocks that have higher frequency than `sys_clk` or a different phase. This PR allows to create multiple `clocker` simulation modules based on `SimConfig`.

This update will be useful for simulating RPC DRAM https://github.com/enjoy-digital/litedram/issues/211, and similar changes were already needed and implemented for Mircon model simulations in https://github.com/enjoy-digital/litex/pull/536 by @mglb. This PR bases on that previous work and extracts only the part of configuring clocks in `litex_sim` and cleans it up.

The simulation now keeps track of the current time and modules' API has been updated to pass the current simulation time to each module's `tick()` method. The whole simulation uses a single timebase with resolution of 1 ps. The time is updated in `sim.c` and passed to modules and `veril.cpp`, where it is used when saving dump files, so the time in GTKWave is correct now.

Because we can have multiple clocks, each simulation module has a `clk_edge_t`, which is a simple clock edge detector added to avoid boilerplate code in each module.

A clocker module now takes a clock and specified frequency/phase, and on each step updates the clock state. Multiple clockers can be added for different clocks (different `_io` pins) using the `SimConfig.add_clocker(self, clk, freq_hz, phase_deg=0)` method.  Other modules still have hard-coded `"sys_clk"` clocks in their C source code, but this could be changed if needed. `SimConfig` has been changed to require to add clockers explicitly (at least one), passing frequency and phase, e.g.
```python
sys_clk_freq = int(100e6)
sim_config = SimConfig()
sim_config.add_clocker("sys_clk",       freq_hz=sys_clk_freq)
sim_config.add_clocker("sys2x_90_clk",  freq_hz=2*sys_clk_freq, phase_deg=90)
```
The data is then stored in the JSON file like this:
```json
[
    {
        "module": "clocker",
        "interface": [
            "sys_clk"
        ],
        "args": {
            "freq_hz": 100000000,
            "phase_deg": 0
        },
        "tickfirst": true
    },
    {
        "module": "clocker",
        "interface": [
            "sys2x_90_clk"
        ],
        "args": {
            "freq_hz": 200000000,
            "phase_deg": 90
        },
        "tickfirst": true
    },
    {
        "module": "serial2console",
        "interface": [
            "serial",
            "sys_clk"
        ]
    },
    {
        "timebase": 1250
    }
]
```

The timebase is calculated by `SimConfig` based on the specification of all clockers. Currently it is very restrictive, as it will not accept non-integer half-periods or phase-shifts. It finds the timebase as the gratest common denominator of all half-periods/phase-shifts. This results in the simulation failing even for clock frequencies like 150 MHz (period = 6666.(6) ps), so the current implementation is best suited for simple clock multiples like x2 and x4 and simple phase shifts, like 90 or 180 degrees.

Here is an example of running litex_sim with more clocks (sys_clk = 100 MHz, calculated timebase = 1250 ps): 
sys_clk_100MHz

### Improvements

I first wanted to loosen the restrictions on the timebase, but this can introduce subtle bugs, and is hard to get right. Ultimately I think that, if we want to support more clock configurations, we should move the clocker logic to the simulation core (`sim.c`), store the time as floating point number, and just use dynamic timebase by always moving to the next edge of all existing clocks.

Another note is that the simulation code is probably a bit too low-level and hard to extend. As Verilator uses C++ anyway, we could probable have it all done in C++ and simplify the whole error handling, memory management (with C++ stdlib instead of manually allocated lists), JSON parsing, etc. After adding the `spdeeprom.c` module and working on this PR I think there's quite a bit of boilerplate work when one wants to add a new module, and we're not forced to use C as this is a simulator.

Both the dynamic time stepping and general rewrite would require some time to complete and are not of high priority as for now, but maybe we could think about it in the future?

### DIFF
--- a/litex/build/sim/config.py
+++ b/litex/build/sim/config.py
@@ -3,14 +3,12 @@
 # License: BSD
 
 import json
+import math
 
 class SimConfig():
-    def __init__(self, default_clk=None, timebase_ps=1):
+    def __init__(self, timebase_ps=None):
         self.modules = []
-        self.default_clk = default_clk
-        self.timebase = timebase_ps
-        if default_clk:
-            self.add_clocker(default_clk)
+        self.timebase_ps = timebase_ps
 
     def _format_interfaces(self, interfaces):
         if not isinstance(interfaces, list):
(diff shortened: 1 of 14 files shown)


enjoy-digital:
Thanks @jedrzejboczar, nice improvements! This was indeed initially planned for LiteDRAM but simulating at the DFI level allowed keeping a single clock. This will be useful for several cases and even with the actual limitation (that are generally fine for the simulations we want to run with this tool) it's a real improvement.

I just added https://github.com/enjoy-digital/litex/commit/696ea468b8c2d1061e91baa147092689a11203e6 and https://github.com/enjoy-digital/litex/commit/0ca99b798fccec85d689188d5846d160c183adff for retro-compatibility with old json-c version and simulation already build with the tool.

For the future improvements, i agree with you: it's currently working and useful, but it's a bit hacky and could be largely improved. If you have ideas on this (and eventually switch to C++) please feel free to go ahead on this if/when you have time. We could also try to get rid of some of the dependency: tapcfg, tinydir, etc... and get back some features that were with the initial proof of concept like the Framebuffer emulation with SDL.

PR #620, opened 2020-08-04
Add Linux DT generation script

mateusz-holenko:
This moves the `json2dts.py` script (with some minor modifications) from `linux-on-litex-vexriscv`.
Having it in LiteX will allow other projects to generate the DT file for Linux much easier.

In the next step (after this is merged to master) I will remove the copy of the script from `linux-on-litex-vexriscv` and modify `soc_linux.py` and `sim.py` to use the one from LiteX (see: https://github.com/antmicro/linux-on-litex-vexriscv/commit/9692ac8105085fceff47ec93f49cf90819e67f49).

In the future I plan to extend this script to generate the DT file compatible with Zephyr.

### DIFF
--- a/litex/tools/litex_json2dts.py
+++ b/litex/tools/litex_json2dts.py
@@ -1,0 +1,521 @@
+#!/usr/bin/env python3
+
+import sys
+import json
+import argparse
+
+
+def generate_dts(d):
+
+    kB = 1024
+    mB = kB*1024
+
+    aliases = {}
+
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @mateusz-holenko, this is merged and i just added https://github.com/enjoy-digital/litex/commit/aed0dcee4ca3bbeb8feafb62ba283147b52bdedc on top to expose it as a script while LiteX is installed.

PR #621, opened 2020-08-04
propose modifcation for 4GiB addressing that doesn't break the API

bunnie:
For consideration as an alternative to

https://github.com/enjoy-digital/litex/commit/70eae5cbf9faf9df5832f39cae455087aceed7b7

This is more meant to be a conceptual approach, of course probably
the naming and execution can be improved.

This is in response to the issue created by @gsomlo commit https://github.com/enjoy-digital/litex/commit/70eae5cbf9faf9df5832f39cae455087aceed7b7

### DIFF
--- a/litex/soc/cores/cpu/vexriscv/core.py
+++ b/litex/soc/cores/cpu/vexriscv/core.py
@@ -104,15 +104,19 @@ class VexRiscv(CPU, AutoCSR):
         flags += " -D__vexriscv__"
         return flags
 
-    def __init__(self, platform, variant="standard"):
+    def __init__(self, platform, variant="standard", extended_addressing=False):
         self.platform         = platform
         self.variant          = variant
         self.human_name       = CPU_VARIANTS.get(variant, "VexRiscv")
         self.external_variant = None
         self.reset            = Signal()
         self.interrupt        = Signal(32)
-        self.ibus             = ibus = wishbone.Interface()
-        self.dbus             = dbus = wishbone.Interface()
+        if extended_addressing:
+            self.ibus = ibus = wishbone.Interface(adr_width=31)
+            self.dbus = dbus = wishbone.Interface(adr_width=31)
+        else:
+            self.ibus = ibus = wishbone.Interface()
+            self.dbus = dbus = wishbone.Interface()
         self.periph_buses     = [ibus, dbus]
         self.memory_buses     = []
 
--- a/litex/soc/cores/cpu/vexriscv/core.py
+++ b/litex/soc/cores/cpu/vexriscv/core.py
@@ -104,19 +104,15 @@ class VexRiscv(CPU, AutoCSR):
         flags += " -D__vexriscv__"
         return flags
 
-    def __init__(self, platform, variant="standard", extended_addressing=False):
+    def __init__(self, platform, variant="standard"):
         self.platform         = platform
         self.variant          = variant
         self.human_name       = CPU_VARIANTS.get(variant, "VexRiscv")
         self.external_variant = None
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @bunnie for looking at this and sorry for the troubles (before you got back to us i received a bunch or mails from you CI :), so was going to contact you to see what  was happening), that's very useful to have early feedback on this and your CI in conjunction with ours has already been very useful!

The orignal adr_width has been reverted with: https://github.com/enjoy-digital/litex/commit/b831dc8c55148fcd9506d19875a057b56a704c0a and the change for the case we wanted to support has been minimized with: https://github.com/enjoy-digital/litex/commit/b1008b01642c12c8f9d2253372f027ebd659b9e5.

bunnie:
Sorry for the CI spam -- the first couple hours of it was actually debugging a deeper problem this triggered in the reproducibility of the build system, and I didn't want to file a bug until I was sure this was the issue, and not a problem with our CI configuration. Basically, the system was too aggressive about checking out a particular version of the code, which made it impossible to test our patches before submitting a PR. 

I think we got it sorted out, and hopefully in the future you won't see as many messages while we trace down such problems....

mithro:
@bunnie Do you have more info about your CI stuff? It sounds like something we might be able to collaborate on and/or reuse for other projects?

bunnie:
You can see it on https://ci.betrusted.io. It's a box that sits in my apartment and runs Ubuntu 18, so we can install pretty much anything on it that we like (critically, the Vivado tools). This is the box I also used to do the fuzzing of the ROM locations on the Spartan-7 part by rebuilding the prjxray tools. 

We're using Jenkins. We have a couple build configurations, one checks our "pinned" versions of Litex, another takes all our gateware submodules (which have CI unit tests that run in xsim with a VCD interpreter back-end so we can extract pass/fail data from waveforms automatically) and continuously builds them against the latest pushes of Litex. This is what picked up the bug in this thread, but as noted, we still are working out the kinks in the configuration....

We also have an actual betrusted device plugged into it, so after we do builds we can automatically flash the bitstream onto it, and there's a webcam so we can confirm that it worked by looking at the screen, and we can use wishbone-tool to poke registers via USB over the Internet. The long term vision is to close the loop from push to actual hardware in CI, but exposing this feature more publicly is a bit of a WIP. However, the basic bits are there to do some pretty deep dev work on the hardware remotely -- we put it together after stuff went into lockdown since we all had to do WFH anyways.

Xobs also uses it to do bleeding-edge CI against Litex as well for the Fomu project. There's also Tensorflow and an RTX2070 on it for unrelated reasons, but it's there, and currently under-utilized.

There's some spare capacity on it, our current main limiting factor is SSD space to hold build artifacts, but that's easy enough to upgrade if you thought this server could be useful.

mithro:
@tcal-x @carlfk @kgugala -- You should check out what @bunnie and @xobs have set up here!

FYI -- I would be more than happy to provide a cloud machine for this type of stuff if needed.

xobs:
To add on to what @bunnie said, he generates a ROM for Betrusted that runs the test. This ROM sets a "PASS" wire to `1` when a test passes. I seem to recall he included a "FAIL" wire as well, or the test is allowed to run out.

He can't use Cocotb because xsim doesn't support VPI -- you insert your code and hit "play", with xsim there's no way to control a simulation in-flight.

He also can't use Verilator because of Xilinx-specific blocks such as the PLLs and IO blocks, all of which have simulation models under xsim and none of which work under Verilator.

But overall the solution works well, as can be evidenced by this PR.

PR #622, opened 2020-08-05
arty: Change USB-uart and I2S Pmod configuration

mateusz-holenko:
This makes it compatible with the Arty A7 expansion board by Antmicro
(https://github.com/antmicro/arty-expansion-board).


enjoy-digital:
Thanks, merged.

PR #624, opened 2020-08-06
soc/cores/cpu/zynq7000: add enet0, enet0_mdio, sdio, sdio_cd and sdio_wp only when configured in EMIO mode

trabucayre:
Most of hardware Zynq's peripherals may be connected to processor dedicated pins (PS_xxx) or, using EMIO mode, to FPGA dedicated pins (IO_xxx).

Corresponding group of pins, in processing_system module, are only present when the peripheral is configured in EMIO mode.

This PR allows to have both mode. Currently limited to ENET0 and SDIO0.

### DIFF
--- a/litex/soc/cores/cpu/zynq7000/core.py
+++ b/litex/soc/cores/cpu/zynq7000/core.py
@@ -73,12 +73,6 @@ class Zynq7000(CPU):
             io_DDR_VRN           = ps7_ddram_pads.vrn,
             io_DDR_VRP           = ps7_ddram_pads.vrp,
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks! This is merged and has been simplified a bit with https://github.com/enjoy-digital/litex/commit/57335b99718e3cd93cd61049cafdb20e1ea71cac to avoid the try/except.

trabucayre:
Thanks for this simplification!
I don't know why I didn't think of this solution. Anyway there are now a good base to continue improving some aspect.

PR #626, opened 2020-08-08
Set default csr paging to 0x1000

Disasm:
This allows to calculate peripheral address on RISC-V with one instruction instead of two.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -482,7 +482,7 @@ class SoCCSRHandler(SoCLocHandler):
     supported_ordering      = ["big", "little"]
 
     # Creation -------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks @Disasm, i'll have to evaluate the impact of changing the default value on existing systems, but unfortunately i'm currently in vacation and will only be able to look at it in early september.

enjoy-digital:
@Disasm: sorry for the delay. I'd like to keep `csr_paging` default value to 0x800 to prevent regressions or update on some previous SoCs built with LiteX. Are you already able to set `csr_paging` to `0x1000` manually? I would however be OK to set it by default to 0x1000 on a RISC-V based SoC.

Disasm:
No problem, thanks for the answer! It's easy to overwrite this parameter anyway.

PR #627, opened 2020-08-10
RFC: cores/dma, liblitesdcard/sdcard: use 64 bits for dma base address

gsomlo:
Currently, the DMA base addresses use 32 bit registers, on the implicit assumption that physical address space is limited to 4GB. That's not true for LiteX SoC versions with 64bit CPUs, which may end up with physical RAM in excess of 4GB, requiring 64 bit address registers to cover all possible addresses that may be accessed by a DMA master.

This patch (tested with rocket and vexrisccv) bumps the DMA base address to 64 bit (length/size is left as a 32 bit register, since a 4GB limit per individual DMA transfer "should be enough for everyone" ;)

### DIFF
--- a/litex/soc/cores/dma.py
+++ b/litex/soc/cores/dma.py
@@ -72,7 +72,7 @@ class WishboneDMAReader(Module, AutoCSR):
             self.add_csr()
 
     def add_csr(self):
-        self._base   = CSRStorage(32)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, this makes sense and has been merged.

PR #631, opened 2020-08-15
build/lattice/trellis: make "-abc9" an optional argument

gsomlo:
Fix up earlier commit (#6c298cb7) and make the '-abc9' optional
argument to yosys' synth_ecp5 actually optional (and off by default)
in LiteX's trellis build infrastructure.

@enjoy-digital @daveshah1 What do you think? IMHO I should have done it like this from the beginning, any reason we shouldn't correct that now? :)

### DIFF
--- a/litex/build/lattice/trellis.py
+++ b/litex/build/lattice/trellis.py
@@ -54,7 +54,7 @@ _yosys_template = [
     "{read_files}",
     "verilog_defaults -pop",
     "attrmap -tocase keep -imap keep=\"true\" keep=1 -imap keep=\"false\" keep=0 -remove keep=0",
-    "synth_ecp5 -abc9 {nwl} -json {build_name}.json -top {build_name}",
+    "synth_ecp5 {nwl} {abc} -json {build_name}.json -top {build_name}",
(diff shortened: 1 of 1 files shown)


mithro:
I'm pretty sure `abc9` should be **on** by default and bugs reported where it performs worse when off?

gsomlo:
abc9 is not on by default in yosys, so maybe making it on by default
in litex was, um, "premature" ?

Right now, about half the time, yosys fails with abc9 (when I build
rocket-based litex socs for versa or the trellisboard).

I reported it upstream here: https://github.com/berkeley-abc/abc/issues/84

At the moment, the proposed solution
(https://github.com/berkeley-abc/abc/pull/89) is applied in the Fedora
package, but doesn't seem to actually fix abc9 failures reliably.

The upstream abc maintainer has not (yet) responded.

enjoy-digital:
@gsomlo: for info, `abc9` was also causing issue with `bus-standard=axi-lite` on ECP5 designs, without it it's now fine, so this confirms it's probably better to disable it by default.

PR #640, opened 2020-08-24
litex_json2dts: Add support for mor1kx

mateusz-holenko:
This PR extends the `litex_json2dts.py` script by adding support for the mor1kx CPU.


enjoy-digital:
Thanks @mateusz-holenko.

PR #642, opened 2020-08-27
RFC: bios/sdcard: use (512 byte) blocks as the smallest addressable data unit

gsomlo:
In liblitesdcard/sdcard.c, replace s/sector/block/ as the 512 byte sized
unit of data transfer to/from the sdcard.

In bios/cmds/cmd_litesdcard.c, do NOT multiply block numbers by 512,
allowing any "block" (a.k.a. "sector") on the sdcard to be addressed for
testing.

Before this patch, user-input "block numbers" were multiplied by 512 and
passed into sdcard_[read|write]() as "sector numbers", with the sdcard
logic internally treating these as 512-byte sized addressable units on
the card. This resulted in only every 512'th 512-byte sized "sector" being
accessible from the bios command line.

### DIFF
--- a/litex/soc/software/bios/cmds/cmd_litesdcard.c
+++ b/litex/soc/software/bios/cmds/cmd_litesdcard.c
@@ -73,7 +73,7 @@ static void sdread(int nb_params, char **params)
 		return;
 	}
 
-	sdcard_read(block*512, 1, buf);
+	sdcard_read(block, 1, buf);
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @gsomlo, sorry for the delay. I wanted to check before merging with a dump from the BIOS and hexdump on the Host.

PR #643, opened 2020-08-30
crt0: fix .data section initialise on [Serv,Minerva]

gregdavill:
The startup code in `crt0.S` was incorrectly initialising the .data section.

The old code was copying from `_fdata` into `_fdata_rom`. 

This likely hasn't been an issue in the bios before, but with the changes to the command parser if the `.data` section isn't initialised then none of the commands work correctly, since they're stored in `.data`
```
--============= Console ================--

litex> [tab]o[tab]
o    o    o
o    o    o    o    o    o    o    
litex> help
Command not found
litex> 
```


enjoy-digital:
Thanks for catching and fixing this!

PR #644, opened 2020-08-31
integration/soc: use csr.add() instead of add_csr()

Xiretza:
`add_csr()` is defined by `SoCCore`, so won't work for any instances of `LiteXSoC` that aren't also `SoCCore`s. Also, use `use_loc_if_exists=True` so `SoCCore.csr_map` can be used without double allocation errors.


enjoy-digital:
Thanks, i was indeed planning to update it, let's do it now if this is causing troubles.

PR #647, opened 2020-09-07
Improve dynamic tracing support in the simulator

jedrzejboczar:
This adds some tools for better integration of simulation tracing. It allows for basic control of the simulator from withing the simulation: stop/start tracing, mark points in time, finish simulation.

The solution is to use a single `sim_trace` output pin from the design. The simulation can enable/disable signal tracing by changing the value of this pin. This way we can easily run a full simulation and enable tracing only in the interesting parts, which speeds up the simulation and avoids creating huge dumps. The time in the dump will still be correct, and the parts of simulation where tracing was disabled will visibly have no changes in signal values.

I also decided to make the simulator print `<DUMP ON>`/`<DUMP OFF>` when the tracing state changes so that it's easy to see when the dump is on while debugging, even though this can mess up the messages printed by the software. 

The debugging must be enabled by calling `SimPlatform.add_debug` method. This add some CSRs: one to control tracing state (on/off), one that will finish the simulation when written and one that holds an 8-bit marker value, which can be then checked in GTKWave to localize parts of the dump.

The CSRs could be written from gateware or from firmware. LiteX BIOS defines helper functions for writing to the CSRs (`sim_debug.c`). One than uses the function as follows:
```c
int foo(void) {
    sim_mark_func();
    // ...
}

sim_trace(1);
sim_mark("Some point in time");
foo();
sim_trace(0);
sim_finish();
```
This will trace only the call to `foo()` and then end simulation. Ending simulation with `sim_finish` will also print a summary of the mapping between the 8-bit marker values and the comments passed to `sim_mark`. These functions fall back to writing error messages when the CSRs are not present. Not sure if this is the best behaviour, it certainly helps to find if we forgot `.add_debug()` but maybe we should get rid of the messages? On top of that BIOS provides commands for the shell that just call the mentioned functions.

The debug CSRs are enabled in `litex_sim` with the `--sim-debug` argument. With the current desing the `sim_trace` pin must always be present in Verilator simulations, so `SimPlatform` will insert it and it will be driven with the value of 1, which makes the behaviour the same as before this PR. With `--sim-debug` the dump is initially disabled if none of the `--trace-start`/`--trace-end` parameters is provided.

This PR also fixes an error in `--trace-start`/`--trace-end` parameters, as they now use the unit of in-simulation picoseconds instead of clock cycles.

### DIFF
--- a/litex/build/sim/core/veril.cpp
+++ b/litex/build/sim/core/veril.cpp
@@ -20,6 +20,7 @@ VerilatedVcdC* tfp;
 uint64_t tfp_start;
 uint64_t tfp_end;
 uint64_t main_time = 0;
+Vsim *g_sim = nullptr;
 
 extern "C" void litex_sim_eval(void *vsim, uint64_t time_ps)
 {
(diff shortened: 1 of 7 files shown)


derekmulcahy:
A small deal, when compiling I get

`litex/soc/software/libbase/sim_debug.c:10:20: warning: 'markers' defined but not used [-Wunused-variable]`

 Perhaps the declaration code could be wrapped in

```
#ifdef CSR_SIM_MARKER_BASE
...
#endif
```

like the function bodies.

PR #649, opened 2020-09-15
misc. fixes for a few recent commits

gsomlo:
A few miscellaneous clean-up fixes for recent commits:
- commit e28e808 defines an array in a .h file, which then "conflicts with itself" at link time when included multiple times in multiple files
- sim_debug marker variables should (probably) be wrapped inside the same `#ifdef` statements as when referenced, or else we're getting "unused" warnings about them during compilation
- memtest: a bunch of places convert pointers to `unsigned int`, which should really be `unsigned long` to work seamlessly on both 32 and 64 bit CPUs


enjoy-digital:
Thanks, sorry for this (i was focusing on other aspects while working on corner LiteDRAM's calibration corner cases on some hardware).

PR #650, opened 2020-09-17
I2c

alanvgreen:
These two changes are independent: you could take none, one or both!

### DIFF
--- a/litex/soc/cores/bitbang.py
+++ b/litex/soc/cores/bitbang.py
@@ -16,8 +16,12 @@ class I2CMaster(Module, AutoCSR):
 
     Provides the minimal hardware to do software I2C Master bit banging.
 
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @alanvgreen, i had a first look. We should probably just keep your `I2CMasterOpenDrain` implementation. I'll have a closer look at it in the next days.

enjoy-digital:
Sorry @alanvgreen for the delay, thanks your feedback and proposed solution. The PR hasn't been merged directly, but i modified `I2CMaster`to use Tristate and only drive SCL/SDA low with https://github.com/enjoy-digital/litex/commit/ce9f24748f8af623bbbcf76b3555a37b0b2d76cc. This should be equivalent to your I2CMasterOpenDrain implementation.

PR #651, opened 2020-09-21
software/liblitedram: fix conditional compilation bug

gsomlo:
Fix conditional compilation bug introduced by a601415b.

@enjoy-digital At least it fixes it for me, on the nexys4ddr with a rocket-based litex build. Not 100% sure we need both or if maybe checking just for `defined(SDRAM_PHY_WRITE_LEVELING_CAPABLE)` may be enough... Either way, if you have a more precise fix, please feel free to discard this PR... :)


enjoy-digital:
Thanks for reporting and fixing @gsomlo, i fixed it a bit differently with https://github.com/enjoy-digital/litedram/commit/0279b770eeb55038676aae4c53c80d57d189a1a5.

PR #652, opened 2020-09-22
RFC: proper way to set sdcard clock?

gsomlo:
With the way the clock divider is currently calculated, it is
possible for the effective sdcard clock to end up **higher** than
the requested `clk_freq` value.

With this PR, we instead calculate the divider starting with the `CONFIG_CLOCK_FREQUENCY/clk_freq`
ratio which is then rounded **up** to the nearest power-of-two, ensuring
that the resulting, effective clock frequency is <= to the requested
frequency.

@enjoy-digital: this is mostly just me wondering if there's any harm in setting the effective clock at a higher frequency than what's officially requested via the argument to `sdcard_set_clk_freq()`. If that's not a problem, please feel free to disregard this PR :)

### DIFF
--- a/litex/soc/software/liblitesdcard/sdcard.c
+++ b/litex/soc/software/liblitesdcard/sdcard.c
@@ -89,29 +89,34 @@ int sdcard_wait_data_done(void) {
 /* SDCard clocker functions                                              */
 /*-----------------------------------------------------------------------*/
 
-static uint32_t log2(uint32_t x)
-{
-	uint32_t r = 0;
-	while(x >>= 1)
-		r++;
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @gsomlo, i'm indeed aware of this. This was implemented like this to avoid lowering too much the frequency in some cases due to the limited steps of the power-of-two divider. The maximum SDCard clock is already lowered in software and constrained to sys_clk_freq, so it should not cause issues, but that's indeed probably better to use your solution and maybe improve the clock divider later or re-discuss this if the transfers are too slow.

PR #654, opened 2020-09-28
Build support for Gowin

pepijndevos:
This PR aims to add support for Gowin FPGAs to LiteX.

Initially it only supports the vendor tools, but I'd really like to add support for Yosys+Apicula in the future.

At this point I can run a basic blinky, but stuff is likely missing and broken, so WIP. I'll try some more elaborate examples, and update as needed.

I just want to get this out here to get some feedback on dumb things I might be doing. I'm basically copying from the other build targets without fully understanding every little detail.

### DIFF
--- a/litex/build/gowin/gowin.py
+++ b/litex/build/gowin/gowin.py
@@ -62,10 +62,6 @@ class GowinToolchain():
     def __init__(self):
         self.options = {}
 
-    def attr_translate(*args):
-        print(args)
-        pass
-
     def build(self, platform, fragment,
         build_dir      = "build",
         build_name     = "top",
@@ -90,7 +86,7 @@ class GowinToolchain():
         platform.add_source(v_file)
         
         if platform.verilog_include_paths:
-            self.options['includ_path'] = '{' + ';'.join(platform.verilog_include_paths) + '}'
+            self.options['include_path'] = '{' + ';'.join(platform.verilog_include_paths) + '}'
 
         # Generate constraints file (.cst)
         _build_cst(
--- a/litex/build/gowin/gowin.py
+++ b/litex/build/gowin/gowin.py
@@ -99,7 +99,7 @@ class GowinToolchain():
 
         # Generate TCL build script
         script = _build_script(
-            name                    = platform.name,
+            name                    = platform.devicename,
             partnumber              = platform.device,
             files                   = platform.sources,
             options                 = self.options)
--- a/litex/build/gowin/gowin.py
+++ b/litex/build/gowin/gowin.py
@@ -59,6 +59,19 @@ def _build_script(name, partnumber, files, options):
         f.write(tcl)
 
 class GowinToolchain():
+
+    attr_translate = {
+        "keep":             None,
+        "no_retiming":      None,
+        "async_reg":        None,
+        "mr_ff":            None,
+        "mr_false_path":    None,
+        "ars_ff1":          None,
+        "ars_ff2":          None,
+        "ars_false_path":   None,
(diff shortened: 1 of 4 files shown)


litex/build/gowin/gowin.py
```diff
@@ -0,0 +1,113 @@
...
+
+    tcl = "\n".join(lines)
+    with open("run.tcl", "w") as f:
+        f.write(tcl)
+
+class GowinToolchain():
+    def __init__(self):
+        self.options = {}
+
+    def attr_translate(*args):
```
enjoy-digital, line 65: You can copy this: https://github.com/enjoy-digital/litex/blob/master/litex/build/microsemi/libero_soc.py#L219-L230

litex/build/gowin/gowin.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Generate verilog
+        v_output = platform.get_verilog(fragment, name=build_name, **kwargs)
+        named_sc, named_pc = platform.resolve_signals(v_output.ns)
+        v_file = build_name + ".v"
+        v_output.write(v_file)
+        platform.add_source(v_file)
+        
+        if platform.verilog_include_paths:
+            self.options['includ_path'] = '{' + ';'.join(platform.verilog_include_paths) + '}'
```
enjoy-digital, line 93: Typo? -->` 'include_path'`

enjoy-digital (changes requested):
Thanks, i just made a few comments but otherwise it looks good for an initial support.

litex/build/gowin/gowin.py
pepijndevos: I copied the code for `get_verilog` from somewhere that had it, but saw that another target did not have it at all. I'm not clear what this function does, and what the correct implementation would be for Gowin. I actually thought I deleted the entire function. So yea I can copy those lines, but I'd like to understand why it's needed and have some code that test the functionality.

litex/build/gowin/gowin.py
pepijndevos: Oops. Another instance of functionality I implemented, but am not clear how it is used and how to test it. All the cores I've tried so far seem to just provide all the source files. I'll fix the typo, but would be glad if there is an example I can test this with.

litex/build/gowin/gowin.py
enjoy-digital: This is used to translate attributes that we add to internal signals in the Migen/LiteX design to verilog attributes (that are different depending the toolchain). The `keep` attribute for example is used to specify that the signal should not be simplified during synthesis and we are generally using it to specify timing constraints. Most of these attributes are only used with the Xilinx toolchains and this should be simplified/refactored on the other toolchains, but i just suggest you copy the same structure to ease refactoring later.

pepijndevos: Alright, added dummy translations.

pepijndevos:
Okay, I've addressed your requests. At the moment I have not yet actually managed to run any code on the Gowin besides the BIOS, so I would like to keep this PR open a little longer until I am actually certain it can produce a working SoC.

Meanwhile, if there is anything I could do to test and improve this target, I'd be glad to hear it. I think my main limitation is code that exercises the target a bit more than the simple.py core. I'm not using any overrides, attributes, include paths, etc. But maybe that's just good enough for "initial support" as you mention.

enjoy-digital:
We can merge this PR to simplify collaborating on this, this initial build support should be enough to create simple SoCs and we'll improve it progressively. Thanks a lot for your work!

PR #655, opened 2020-09-28
add memory regions to soc.svd

bunnie:
svd2rust does not recognize memory regions, but we'd like to
make an access crate for Rust that does.

This patch adds memory regions to soc.svd using the "vendorExtensions"
tag, as specified in https://www.keil.com/pack/doc/cmsis/SVD/html/svd_Format_pg.html

The vendorExtensions is added as a block after the Peripherals level, and has a format
like this:

```xml
        <memoryRegion>
            <name>SRAM</name>
            <baseAddress>0x10000000</baseAddress>
            <size>0x00020000</size>
        </memoryRegion>
        <memoryRegion>
            <name>VEXRISCV_DEBUG</name>
            <baseAddress>0xEFFF0000</baseAddress>
            <size>0x00000100</size>
        </memoryRegion>
        <memoryRegion>
            <name>CSR</name>
            <baseAddress>0xF0000000</baseAddress>
            <size>0x00040000</size>
        </memoryRegion>
```


litex/soc/integration/export.py
```diff
@@ -428,6 +428,15 @@ def print_svd_register(csr, csr_address, description, length, svd):
             svd.append('            </interrupt>')
         svd.append('        </peripheral>')
     svd.append('    </peripherals>')
+    if len(soc.mem_regions) > 0:
+        svd.append('    <vendorExtensions>')
+        for name, region in soc.mem_regions.items():
```
xobs, line 433: Maybe put all of the memory regions inside a `<memoryRegions>` tag, similar to how all the `<peripheral>` tags are nested inside of a big `<peripherals>` tag? That way if we need to add more vendor extensions later on, we have the ability to do so.

xobs (commented):
Overall I like the idea. It's non-intrusive and gets the job done. I just think it needs to go inside a `<memoryRegions>` tag.

mithro:
I like this, but why not add memory region support to upstream svd2rust?

xobs:
@mithro that's one idea. And we should open a dialogue with them about it.

This stems from a fundamental difference in how register accesses work in an operating system.

rust2svd compiles register accesses down to a single memory operation, and is designed around a fixed address for memory regions. This doesn't work so well in a virtual memory environment, where those addresses could be used by the operating system.

If you take the sizeof() a rust2svd Peripheral struct, you'll find it's 0 bytes large. A fantastic trick, and great for many embedded systems.

For an os you need a library that defines fields, yet leaves offsets settable.

Maybe these both can work! Either way, we'll need memory ranges reported in the svd file, and I'm surprised that format doesn't natively support it.

bunnie:
I've incorporated @xobs suggested and added commit #7b42992.

bunnie:
> I like this, but why not add memory region support to upstream svd2rust?

Also for what it's worth after using soc2svd for over a year, I have disagreements with some fundamental choices they made which interfere with code clarity and auditability, so in addition to adding a few missing features (like memory regions and virtual memory) the entire access methodology is being reworked to be more friendly for our purposes, while making it easier to on-board systems programmers with backgrounds in C.

bunnie:
@xobs if you're OK then can @enjoy-digital merge this?

xobs:
Looks good to me.

enjoy-digital:
Thanks @bunnie,  @xobs, this is merged.

PR #656, opened 2020-09-29
Fix build issue where sdram_leveling is not found

kbeckmann:
4f76656 rewrote how sdram_leveling() was called, leading
to linking problems for targets with sdram but with
write leveling disabled, e.g. ulx3s.


enjoy-digital:
Thanks!

PR #657, opened 2020-09-30
tools/{litex_sim, litex_server}.py: Minor clean-up

jersey99:
This should allow users (hopefully not just me) to use SimSoC in a more modular way.


enjoy-digital (approved):
Thanks.

PR #662, opened 2020-10-05
Unconditionally ask dfu-util to "Issue USB Reset"

fidergo-stephane-gourichon:
Unconditionally ask dfu-util to "Issue USB Reset signalling once we're finished".

Some host machines need it.
Else host machine does not detect the USB in the design, lxterm fails, etc.
Issuing -R here fixes everything.

If issuing -R always does not cause any trouble to anyone, then this PR is relevant.


enjoy-digital:
Thanks this makes sense. I'll just do some testing before on boards that are relying on `dfu_util`.

enjoy-digital:
I've done some testing have haven't had issue with it, so think we can merge it. Thanks.

PR #663, opened 2020-10-05
Fix SoC CPU crash on minimal variants on call to flush_cpu_dcache().

fidergo-stephane-gourichon:
Generated soc.h says for example
#define CONFIG_CPU_TYPE_VEXRISCV
#define CONFIG_CPU_VARIANT_MINIMAL
#define CONFIG_CPU_HUMAN_NAME "VexRiscv_Min"

but code tested for CONFIG_CPU_VARIANT_MIN not MINIMAL.
Attempted to run instruction unknown to this CPU, most likely cause of hang.

This PR causes the code to correctly checks CONFIG_CPU_VARIANT_MINIMAL.

### DIFF
--- a/litex/soc/cores/cpu/vexriscv/system.h
+++ b/litex/soc/cores/cpu/vexriscv/system.h
@@ -11,7 +11,7 @@ extern "C" {
 
 __attribute__((unused)) static void flush_cpu_icache(void)
 {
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Good catch! Thanks.

PR #665, opened 2020-10-05
More precise memory performance test.

fidergo-stephane-gourichon:
On mem_speed, confirm parameters on terminal output.
Show speeds in bytes per second.

Rationale:
    
* Forcing megabytes per second for everyone does not make sense and clamps less-than-1MB/s to zero.
* Showing bytes per second allows to distinguish between low performance and a performance measurement bug.
* Anyway previous code claims speeds were in MiB/s, they were not, actually MB/s.

### DIFF
--- a/litex/soc/software/libbase/memtest.c
+++ b/litex/soc/software/libbase/memtest.c
@@ -185,7 +185,7 @@ void memspeed(unsigned int *addr, unsigned long size, bool read_only)
 	__attribute__((unused)) unsigned long data;
 	const unsigned int sz = sizeof(unsigned long);
 
-	printf("Memspeed at 0x%p...\n", addr);
+	printf("Memspeed at 0x%p size 0x%x (%u read or write operations), read_only? %s...\n", addr, size, (size/sz), read_only?"true":"false");
 
 	/* init timer */
 	timer0_en_write(0);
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, this has been merged, but i reverted some changes: https://github.com/enjoy-digital/litex/commit/0a80e4c3d6d368c7cf39c84e3734416b0a1e414b and added a function to show the speed with the appropriate unity with https://github.com/enjoy-digital/litex/commit/4f30a5b8e5804ba7ac45f3a7de9358f50bedb93e.

fidergo-stephane-gourichon:
> Thanks, this has been merged, but i reverted some changes: [0a80e4c](https://github.com/enjoy-digital/litex/commit/0a80e4c3d6d368c7cf39c84e3734416b0a1e414b) and added a function to show the speed with the appropriate unity with [4f30a5b](https://github.com/enjoy-digital/litex/commit/4f30a5b8e5804ba7ac45f3a7de9358f50bedb93e).

Thanks for the merge. Showing speed with units is indeed more readable whatever the performance level.

Regarding removing part of the added log for the test, the read-only vs read-write variant is available if the test succeeded, not if the thing crashed. The length of the test is not logged at all and can be a hint for a reason of a crash, hence I added it.

Thanks for all the hard work done before, happy to contribute.

enjoy-digital:
Thanks for the feedback, i just added back the test size with https://github.com/enjoy-digital/litex/commit/b904aa7d1815d20a7798eed83e6b2f72c6c03148. I'll try to add the read-only/write-only information in the next days.

PR #666, opened 2020-10-05
bios: add command returning card-detect pin status

gsomlo:
low priority, I just wanted to convince myself that this pin works in bare-metal, and figured now that I created the bios command, might as well share it. Feel free to ignore if it comes off as more "spammy" than useful :)


enjoy-digital:
Thanks @gsomlo, this can be useful yes.

Issue #667, opened 2020-10-06
Avoid changes of the verilog netlist prefix on unrelated changes.


enjoy-digital:
When minor changes are done to the design, the verilog netlist prefix can also change (`soc_`, `main_`, etc...) which is fine when constraints are directly provided by LiteX but causes issues when applying manual constraints to internal signals or when comparing changes.

This needs to be understood and fixed.

cc @bunnie.

bunnie:
I also have this problem, if I get a chance to solve it I'll post a commit to this thread, at the moment haven't had a chance to dig into it!

enjoy-digital:
I did an initial pass on this and pushed experimental fixes to:
  https://github.com/enjoy-digital/litex/tree/issue-667-stable-verilog-names

This currently contains two changes:
  - stabilize generated FIFO storage and CDC temporary signal names;
  - stabilize flat Verilog names for signals inside named submodules, so adding an unrelated sibling should no longer rename an existing signal from `a` to `main_a`/`soc_a`.

The branch is not merged yet since this touches the generic Verilog naming flow and needs more review/testing to avoid regressions. The goal is to reduce name churn while keeping the change scoped, especially for flat Verilog where internal constraints and diffs are most affected.

PR #671, opened 2020-10-08
WIP: suggest a method for handling constraints

bunnie:
Here is a potential way to resolve issue #667. It's not super elegant yet, but wanted some feedback on the general approach.

The idea here is to add another platform method called `add_signal_command()`, that would be used like this:

```Python
        self.platform.add_signal_command('create_generated_clock -name dna_cnt -source [get_nets {{ @ }}] -divide_by 2 [get_pins {{DNA_PORT/CLK}}]', self.info.dna.clk)
        self.platform.add_signal_command(
                    'set_false_path -rise_from [get_clocks @] -rise_to [get_clocks @] -through [get_nets @]',
                   ClockSignal('usb_12'), ClockSignal(), self.usb.debug_bridge.cmd)
```

Basically, it will take a generic template string, and replace every instance of `@` with positional arguments that are resolved from signal names by the signal resolver within Migen. 

It doesn't handle the case of resolving instance names, so if you needed to refer to a port on an instance it can't help you there, which may be a limitation depending on what you are trying to do.

But, wanted to see what people thought about this approach before trying to refine it any further.

### DIFF
--- a/litex/build/generic_platform.py
+++ b/litex/build/generic_platform.py
@@ -179,7 +179,6 @@ class ConstraintManager:
         self.available = list(io)
         self.matched = []
         self.platform_commands = []
-        self.signal_command = []
         self.connector_manager = ConnectorManager(connectors)
 
     def add_extension(self, io):
@@ -243,9 +242,6 @@ class ConstraintManager:
 
     def add_platform_command(self, command, **signals):
         self.platform_commands.append((command, signals))
-
-    def add_signal_command(self, command, *signals):
-        self.signal_command.append((command, signals))
 
     def get_io_signals(self):
         r = set()
@@ -327,9 +323,6 @@ class GenericPlatform:
     def add_platform_command(self, *args, **kwargs):
         return self.constraint_manager.add_platform_command(*args, **kwargs)
 
-    def add_signal_command(self, command, *signals):
-        return self.constraint_manager.add_signal_command(command, *signals)
-
     def add_extension(self, *args, **kwargs):
         return self.constraint_manager.add_extension(*args, **kwargs)
 
@@ -392,32 +385,6 @@ class GenericPlatform:
 
     def add_verilog_include_path(self, path):
         self.verilog_include_paths.append(os.path.abspath(path))
-
-    def resolve_signal_commands(self, vns):
-        xdc_output = '\n\n################################################################################\n#  Resolved signal XDCs\n################################################################################\n\n'
-        for command, sigs in self.constraint_manager.signal_command:
-            resolved_sigs = []
-            for signal in sigs:
-                rname = vns.get_name(signal)
-                if signal.nbits > 1:
-                    rname += '*'
-                resolved_sigs.append(rname)
-
-            frags = command.split('@')
-
-            xdc = ''
-            index = 0
-            for frag in frags:
-                xdc = xdc + frag
-                if index < len(resolved_sigs):
-                    xdc = xdc + resolved_sigs[index]
-                else:
-                    break
-                index = index + 1
-
-            xdc_output = xdc_output + xdc + '\n'
-
-        return xdc_output
 
     def resolve_signals(self, vns):
         # resolve signal names in constraints
--- a/litex/build/generic_platform.py
+++ b/litex/build/generic_platform.py
@@ -179,6 +179,7 @@ class ConstraintManager:
         self.available = list(io)
         self.matched = []
         self.platform_commands = []
+        self.signal_command = []
         self.connector_manager = ConnectorManager(connectors)
 
     def add_extension(self, io):
@@ -242,6 +243,9 @@ class ConstraintManager:
 
     def add_platform_command(self, command, **signals):
         self.platform_commands.append((command, signals))
+
+    def add_signal_command(self, command, *signals):
+        self.signal_command.append((command, signals))
 
     def get_io_signals(self):
         r = set()
@@ -323,6 +327,9 @@ class GenericPlatform:
     def add_platform_command(self, *args, **kwargs):
         return self.constraint_manager.add_platform_command(*args, **kwargs)
 
+    def add_signal_command(self, command, *signals):
+        return self.constraint_manager.add_signal_command(command, *signals)
+
     def add_extension(self, *args, **kwargs):
         return self.constraint_manager.add_extension(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


mithro:
Do you have examples other than `create_generated_clock` and `set_false_path`? 

If possible I think we want both of these examples should be automatically created / generated by Migen / LiteX?

bunnie:
Most generated clocks are automatically generated already, the need for a generated DNA clock is pretty unique to my use case.

False paths in this instance for the USB debug bridge command as you cross between domains could maybe be inferred but it's not always safe to assume it's a false path. For this path I had to double check the design that it's actually OK for the timing to be very sloppy on it. I think it might be neat to have a "false path" attribute maybe on signals so it's easier to deal with them in retimers, but this is orthogonal to this PR: if that's a desired feature I think a separate issue should be opened around that, this PR is around creating a generic method for creating complex Vivado constraints that can refer directly to Migen objects so the name used in the XDC can reflect what the finalizer eventually produces. 

As for making more generic constraints, the constraint guide for Vivado is like..a few hundred pages long. There's a lot it can do. Here's some other constraints that I've recently used.

```
            self.platform.add_platform_command("set_input_delay -clock spidqs -max 0.6 [get_ports {{spiflash_8x_dq[*]}}]")
            self.platform.add_platform_command("set_input_delay -clock spidqs -min -0.6 [get_ports {{spiflash_8x_dq[*]}}]")
            self.platform.add_platform_command("set_output_delay -clock [get_clocks spiclk_out] -max 1 [get_ports {{spiflash_8x_dq[*]}}]")
            self.platform.add_platform_command("set_output_delay -clock [get_clocks spiclk_out] -min -1 [get_ports {{spiflash_8x_dq[*]}}]")
        platform.add_platform_command("set_multicycle_path 2 -setup -start -from [get_clocks sys_clk] -to [get_clocks clk50] -through [get_cells microcode_reg*]")
        platform.add_platform_command("set_multicycle_path 1 -hold -from [get_clocks sys_clk] -to [get_clocks clk50] -through [get_cells microcode_reg*]")
```

These examples are done using manual specification of the instance names or net names, but I would like to convert to use migen `Signal` objects which would make the constraints not "drift" from the code as the blocks are maintained or re-used in other designs. 

Note that the `set_multicycle_path` command in particular is quite tricky, to do it right you need two commands to constrain both the setup and the hold side and it needs to be quite specific so you don't accidentally end up introducing a bunch of timing exceptions that break the design. I'm using this constraint a ton in my designs because I run the CPU core at 100MHz but the crypto cores run at speeds ranging from 50-200MHz depending upon the complexity and nature of the crypto algorithm, and I make aggressive use of data enables between clock domains that are not automatically inferrable, and thus manual constraint interventions are required.

bunnie:
So i went through and converted the entire betrusted-soc project to use this proposed form of signal constraints, to really push it through its paces. I think my WIP has mixed utility:

- (good) Using signal names is much more elegant than digging into verilog netlists and calling out names
- (bad) Vivado name mangling means which signal gets the final name is hard to predict, so you're once again in a guessing game

What I'm realizing is that maybe the more valuable thing to do is to constrain pins on instances, because Vivado tends not to mangle instance names, but net names definitely get mangled because Migen does so many redundant name changes it's very hard to predict which name is the final one that Vivado prefers to use. So for example, `get_pins soc_usb_debug_bridge_wishbone_adr*/D` seems to be more solid than `get_nets soc_usb_debug_bridge_wishbone_adr*`. The name of the net that actually gets connected to the `soc_usb_debug_bridge_wishbone_adr_reg/D` pins can be very unpredictable, but it seems the register intstance itself has a durable name that doesn't get mangled.

@enjoy-digital I would propose that for now, we merge this WIP, so that I can continue to use it while staying on LiteX mainline, thus allowing me to stay in sync on other features while this evolves. The function itself is pretty harmless, I don't think it will side-effect any of the current functionality. But, ideally, over the coming weeks as I use this more I will get a better sense of how to work with Vivado's name mangling and create maybe a cleaner PR for binding constraints to internal net names.

For what it's worth, Vivado never mangles ports or global net names, so probably whatever the final version of this would be good to handle those more gracefully as well, but for now that's not part of this PR.

mithro:
When doing optimization;
 * Registers only really change when re-timing is done. 
 * Combinational logic changes all the time.

Hence, the suggestion I have heard is to always put constraints on the registers pins....

enjoy-digital:
@bunnie: sorry for the delay and thanks for the PR. I however think you are working on a feature that already exists:

````
        sig1 = Signal()
        sig2 = Signal()
        self.comb += sig1.eq(sig2) # dummy assignement to avoid signals to be simplified
        platform.add_platform_command("dummy constraints between {sig1} and {sig2}", sig1=sig1, sig2=sig2)
````

On a Xilinx/Vivado platform will produces the following constraints in the `.xdc`:
````
dummy constraints between main_sig1 and main_sig2
````

Can you verify it covers your use case?

bunnie:
oh! i didn't even know this was a thing. I don't quite fully follow the example you sent, but I think there's enough hints there that I'll eventually figure it out. It sounds like this may actually cover the use case, especially if I can use it in a form that allows me to add the `_reg/D` pin specifier syntax at the end of the signal name, so i'm constraining registers and not signals.

bunnie:
After a couple tests, I think this can work! The only modification I might suggest is automatically adding a `*` after signals that have a width bigger than 1, but for now I will just manually add that into my constraints. 

I guess @enjoy-digital, do you have an example of a constraint that results in issue #667 that can't be resolved using the technique you just taught me? Because if you are using the derived names for nets after finalization, it shouldn't matter if the names change, but maybe there is a different technique or something I am missing that would be very helpful for me to know about.

bunnie:
I've unstaged the commits related to modifying the Litex core, but I'm leaving the PR open because I need some minor modifications to the S7OPI block to give me visibility into signals so that I may constraint them from the top level. I think it's OK to just accept the PR in this case and close it out. 

Thanks! Sorry ... I didn't realize you already had this cool feature.

enjoy-digital:
@bunnie: thanks for the feedback. I'll have a closer look at some of my designs for #667. Even if calling `add_platform_command` this way prevent the issue, it would still be good to avoid internal names to change too much when doing unrelated changes and we can probably improve this.

For the `spi_opi` changes,  what do you think of https://github.com/enjoy-digital/litex/pull/675? That's what I'm doing generally for to expose internal signals I need to observe or constraint. this avoid adding to much complexity to the code just for that.

bunnie:
I do agree that trying to understand why name mangling gets so crazy sometimes would be good. It's a little distressing that it changes from run to run with seemingly unrelated or minor changes, it makes me concerned for repeatability of builds, but it's no longer blocking my progress. 

Now that I know about this `"{net}", net=signal` idiom, I will close this PR, it's not actually needed. Thanks, and sorry to waste your time!

PR #674, opened 2020-10-13
build/radiant: Allow synthesis with Yosys

daveshah1:
This for testing YosysHQ/yosys#2397 

It's a bit of a fudge because it doesn't actually skip Radiant's synthesis - to do that would require use of the map, par, trce, bitgen commands individually without using Tcl. I can change to that approach if it is preferred, but probably for Linux only.


enjoy-digital:
Thanks @daveshah1, the approach is fine for now. We'll improve it as Yosys/LiteX's Radiant support improves.

PR #677, opened 2020-10-16
Add initial interrupt support for Microwatt in LiteX

madscientist159:
There is a conflict between the LiteX way of doing things and the POWER
way of handling interrupt tables.  LiteX expects to be able to put a ROM
at address 0 and load an application into RAM at a higher address; POWER
is architected to jump to exception handlers at 0x100...0x1000.

As a result of this, we have taken the approach of placing generic exception
handler entry / exit routines into ROM, and reserving a single pointer in
SRAM to determine the C ISR handler location.  If no application is loaded,
this pointer is set to the BIOS ROM ISR.  When an application loads, before
reenabling interrupts, it needs to set __rom_isr_address to the address of
the application's ISR, otherwise the BIOS ROM ISR will continue to be used.

Tested to operate with the built-in UART in IRQ mode, both in BIOS and in
loaded RAM application.

### DIFF
--- a/litex/soc/cores/cpu/microwatt/core.py
+++ b/litex/soc/cores/cpu/microwatt/core.py
@@ -3,6 +3,7 @@
 #
 # Copyright (c) 2019 Florent Kermarrec <florent@enjoy-digital.fr>
 # Copyright (c) 2019 Benjamin Herrenschmidt <benh@ozlabs.org>
+# Copyright (c) 2020 Raptor Engineering <sales@raptorengineering.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 import os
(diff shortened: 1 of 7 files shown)


enjoy-digital:
@madscientist159: Thanks a lot! This is merged.

enjoy-digital:
@madscientist159: it seems i'll need to update microwatt submodule (https://github.com/enjoy-digital/litex/blob/master/litex_setup.py#L55) can you provide the git sha1 you've been using to validate this work?

madscientist159:
@enjoy-digital I can put a pull request together with the internal patches we are carrying on Microwatt; will work on that over the weekend.  Thanks for the merge of the IRQ support!

madscientist159:
@enjoy-digital Sorry it took so long to get back to this...I looked in our repos and there were no significant changes to the Microwatt HDL, just a couple of parameter tweaks to get it to fit better on our development hardware..  '''f9807b''' should work just fine with the new IRQ support.

enjoy-digital:
Thanks @madscientist159, with https://github.com/enjoy-digital/litex/commit/b3a42d76cef429041a077dd9e9b57554e7763a0c i added `irq` variants (until we can make it the default) but i still have issue with the IRQ version and `f9807b`. I just created https://github.com/enjoy-digital/litex/issues/693 to discuss this, can you help me figuring out the integration issues?

PR #678, opened 2020-10-20
DFUProg now checks and propagate return codes of called executables...

fidergo-stephane-gourichon:
…so that failure gets noticed.

And adjust some caller scripts that happen to be in same repo.

### DIFF
--- a/litex/boards/targets/arty.py
+++ b/litex/boards/targets/arty.py
@@ -149,7 +149,7 @@ def main():
 
     if args.load:
         prog = soc.platform.create_programmer()
-        prog.load_bitstream(os.path.join(builder.gateware_dir, soc.build_name + ".bit"))
+        return prog.load_bitstream(os.path.join(builder.gateware_dir, soc.build_name + ".bit"))
 
 if __name__ == "__main__":
-    main()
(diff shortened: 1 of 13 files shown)


fidergo-stephane-gourichon:
For caller scripts that don't test for the return value of DFUProg.load_bitstream(), they won't benefit but at least this change is non-breaking.

enjoy-digital:
@fidergo-stephane-gourichon: indeed, that's also what i want to implement. Sorry for the delay, i haven't forgotten the PR but haven't been able to look at it yet.

enjoy-digital:
@fidergo-stephane-gourichon: just for info, i started added more checks and improving error reporting for the toolchain's calls with https://github.com/enjoy-digital/litex/commit/db836e8e5d9d66caade9294f1a8b67eb2b4d3611. I'll do the same on the programmers soon.

enjoy-digital:
Thanks @fidergo-stephane-gourichon for the feedback and proposed PR, with https://github.com/enjoy-digital/litex/commit/2741fc2ba52aa1b98f8c1ab76d69278555b4e1f6 we now no longer silently ignore the programming errors.

PR #680, opened 2020-10-21
radiant: Use {} string for bus port names

daveshah1:
Otherwise [] is parsed as a Tcl command resulting in an error


enjoy-digital:
Thanks!

PR #684, opened 2020-11-02
cores/cpu/zynq7000: fix axi hp slave registration

sergachev:
Zynq HP AXI ports are slaves, not masters - see the lines right above.


enjoy-digital:
Thanks! that's merged.

PR #688, opened 2020-11-07
Fix check for wrong named attributes

rprinz08:
- Fix for some wrong named attribute checks: "tx_socket" should be "socket"

- commUDP blocks in socket receive in case of unreliable Ethernet connection. This patch introduces a timeout and error handling in case timeout occurs


enjoy-digital:
Thanks @rprinz08, that's merged. That's possible I'll do revisit a bit the exception handling in the next days since I'm also doing some work on this, but I'll try to keep the mechanism in place.

PR #690, opened 2020-11-08
Add arbitrary command (eg. write) capability to SPI DOPI

bunnie:
This set of patches enables arbitrary commands to be issued to SPI DOPI while in DOPI mode.

Among other things, this enables writing to happen, with the assistance of high-level firmware (such as found in wishbone-util for USB-based firmware updates).

This also moves the constraints into the DOPI block, thus cleaning up the API so no constraints need to be specified on the top-level integration.

Current status is tested and working on the reference setup, but this block is tricky enough that there will probably be a couple follow-on patches in the days to come as we start to uncover more corner cases.


bunnie:
ah. interesting. It seems that my method of asking for a "platform" object instead of "pads" (so that I can attach the timing constraints) is not compatible with the test benching methodology.

What's your preference on how I should handle this? I could either:

- move the platform object back out and into the upper module
- modify the test bench to create a dummy platform object that will allow the test to pass

enjoy-digital:
Thanks @bunnie, for the timings constraints, can you add a `add_timing_constraints` method to `S7SPIOPI` as done for example on the DNA (https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/icap.py#L111) or ICAP (https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/icap.py#L111) modules? This will make a separation between logic and timing constraints and should solves the integration issue: pass `pads` to `__init__` (as it was before) and `platform` to `add_timing_constraints`.
`

bunnie:
Awesome. That's a clever solution and I'll need to do a pass back on my own gateware modules to adopt that methodology later, but I think this should solve the issue, I've committed the change here.

enjoy-digital:
Thanks for the update, i just merged it but moved the timing constraints to the end with https://github.com/enjoy-digital/litex/commit/0627c01d8947e8a91856f006408488b551ae521e: having such constraints at the top is probably a good way to prevent anyone from looking at the rest of the code :)

bunnie:
Hahah good point! thanks! Sorry about extra the CI build messages, I forgot to update the testbench to match the Litex methodology, it should be fixed now.

PR #692, opened 2020-11-09
Lattice NX PLL Support

davidcorrigan714:
Adding support for the PLL blocks in the Lattice NX chips. Most of the work was in understanding how the analog parameters are calculated based on the Lattice IP Generator source which is published under their reference design license. The algorithms were reimplemented and results compared against the reference designs.

There seems to be some timing constraint issues with clocks generated through the PLL block. There's a workaround in the PLL code which creates new clocks on the PLL output pins instead of referencing the input clock, but I'm hoping that either Lattice responds to my inquiries on the matter, or the next release of Radiant either fixes the issues or provides more documentation on the matter.

### DIFF
--- a/litex/soc/cores/clock_nxpll.py
+++ b/litex/soc/cores/clock_nxpll.py
@@ -1,0 +1,424 @@
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2020 David Corrigan <davidcorrigan714@gmail.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from collections import namedtuple
+import logging
+import math
+import pprint
+from math import log, log10, exp, pi
+from cmath import phase
+
+from migen import *
+
+from litex.soc.cores.clock import register_clkin_log, create_clkout_log, compute_config_log
+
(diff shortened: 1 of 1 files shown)


davidcorrigan714:
This PR should be accepted alongside [this PR](https://github.com/litex-hub/litex-boards/pull/121) as the boards will break if only one is accepted.

enjoy-digital:
Thanks a lot @davidcorrigan714, that's merged.

enjoy-digital:
@davidcorrigan714: just for info, to ease maintenance and future additions, i created a clock directory, moved the code to it and split it with: https://github.com/enjoy-digital/litex/commit/14e196ab5d97ad347343154a81d70c950f3f5c76 So your code has been moved to` litex.soc.cores.clock.lattice_nx` and I also re-integrated `NXOSCA` into it. `__init__.py` provides a compatibility layer, so the targets don't need to be updated expect the crosslink_nx_evn that has been updated with https://github.com/litex-hub/litex-boards/commit/afe44e2bd6e31159f091c2ae20247bc3352815be.

PR #696, opened 2020-11-12
[fix] support 12f devices with nextpnr

chmousset:
Nextpnr does not support the '12k' density of ECP5, since it considers it's physically the same devices as the '25k' ones. Thus, it is currently impossible to build designs for the 12k devices with trellis.

The only difference between 12k and 25k devices lays in the `idcode`. nextpnr must be passed ` --12k` and ecppack must be passed `--idcode 0x21111043`
Original thread where I found the information: https://gitter.im/ulx3s/Lobby?at=5c7284945df86a4568875a85

Tested with two ULX3s (the 12k and 85k variants), it produces a working bitstream in both cases.

It would probably would be 'cleaner' that this is sorted out in nextpnr/ecppack, but it won't hurt to do it here. It's probably much simpler aswell.


enjoy-digital:
Thanks for the PR @chmousset. Are you sure you are using an up to date version of Nextpnr? We are already supporting the 12F variant for the ULX3S (https://github.com/litex-hub/litex-boards/pull/120) and i just generated bitstreams for 12F and 25F and the idcode seems already inserted correctly:

image

enjoy-digital:
@chmousset: @daveshah1 just confirmed me that passing the `idcode` is no longer required for the `12F` with newer versions of `nextpnr`/`ecppack` so you just need to update your toolchain :)

daveshah1:
To confirm `--12k` to nextpnr should set the IDCODE already so this isn't needed. If this isn't working, please file an issue against nextpnr.

PR #699, opened 2020-11-15
Create EventManager option for documented bits

bunnie:
Currently there is no way to communicate bit masks for an interrupt register defined using EventManager() into the firmware (that I know of). 

This PR proposes to add an option to EventManager where if `document_events` is set to `True`, instead of handing `soc.svd` a single multi-bit field with no definition of what each bit means, it automatically generates sub-fields that are named. 

This will help downstream software adjust more easily to changes in interrupt APIs without having to rely upon some manual mapping of bits on event fields. 

The API is turned off by default because turning it on breaks existing firmware, but I'd like the new firmware I'm developing to rely on register fields instead of hard-coded bitmasks.

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -308,7 +308,7 @@ def get_csr_svd(soc, vendor="litex", name="soc", description=None):
         svd.append('                    <addressOffset>0x{:04x}</addressOffset>'.format(csr_address))
         svd.append('                    <resetValue>0x{:02x}</resetValue>'.format(csr.reset_value))
         svd.append('                    <size>{}</size>'.format(length))
-        svd.append('                    <access>{}</access>'.format(csr.access))
+        # svd.append('                    <access>{}</access>'.format(csr.access))  # 'access' is a lie: "read-only" registers can legitimately change state based on a write, and is in fact used to handle the "pending" field in events
         csr_address = csr_address + 4
         svd.append('                    <fields>')
         if hasattr(csr, "fields") and len(csr.fields) > 0:
--- a/litex/soc/interconnect/csr_eventmanager.py
+++ b/litex/soc/interconnect/csr_eventmanager.py
@@ -142,18 +142,22 @@ class EventManager(Module, AutoCSR):
         sources_u = [v for k, v in xdir(self, True) if isinstance(v, _EventSource)]
         sources = sorted(sources_u, key=lambda x: x.duid)
         n = len(sources)
-        self.status = CSR(n)
-        self.pending = CSR(n)
-        self.enable = CSRStorage(n)
+
+        fields = []
+        for i, source in enumerate(sources):
+            fields += [CSRField(source.name, size=1, description="Mask bit for {}".format(str(source.name)))]
+        self.status = CSRStatus(n, fields=fields)
+        self.pending = CSRStatus(n, fields=fields)
+        self.enable = CSRStorage(n, fields=fields)
 
         for i, source in enumerate(sources):
             self.comb += [
-                self.status.w[i].eq(source.status),
-                If(self.pending.re & self.pending.r[i], source.clear.eq(1)),
-                self.pending.w[i].eq(source.pending)
+                getattr(self.status.fields, source.name).eq(source.status),
+                getattr(self.pending.fields, source.name).eq(source.pending),
+                If(self.pending.re & getattr(self.pending.fields, source.name), source.clear.eq(1)),
             ]
 
-        irqs = [self.pending.w[i] & self.enable.storage[i] for i in range(n)]
+        irqs = [self.pending.status[i] & self.enable.storage[i] for i in range(n)]
         self.comb += self.irq.eq(reduce(or_, irqs))
 
     def __setattr__(self, name, value):
--- a/litex/soc/interconnect/csr_eventmanager.py
+++ b/litex/soc/interconnect/csr_eventmanager.py
@@ -135,8 +135,9 @@ class EventManager(Module, AutoCSR):
         asserted.
     """
 
-    def __init__(self):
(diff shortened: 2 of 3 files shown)


litex/soc/interconnect/csr_eventmanager.py
```diff
@@ -135,25 +135,116 @@ class EventManager(Module, AutoCSR):
...
+                if source.description == None:
+                    desc = "This register contains the current raw level of the {} event trigger.  Writes to this register have no effect.".format(str(source.name))
+                else:
+                    desc = source.description
+
+                if hasattr(source, "name") and source.name is not None:
+                    fields.append(CSRField(
+                        name=source.name,
+                        size=1,
+                        description="Level of the `{}` event".format(source.name)))
```
xobs, line 176: Shouldn't this have double-backticks here? i.e.

```
description="Level of the ``{}`` event".format(source.name)))
```

litex/soc/interconnect/csr_eventmanager.py
```diff
@@ -135,25 +135,116 @@ class EventManager(Module, AutoCSR):
...
+                else:
+                    fields.append(CSRField(
+                        name="event{}".format(i),
+                        size=1,
+                        description="Level of the `event{}` event".format(i)))
```
xobs, line 181: This probably also requires double-backticks, because ReStructured Text

litex/soc/interconnect/csr_eventmanager.py
```diff
@@ -135,25 +135,116 @@ class EventManager(Module, AutoCSR):
...
+            for i, source in enumerate(sources):
+                if source.description is None:
+                    desc = "This register enables the corresponding {} events.  Write a `0` to this register to disable individual events.".format(str(source.name))
+                else:
+                    desc = source.description
+                if hasattr(source, "name") and source.name is not None:
+                    fields.append(CSRField(
+                        name=source.name,
+                        offset=i,
+                        description="Write a `1` to enable the `{}` Event".format(source.name)))
```
xobs, line 215: Double-backticks

litex/soc/interconnect/csr_eventmanager.py
```diff
@@ -135,25 +135,116 @@ class EventManager(Module, AutoCSR):
...
+                else:
+                    fields.append(CSRField(
+                        name="event{}".format(i),
+                        offset=i,
+                        description="Write a `1` to enable the `{}` Event".format(i)))
```
xobs, line 220: Double-backticks

litex/soc/integration/export.py
```diff
@@ -308,7 +308,7 @@ def print_svd_register(csr, csr_address, description, length, svd):
         svd.append('                    <addressOffset>0x{:04x}</addressOffset>'.format(csr_address))
         svd.append('                    <resetValue>0x{:02x}</resetValue>'.format(csr.reset_value))
         svd.append('                    <size>{}</size>'.format(length))
-        svd.append('                    <access>{}</access>'.format(csr.access))
+        # svd.append('                    <access>{}</access>'.format(csr.access))  # 'access' is a lie: "read-only" registers can legitimately change state based on a write, and is in fact used to handle the "pending" field in events
```
xobs, line 311: This is a good catch. `read-only` doesn't mean the same thing in LiteX that it does in SVD. Omitting it is probably the best solution here.

What we really want is to be able to indicate that each Field's `modifiedWriteValues` is `oneToClear` within an interrupt pending register, however we don't currently have the ability to do that here.

xobs: Reference: https://www.keil.com/pack/doc/cmsis/SVD/html/elem_registers.html#elem_field

litex/soc/interconnect/csr_eventmanager.py
```diff
@@ -135,25 +135,116 @@ class EventManager(Module, AutoCSR):
...
+                        name="event{}".format(i),
+                        size=1,
+                        description=source_description(source)))
+            self.pending = CSRStatus(n, description=desc, fields=fields)
+
+            # annotate enable
+            fields = []
+            for i, source in enumerate(sources):
+                if source.description is None:
+                    desc = "This register enables the corresponding {} events.  Write a `0` to this register to disable individual events.".format(str(source.name))
```
xobs, line 208: ``

xobs (approved):

enjoy-digital:
Thanks @bunnie, that's indeed a good idea. I'm fine with the general idea and we could merge it once you take into account @xobs's suggestions. Once merged I'll play a bit with it and will try to remove some duplications.

mithro:
I would support this being the default on option with a clear documentation on how to disable it for people who want to go back to the old method.

bunnie:
Resolved all of xobs comments, i think...should be good for merge.

bunnie:
I also added an option to pass-through event documentation to the I2S IP block.

bunnie:
> I would support this being the default on option with a clear documentation on how to disable it for people who want to go back to the old method.

This was kind of the original goal of the changes but forcing the system over to this as the default caused a lot of regressions, and that was just in the Rust-based firmware builds. I think it could be fairly disruptive to a lot of code bases (although maybe C-based code bases are less sensitive to the types of register fields, so maybe they won't care as much?), so I'm more in favor of an incremental approach to phase it out over time.

enjoy-digital:
Thanks @bunnie, it's merged. I'll have a closer look and will eventually do changes or create another PR with the changes.

enjoy-digital:
@bunnie, @xobs: I did some simplifications/cleanups on the code with https://github.com/enjoy-digital/litex/pull/701 and also did the changes to be able to enable it by default (and so only keep this new mode). Can you review and test on your design and merge if this is good for you?

PR #700, opened 2020-11-17
Add SoC timer peripheral timer-uptime CLI parameter

lschuermann:
This allows enabling the uptime register in the timer core from the command line.

I've come to use the uptime feature quite frequently and having it available as a CLI parameter allows for easier CI integration.


enjoy-digital:
Thanks @lschuermann, that's merged.

PR #702, opened 2020-11-18
litex: soc: do not add the timestamp in the BIOS if it was disabled

acomodi:

The build time addition to the LiteX can be enabled with `with_build_time`.
The problem was that this functionality was not reflected in the BIOS source code, where the timestamp of the build time was added regardless of the `with_build_time` value.

This PR fixes this, allowing the user to disable the timestamp generation in the BIOS.

The main reason to do so is that the timestamp generates non-deterministic outputs, while in several cases, it is necessary to have stable and deterministic outputs, regardless of how small a difference is between two consecutive builds.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1061,6 +1061,7 @@ class LiteXSoC(SoC):
         self.check_if_exists(name)
         if with_build_time:
             identifier += " " + build_time()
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, it's merged. I just renamed the constant to `WITH_BUILD_TIME`.

PR #703, opened 2020-11-19
Add interrupt only if interrupts are supported

Disasm:
This check exists in other parts of the code, so I don't see any reasons to not have it here. This check prevents adding an irq entry (and generating interrupt entries in SVD) for cores that don't support interrupts.

### DIFF
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -192,7 +192,8 @@ class SoCCore(LiteXSoC):
     # Methods --------------------------------------------------------------------------------------
 
     def add_interrupt(self, interrupt_name, interrupt_id=None, use_loc_if_exists=False):
-        self.irq.add(interrupt_name, interrupt_id, use_loc_if_exists=use_loc_if_exists)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @Disasm, I'm looking at this, I'll integrate this soon, I just want to think a bit about it (to see if it's better to generate raise an error when interrupts are added to a SoC not supporting them or silently ignore this as you implemented).

enjoy-digital:
@Disasm: with https://github.com/enjoy-digital/litex/commit/146068b0485921aa0e57e7c24ccca55226c418cc and https://github.com/enjoy-digital/litex/commit/c8fcaaea2df5b6cd11cec4b5fb86eccf3d424bce it's no longer possible to add an interrupt to a SoC not supporting them. When IRQ support is optional on a peripheral (supported for better performance but the peripheral can also work in polling mode) the integration code of the peripheral should only add the IRQ if the SoC support it. Using this will also prevent adding an IRQ entry to SoCs that don't support IRQs, which was your initial goal. Is it fine for you?

Disasm:
Nice, thank you!

PR #704, opened 2020-11-23
symbiflow: remove workarounds

acomodi:

This PR fixes the symbiflow build and removes the various workarounds that were needed to have symbiflow working.


pgielda:
@enjoy-digital it would really help a lot of things if this was merged quickly :) (of course unless you see any issues with this commit)

enjoy-digital (approved):
Thanks, that's indeed a great step, nice work! I'll maybe revisit a bit the `--no-ident-version` addition later to enable it on all targets and simplify maintenance, but we the rest seems good so let's merge.

PR #705, opened 2020-11-23
correct the documentation for the ctrl reset register

bunnie:
@xobs pointed out to me today that the doc note on the CTRL reset register is incorrect.

As implemented, writing *anything* to the register triggers a reset as it is kicked off of the `.re` signal, and not the actual data signal itself.

Rather than change the behavior of the block to match the docs (which could cause compatibility issues elsewhere), this PR just fixes the doc string. Should be an easy review but wanted to clear it with @enjoy-digital instead of just merging it.


enjoy-digital:
Thanks @bunnie, merged.

PR #709, opened 2020-11-25
Add Yosys/nextpnr-nexus/oxide flow for CrossLink-NX

daveshah1:
Current caveats:

 - nextpnr-nexus isn't upstream yet, but will be in the coming weeks
 - it's all still experimental, occasionally more complex stuff routes slowly for a reason not yet identified
 - neither PLL nor LRAM are currently supported so the only working SoC is VIP+HyperRAM, the EVN SoC doesn't build yet, this will also be fixed in a week or two
 - prjoxide can't generate SVF files at the moment, only bitstreams, but I'm using ecpprog for programming which doesn't need this anyway
 - for the current ES devices, I've added a command line argument `--nexus-es-device` to enable the ES IDCODE - there isn't an equivalent for Radiant because Radiant never supports both, rather one or the other depending on the Radiant version installed. I've decided not to default to the ES IDCODE because that would be obsolete come Feb/March

If you'd rather hold off merging this until things are a bit more stable, that's not a problem for me.


enjoy-digital:
Thanks @daveshah1! I think this is fine merging it now since the changes are only related to the new oxide build backend; this will simplify testing for people wanting to try it and could also provide you more feedback.

PR #711, opened 2020-11-28
zynq7000: add tcl to create zynq IP based on board preset and custom configuration

trabucayre:
`zynq7000` IP can be added using `xci` file (current solution), using a block design or by using `create_ip` (this PR).
With this way a preset, and/or list of custom CONFIG.xxx, is used to configure zynq IP. No more needs to store somewhere an `xci` file.

Board's preset are located in vivado directory (shipped with this tool for zedboard, zc706,... or manually installed for zybo). This file is not mandatory (and not available for redpitaya), the configuration can be done using a dict provided at the platform level (with the drawback to add many lines in this file).

To resume for using this approach:
- an variable called `ps7_preset` may be added at platform and provides preset name in `vivado`
- a dict pointed by `ps7_config` again at the platform level may be added to provides all configuration parameters or just some  parameters to modify default configuration.

xci approach is left for backward compatibility 
Tested with zedboard

### DIFF
--- a/litex/soc/cores/cpu/zynq7000/core.py
+++ b/litex/soc/cores/cpu/zynq7000/core.py
@@ -143,6 +143,27 @@ class Zynq7000(CPU):
         self.ps7_xci = ps7_xci
         self.platform.add_ip(ps7_xci)
 
+    def set_ps7_config(self):
+        ps7_name = f"{self.platform.name}_ps7"
+        tcl = []
+        tcl.append(f"set ps7 [create_ip -vendor xilinx.com -name processing_system7 -module_name {ps7_name}]")
+
+        if hasattr(self.platform, "ps7_preset") and self.platform.ps7_preset:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre! This indeed gives more flexibility to configure the PS7. It's merged and I'm currently doing some changes on the integration. I'll post more info soon.

enjoy-digital:
@trabucayre: can you verify that with https://github.com/enjoy-digital/litex/commit/18f66a79f2cb8e2e9d54d1eade7d7a331560c593 you still can do what you initially wanted to support?

For your case, using the preset can be done with:
self.cpu.set_ps7(preset="preset_name")
And then pass optional additional config with:
self.cpu.add_ps7_config({"param0": "0", "param1": "1"})

trabucayre:
Looks good for me.
Not built the bitstream but checked tcl's content.
Now I need to find how to limit size for the redpitaya config dict.
Thanks

PR #713, opened 2020-11-28
rocket: Fix UB due to optimised away DFFs

daveshah1:
As both clock and async reset for the debug DFFs were 0, and there was no initial value on them, they were being validly optimised away by newer Yosys versions to 1'bx which was propagating into and breaking the core.

This fixes the problem by tying the async resets to the CPU reset signal.

Thus far only tested with my hacky post-early-synth simulation test, I'll have hardware results in half an hour or so -- edit, seems good on hardware too.


gsomlo:
@daveshah1 Thanks for tracking this down -- it does indeed fix the issue!

enjoy-digital:
Thanks a lot @daveshah1 for looking at this.

PR #714, opened 2020-11-29
Use asynchronous transfers when loading kernel

davidlattimore:
I observed that loading a 427KB kernel took about 2 minutes, so I
switched to using 4Mbit serial and it still took about 2 minutes. This
change fixes that by allowing load requests to continue while awaiting
confirmation that the CRC was valid and the load succeeded.

Prior to this change, with serial speed set to 4Mbits and CRC checking
enabled, a 427KB file would take 116 seconds to load.

With this change, it now takes 1.1 seconds.

I tested by randomly changing, deleting or inserting a byte of the
output stream.

I've attempted to keep the communication protocol backwards compatible.
So old versions of litex_term should still work with new BIOS code and
vice versa.

I also increased the size of the payload size in litex_term to match the
buffer size that already existed in the BIOS.

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -143,7 +143,23 @@ static uint32_t get_uint32(unsigned char* data)
 			(uint32_t) data[3];
 }
 
+static void uart_write_uint32(uint32_t value)
+{
+	uart_write(value >> 24);
+	uart_write((value >> 16) & 0xff);
+	uart_write((value >> 8) & 0xff);
+	uart_write(value & 0xff);
+}
+
 #define MAX_FAILED 5
+
+static void read_frame_header(struct sfl_frame* frame_out)
+{
+	frame_out->payload_length = uart_read();
+	frame_out->crc[0] = uart_read();
+	frame_out->crc[1] = uart_read();
+	frame_out->cmd = uart_read();
+}
 
 /* Returns 1 if other boot methods should be tried */
 int serialboot(void)
@@ -153,6 +169,7 @@ int serialboot(void)
 	static const char str[SFL_MAGIC_LEN+1] = SFL_MAGIC_REQ;
 	const char *c;
 	int ack_status;
+	int use_v2_resync_protocol = 0;
 
 	printf("Booting from serial...\n");
 	printf("Press Q or ESC to abort boot completely.\n");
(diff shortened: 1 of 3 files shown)


mithro:
(Just as an FYI, if your board has Ethernet, it is **highly** recommended to get tftp booting of your kernel going.)

enjoy-digital:
Thanks @davidlattimore! The aim of `--no-crc` args in **litex_term** was also to provide a simple solution to speed up transfer by disabling the CRC and has been heavily used when testing [Linux-on-LiteX-VexRiscv](https://github.com/litex-hub/linux-on-litex-vexriscvl) without troubles. Were you aware of it?

UART is generally used for small/medium binaries (a few KB) and aims to be simple. For more (up to a few MB) the `--no-crc` argument was doing the trick when no other interface was available to load the binaries, but for development, I would recommend Ethernet (and for deployed devices SPI Flash/SDCard/SATA) you can find more information here: https://github.com/enjoy-digital/litex/wiki/Load-Application-Code-To-CPU

I'm aware your solution is better (and thanks a lot for implementing it), but it also adds complexity which I'm currently trying to minimize in LiteX to ease maintenance, so I think we should consider integrating your solution only if the current one (`--no-crc`) is causing issues for you: because it does not work reliably or because you have a real need for CRC check.

davidlattimore:
Thanks for the response @enjoy-digital. You're right that this PR does add significant complexity.

My concern with --no-crc is that, not surprisingly, no CRC checking is done. I'd be worried that if a transmission error were to occur, my program would be silently corrupted and I might waste time debugging. As it says on the page you linked to "For large binaries, it's possible boost upload speed (and live dangerously :)) by disabling the CRC check with --no-crc". I'm not sure that I want to live dangerously :)

Ethernet might be an option for me some of the time. One of the boards I'm working with already has ethernet, however others don't.

A much simpler option than what I did in this PR would be to transfer the whole kernel without CRC checking, then do a CRC check of the whole memory range in one go. That wouldn't allow recovery in the case of transfer errors, but at least you'd know that corruption occurred and could retry. Do you think something like that might be acceptable?

enjoy-digital:
@davidlattimore: thanks, I can understand your concern. I implemented `--no-crc` as a quick workaround and kept it since haven't had troubles with it. But I remember wanting to implement a solution similar to your allowing a number of outstanding `SFL_CMD_LOAD`.

So here there are 3 things:
1) Speed.
2) CRC check.
3) Recovery in case of CRC error.

With the default behavior, we are checking CRC and capable of recovering on errors but the speed is very slow since for each `SFL_CMD_LOAD ` we are waiting for a `SFL_ACK_XY`, so have to pay for the serial round-trip delay.

With your solution, you are allowing outstanding `SFL_CMD_LOAD` for speed but are also introducing a mechanism to also recover in this case and I think the latter introduce most of the complexity. 

Maybe in a first time, we could just allow a number of outstanding `SFL_CMD_LOAD`? `litex_term.py`would then just allow a maximum of outstanding `SFL_CMD_LOAD` by counting the returned `SFL_ACK_XY` (and would only wait on it when the maximum is reached). We could then keep the CRC check and just generate an error in `litex_term.py`  when a `SFL_ACK_CRCERROR` is returned. So we keep speed and CRC check but loose recovery (and I'm not sure recovery been that useful in the past). This would also allow getting rid of the `--no-crc` mode, Would that be fine for you in a first time? (We could see to add recovery later if we think it's useful). Do you want to look at it or should I look at it?

davidlattimore:
@enjoy-digital that sounds like a good approach to me. I like that it can be done without changing the protocol (and the C code). I assume we should leave no-crc support in the BIOS in case someone is using an old version of lxterm? Or does that not matter so much?

I should hopefully be able to have a go at that tomorrow.

What should be the behavior in case of a CRC error? Should lxterm print an error then exit with a non-zero status? Currently failures in `upload` abort the upload, but then drops you to the lxterm shell, which can result in any unhandled responses being printed to the terminal. Easiest option would be to just sys.exit directly from `upload`, unless you'd prefer I propagate the error up through each function call? Or I could read all remaining serial data before dropping to the lxterm shell?

davidlattimore:
I've created a separate PR - #719

PR #719, opened 2020-12-03
lxterm: Speed up CRC checked uploads

davidlattimore:
- Don't wait for responses when loading
- Deprecate --no-crc flag
- Increase buffer size to match BIOS

### DIFF
--- a/litex/tools/litex_term.py
+++ b/litex/tools/litex_term.py
@@ -203,6 +203,19 @@ class LiteXTerm:
                 retry = 0
         return 1
 
+    def receive_upload_response(self):
+        # If we're not doing CRC checking, then there's no response to read.
+        if self.no_crc:
+            return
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @davidlattimore! That's seem exactly what I had in mind when discussing in https://github.com/enjoy-digital/litex/pull/714. I'm going to do some tests with it.

enjoy-digital:
I've been able to test with various boards supported in https://github.com/litex-hub/linux-on-litex-vexriscv and it's working fine: Thanks! With the USB-FIFO, it's in fact not possible currently to use a payload_size of 255, that's why it was reduced to 64, but I should have mentioned it somewhere. So I'll reduce it for now and will add a comment and fill an issue to look at this later. I'll also remove the `--no-crc` handling code from the BIOS since no longer useful.

PR #721, opened 2020-12-03
clock/lattice_nx: Set PLLRESET_ENA parameter

daveshah1:
If this parameter isn't set to ENABLED; then the PLLRESET signal is ignored.


enjoy-digital:
Thanks! Merged.

PR #724, opened 2020-12-05
soc/interconnect/axi: let connect_to_pads() be used by AXIInterface too

sergachev:
- connect_to_pads() used to be just in AXILiteInterface, but it can be used by AXIInterface just as well, so I moved it out of the AXILiteInterface class
 - removed a bit of code duplication with getattr(bus, channel)


enjoy-digital:
Thanks @sergachev!

PR #725, opened 2020-12-07
Csrfield generate convenience functions

fidergo-stephane-gourichon:
# Examples of use

## Changing one CSRField in a hardware register

Before:

```c
// This example is relatively simple because there is only one bit to stuff.
// When stuffing several bits, boolean operations are actually more complicated.
uint32_t value = rgb_ctrl_read();
value &= ~( 1 << CSR_RGB_CTRL_RRAW_OFFSET );
if (condition) value |= ( 1 << CSR_RGB_CTRL_RRAW_OFFSET );
rgb_ctrl_write(1 << CSR_RGB_CTRL_RRAW_OFFSET);
```

This PR generates `*_read` and `*_write` functions that allow code much simpler, readable, safe!

After:

```c
rgb_ctrl_rraw_write(condition); // this reads whole register, modify relevant bits, writes register
```

## Changing several fields, with only one register read and one register write

```c
        uint32_t value = myhw_conf_read();
        uint32_t someint = myhw_conf_somefield_extract( value );
        uint32_t someotherint = myhw_conf_someotherfield_extract( value );
        // perform processing, compute new someint, someotherint (or not)
        value = myhw_conf_somefield_replace( value, someint );
        value = myhw_conf_someotherfiels_replace( value, someotherint );
        myhw_conf_write( value ); // Existing function, just write whole register.
```

# What this PR changes

Additionally generate convenience functions to access CSRFields in C.

When interested in only one CSRField, makes shorter and more readable code:
* read CSRField value (read full register then extract)
* write CSRField value (read full register, replace, write full register)

When you wish to make only one register read, extract several values, and/or prepare a full register word with a number of values to be written in one write:
* extract CSRField integer value from a previous full register read, without doing another bus read
* replace CSRField integer value into a full register word, without doing yet another bus write, to be eventually written

Notice that generated code makes shifts and masks fully visible instead of showing final values. This is low level code. We think this style allows the developer to understand what is done and do shortcuts if they want. C compiler will generate exactly the same code anyway. If you prefer fully precomputed values, say so.

# Possible wished modifications

Tell me:

* Make a separate function like _get_rw_functions_c
* Replace `string + number + string` with `string.format(number)`

*Squash merge suggested*


enjoy-digital:
Thanks @fidergo-stephane-gourichon! This will be useful and looks good. Would using `set/get` instead of `inject/extract` be fine for you?  If so I could do the change.

fidergo-stephane-gourichon:
As discussed on IRC, it seems like things settled for 'extract'/'replace', so I updated the code and text above.

enjoy-digital:
Thanks @fidergo-stephane-gourichon for the discussion on IRC and PR. I'll merge it tomorrow.

enjoy-digital:
Thanks @fidergo-stephane-gourichon, this has been merged.

PR #727, opened 2020-12-10
fix a timing error in the S7 OPI block

bunnie:
should have no impact on normal operation, the path is
only for registering addresses that are correlated with
ECC errors as reported by the OPI device.


bunnie:
this is a very minor PR so I'm just merging it in

PR #737, opened 2020-12-19
mor1kx: Enable rotate, fpu and sign extensions under linux

stffrdhrn:
My thought is that if we are running linux the FPGA should be able to
handle these extra instruction's footprint.  Also, since we are running
on linux there may be any kind of software running on the CPU, so allow
handling these instructions.

But really, I am running GLIBC tests and they run faster with this
enabled.

### DIFF
--- a/litex/soc/cores/cpu/mor1kx/core.py
+++ b/litex/soc/cores/cpu/mor1kx/core.py
@@ -50,6 +50,7 @@ class MOR1KX(CPU):
     def gcc_flags(self):
         flags =  "-mhard-mul "
         flags += "-mhard-div "
+        flags += "-mcmov "
         flags += "-D__mor1kx__ "
 
         if "linux" in self.variant:
--- a/litex/soc/cores/cpu/mor1kx/core.py
+++ b/litex/soc/cores/cpu/mor1kx/core.py
@@ -15,7 +15,7 @@ from litex import get_data_mod
 from litex.soc.interconnect import wishbone
 from litex.soc.cores.cpu import CPU
 
-CPU_VARIANTS = ["standard", "linux"]
+CPU_VARIANTS = ["standard", "standard+fpu", "linux", "linux+fpu"]
 
 
 class MOR1KX(CPU):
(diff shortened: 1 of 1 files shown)


mithro:
@stffrdhrn - I agree that on a Linux configuration rotate + sign extensions make 100% sense (which should only be small increases in resource usage). The FPU however is normally a pretty big increase in resources.

My suggestion is kind of described at https://github.com/timvideos/litex-buildenv/wiki/Soft-CPU#soft-cpu-extensions -- A CPU configuration of `linux` shouldn't include the FPU, but it should be very easy to enable with a CPU configuration of `linux+fpu` .

stffrdhrn:
@mithro , Alright, that is actually what I wanted.  I figured FPU would be too big of a resource increase but didn't see any parameters to allow for configuration, so I added it for feedback. Will fix.

mithro:
@stffrdhrn - I think these changes are correct for the CPU core configuration, however we still need to update the gcc command line to enable these, right?

mithro:
See https://github.com/enjoy-digital/litex/blob/57f11c1181a8ed9a0df35b1eb4e34a852ee105b2/litex/soc/cores/cpu/mor1kx/core.py#L49-L63

stffrdhrn:
> @stffrdhrn - I think these changes are correct for the CPU core configuration, however we still need to update the gcc command line to enable these, right?

@mithro , yes we can do that.  I didn't do that as differnt GCC versions have different flags available and I have no idea what clang has.  I can do gcc and some basics for -mror and -msext.

mithro:
@stffrdhrn -- I think only supporting the GCC >9.1 should be reasonable as we provide mor1kx compiler versions at https://github.com/litex-hub/litex-conda-compilers ?

mithro (approved):
LGTM!

enjoy-digital:
@mithro: also LGTM, feel free to merge. Thanks @stffrdhrn.

stffrdhrn:
I had to make one change to this patch, the `self.mem_map = self.mem_map_linux` was not getting setup when using `linux+fpu` due to using `variant == "linux"`, I changed to use `in`.

PR #745, opened 2020-12-26
RFC dts: Support generating interrupt config

stffrdhrn:
Having interrupts defined is better than falling back to polling.  However, I am not sure of any compatibility issues with this patch.

### DIFF
--- a/litex/tools/litex_json2dts.py
+++ b/litex/tools/litex_json2dts.py
@@ -28,6 +28,7 @@ def generate_dts(d):
 / {
         #address-cells = <1>;
         #size-cells    = <1>;
+        interrupt-parent = <&intc0>;
 
 """
 
@@ -180,9 +181,12 @@ def generate_dts(d):
                 device_type = "serial";
                 compatible = "litex,liteuart";
                 reg = <0x{uart_csr_base:x} 0x100>;
+                interrupts = <{uart_interrupt}>;
                 status = "okay";
             }};
-""".format(uart_csr_base=d["csr_bases"]["uart"])
+""".format(
+    uart_csr_base = d["csr_bases"]["uart"],
+    uart_interrupt = d["constants"]["uart_interrupt"])
 
     # Ethernet -------------------------------------------------------------------------------------
 
@@ -195,13 +199,15 @@ def generate_dts(d):
                       <0x{ethmac_mem_base:x} 0x2000>;
                 tx-fifo-depth = <{ethmac_tx_slots}>;
                 rx-fifo-depth = <{ethmac_rx_slots}>;
+                interrupts = <{ethmac_interrupt}>;
             }};
 """.format(
     ethphy_csr_base = d["csr_bases"]["ethphy"],
     ethmac_csr_base = d["csr_bases"]["ethmac"],
     ethmac_mem_base = d["memories"]["ethmac"]["base"],
     ethmac_tx_slots = d["constants"]["ethmac_tx_slots"],
-    ethmac_rx_slots = d["constants"]["ethmac_rx_slots"])
+    ethmac_rx_slots = d["constants"]["ethmac_rx_slots"],
+    ethmac_interrupt = d["constants"]["ethmac_interrupt"])
 
    # SPI Flash -------------------------------------------------------------------------------------
 


stffrdhrn:
Note, I build and boot this with my litex scripts here: https://github.com/stffrdhrn/or1k-utils/tree/master/litex

The `litex.build` contains

```
  $HOME/work/litex/tools/litex_json2dts.py litex-mor1kx-csr.json | \
    dtc > $DIR/tftpd/litex-mor1kx.dtb
```

And `tftpd/boot.json` contains:

```
{
  "boot.bin":         "0x00000000",
  "litex-mor1kx.dtb": "0x06ff0000",
  "rootfs.cpio.gz":   "0x07000000",
  "bootargs": {
    "r1":             "0x06ff0000",
    "r2":             "0x00000000",
    "r3":             "0x00000000",
    "addr":           "0x00000000"
  }
```

Actually boot hangs on:

```
...
[    1.370000] Block layer SCSI generic (bsg) driver version 0.4 loaded (major 253)
[    1.370000] io scheduler mq-deadline registered
[    1.370000] io scheduler kyber registered
[    2.930000] e0002000.serial: ttyLXU0 at MMIO 0x0 (irq = 0, base_baud = 0) is a liteuart
[    3.330000] printk: console [liteuart0] enabled
[    3.380000] libphy: Fixed MDIO Bus: probed
[    3.380000] liteeth e0004800.mac: IRQ index 0 not found
[    3.390000] liteeth e0004800.mac: Failed to get IRQ, using polling
[    3.410000] liteeth e0004800.mac eth0: irq 0, mapped at bc004800
[    3.430000] NET: Registered protocol family 17
[    3.440000] Key type dns_resolver registered
[    3.450000] Waiting for root device /dev/ram0...
```

So, there are still issues here.
  - `/dev/ram0` is not configured correctly in my kernel and I don't think I need it
  - It complains `IRQ index 0 not found` but it is configured `interrupts = <2>` maybe a trivial issue.

stffrdhrn:
OK, so adding `interrupt-parent = <&intc0>;` has fixed the issue with booting.  The other issue is that the `mor1kx` generated dts config assumes a 8mb rootfs, mine is a lot bigger so I manually amended the initrd-end address and everything just works now.

I think having an argument to override the default rootfs/kernel sizes would be helpful.

mithro:
FYI - @shenki

litex/tools/litex_json2dts.py
```diff
@@ -133,6 +133,7 @@ def generate_dts(d):
             bus-frequency  = <{sys_clk_freq}>;
             compatible = "simple-bus";
             ranges;
+            interrupt-parent = <&intc0>;
```
shenki, line 136: The interrupt controller is global for the soc in both of our supported systems, so it should go at the top level. This way it is inherited by all nodes below it, ie, all of them.

shenki (changes requested):

litex/tools/litex_json2dts.py
stffrdhrn: Alright, I can move it.

stffrdhrn: > Alright, I can move it.

Done. And tested.

shenki (approved):

enjoy-digital:
Thanks @stffrdhrn. The changes look good. When testing this with VexRiscv-SMP (Linux-on-LiteX-VexRiscv) the IRQ seem to be mapped correctly:

````
[    2.713435] libphy: Fixed MDIO Bus: probed
[    2.726843] liteeth f0004800.mac eth0: irq 2, mapped at a040b800
[    2.731663] i2c /dev entries driver
````

but ethernet is not working (not able to ping another machine after configuring eth0 while it's the case in polling mode). This could however be specific to the VexRiscv-SMP SoC. Can you confirm ethernet was working with mor1kx with the IRQ enabled?

**Comment by stffrdhrn** [comment]
It is working fine on mor1kx:

```
[   29.620000] liteeth e0004800.mac eth0: irq 2, mapped at bc004800
```

```
# ping 10.0.0.27
PING 10.0.0.27 (10.0.0.27): 56 data bytes
64 bytes from 10.0.0.27: seq=0 ttl=64 time=3.804 ms
64 bytes from 10.0.0.27: seq=1 ttl=64 time=2.465 ms
^C
--- 10.0.0.27 ping statistics ---
2 packets transmitted, 2 packets received, 0% packet loss
round-trip min/avg/max = 2.465/3.134/3.804 ms
# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast qlen 1000
    link/ether de:9b:8a:38:7f:6c brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.5/24 scope global eth0
       valid_lft forever preferred_lft forever
```

enjoy-digital:
Thanks @stffrdhrn for the feedback. @gsomlo also confirmed ethernet was working fine on Rocket with the IRQ, so the issue is specific to VexRiscv-SMP. I'm going to merge the PR and will add a command line args to litex_json2dts.py to allow forcing polling mode. This will be useful for current VexRiscv-SMP and to do tests in polling mode in the future.

enjoy-digital:
With https://github.com/enjoy-digital/litex/commit/bf32d23d9a5f6e88a5e7e226b54e0414cf6f1eec is now also possible to switch to old polling mode using `--polling` args. This will be useful for VexRiscv-SMP until we figure out the issue with IRQs and could also ease bringing up new architectures.

Issue #748, opened 2020-12-28
litex_json2dts.py: Has hard coded initrd settings that are too small


stffrdhrn:
The settings for initrd are hard coded in the dts generator script as below, this gives only 8mb for the OpenRISC linux rootfs, which is too small.  It would be nice if we could pass sizes in via the command like. For example:  `--initrd-start 0x00800000 --initrd-size 1234567`, the values hard coded now in the script can be used as fall back defaults.

https://github.com/enjoy-digital/litex/blob/bddb170650242e3f993b882806474ca3480a6a7c/litex/tools/litex_json2dts.py#L34-L42

stffrdhrn:
@mateusz-holenko

mateusz-holenko:
@stffrdhrn Sounds like a good idea to me, added some comments to #749.

enjoy-digital:
Thanks @stffrdhrn, @mateusz-holenko for looking at this. The dicts were just added to avoid breaking mork1x support  when merging VexRiscv SMP, being able to configure this while keeping current default values also sounds like a good idea to me.

enjoy-digital:
https://github.com/enjoy-digital/litex/pull/749 has been merged, thanks @stffrdhrn, @mateusz-holenko.

PR #749, opened 2020-12-28
dts: Allow specifying initrd position and size via args

stffrdhrn:
This is needed as my initrd size is much larger than the default 8mb.
Also, sometimes the kernel I build is also large, so its good to be
able to move the initrd starting position.

Issue #748

### DIFF
--- a/litex/tools/litex_json2dts.py
+++ b/litex/tools/litex_json2dts.py
@@ -12,7 +12,7 @@ import json
 import argparse
 
 
-def generate_dts(d):
+def generate_dts(d, initrd_start_offset=None, initrd_size=None):
 
     kB = 1024
     mB = kB*1024
@@ -32,14 +32,16 @@ def generate_dts(d):
 """
 
     # Boot Arguments -------------------------------------------------------------------------------
-    linux_initrd_start_offset = {
+    default_initrd_start_offset = {
         "mor1kx":               8*mB,
         "vexriscv smp-linux" : 16*mB,
     }
-    linux_initrd_end_offset = {
-        "mor1kx":              16*mB,
-        "vexriscv smp-linux" : 24*mB,
-    }
+
+    if initrd_start_offset is None:
+        initrd_start_offset = default_initrd_start_offset[cpu_name]
+
+    if initrd_size is None:
+        initrd_size = 8*mB
 
     dts += """
         chosen {{
@@ -51,8 +53,8 @@ def generate_dts(d):
     main_ram_base      = d["memories"]["main_ram"]["base"],
     main_ram_size      = d["memories"]["main_ram"]["size"],
     main_ram_size_mb   = d["memories"]["main_ram"]["size"] // mB,
-    linux_initrd_start = d["memories"]["main_ram"]["base"] + linux_initrd_start_offset[cpu_name],
-    linux_initrd_end   = d["memories"]["main_ram"]["base"] + linux_initrd_end_offset[cpu_name])
+    linux_initrd_start = d["memories"]["main_ram"]["base"] + initrd_start_offset,
+    linux_initrd_end   = d["memories"]["main_ram"]["base"] + initrd_start_offset + initrd_size)
 
     # CPU ------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


stffrdhrn:
@mateusz-holenko @shenki

litex/tools/litex_json2dts.py
```diff
@@ -542,11 +549,13 @@ def add_clkout(clkout_nr, clk_f, clk_p, clk_dn, clk_dd, clk_margin, clk_margin_e
...
     parser = argparse.ArgumentParser(description="LiteX's CSR JSON to Linux DTS generator")
     parser.add_argument("csr_json", help="CSR JSON file")
+    parser.add_argument("--initrd-start", type=int, help="The location the initrd will be loaded, used to define linux,initrd-start.")
+    parser.add_argument("--initrd-size", type=int, help="The size of the initrd, used to calculate linux,initrd-end.")
     args = parser.parse_args()
 
     d = json.load(open(args.csr_json))
 
-    print(generate_dts(d))
+    print(generate_dts(d, getattr(args, "initrd_start"), getattr(args, "initrd_size")))
```
mateusz-holenko, line 558: Shouldn't `getattr(args, {"initrd_start", "initrd_size"}, None)` be used here to allow for automatic default initrd start/end handling in `generate_dts()`?

litex/tools/litex_json2dts.py
```diff
@@ -12,7 +12,7 @@
 import argparse
 
 
-def generate_dts(d):
+def generate_dts(d, initrd_start_offset, initrd_size):
```
mateusz-holenko, line 15: Maybe set `None` as default values for both `initrd_start_offset` and `initrd_size` to make this function backwards-compatible?

litex/tools/litex_json2dts.py
```diff
@@ -32,14 +32,21 @@ def generate_dts(d):
...
         "mor1kx":              16*mB,
         "vexriscv smp-linux" : 24*mB,
     }
+    if initrd_start_offset is None:
+        initrd_start_offset = default_initrd_start_offset[cpu_name]
+
+    if initrd_size is None:
+        initrd_end_offset = default_initrd_end_offset[cpu_name]
+    else:
+        initrd_end_offset = initrd_start_offset + initrd_size
```
mateusz-holenko, line 49: This is a low-prio, but we are mixing `size` and `end_offset` which might lead to some confusion. 

Wouldn't it be better to change `default_initrd_end_offset` to `default_initrd_size`?
In fact from what I see `default_initrd_size` seems to be 8mB for both cpus, so maybe a dictionary is not needed at all?

stffrdhrn: Yeah, I was just trying to keep the structure the same.  I will change it.

litex/tools/litex_json2dts.py
stffrdhrn: Sure.

litex/tools/litex_json2dts.py
stffrdhrn: The default value of `None` is already there in the args dict, it seems to be working that way.  Maybe I am missing something?

litex/tools/litex_json2dts.py
stffrdhrn: Fix this.

litex/tools/litex_json2dts.py
stffrdhrn: Fix, this too, removing the second dictionary make it much cleaner now.

stffrdhrn:
I pushed an update to document the default values in `add_argument` help text, also removed the `.` from the end of the help text as it seems we don't usually do that.

enjoy-digital:
Thanks @stffrdhrn, @mateusz-holenko, this looks good. I'm merging it now to avoid conflicts with the other pending changes and changes I need to do on my side.

litex/tools/litex_json2dts.py
mateusz-holenko: Oh, you are right. I was convinced that if you don't provide `initrd_start` argument in the command line there will be no property in `args` at all and `getattr` will throw an exception. This is not the case, however, as it turns out to be `None` in such case (which plays nicely with the rest of the logic). Sorry for that.

What confused me a little was using the `getattr(x, y)` notation instead of the `x.y` one - out of curiosity: is there any practical reason for that (except for using the default value I mentioned above?).

enjoy-digital:
Just for info, I did some minor changes on top of it: https://github.com/enjoy-digital/litex/commit/d9a44ce10fb31ef7035c21ee743a18b0dc5064e8.

PR #750, opened 2020-12-29
boards: arty: add board variant

acomodi:

This PR allows having the board variant option for the arty board, as there can be both a-35 and a-100.


enjoy-digital:
Thanks @acomodi for the PR. I just pushed a simplified version with https://github.com/enjoy-digital/litex/commit/8ff26b730449f0e6947b9ba22ef6362e2c45e0c8.

PR #754, opened 2020-12-31
Use riscv32 gcc triples for picorv32- superset of riscv64 triples.

cr1901:
Self-explanatory. Can confirm a working CPU and BIOS on OrangeCrab using the following invocation in `litex-boards`:

```
python3 litex_boards/targets/orangecrab.py --cpu-type=picorv32 --l2-size=0 --integrated-main-ram-size=32768 --build
```


enjoy-digital:
Thanks @cr1901, merged.

PR #761, opened 2021-01-04
extend builder to take custom bios options

hansfbaier:
Hello I had the problem that I could not customize the bios to
use different IP addresses. This is a general solution which
allows to insert custom variables into the makefile,
which solves that problem.


enjoy-digital:
Thanks @hansfbaier, I'll have a closer look very soon. This can indeed be useful.

hansfbaier:
@enjoy-digital Hi I still pushed some changes to common.mak to allow customizations for all compiler flags (and not only $CPUFLAGS). Another alternative would be to reverse the order of inclusion of variables.mak and common.mak, but I was not so sure if that would break something

hansfbaier:
This issue also would be related to  #645

PR #762, opened 2021-01-04
handle cpu_count argument always as int

dayjaby:
Without this change, the following scenario failed for the master branch of linux-on-litex-vexriscv:

```
./make.py --board=arty --cpu-count=2 --build
```

because it resulted in csr.json containing the constant "config_cpu_count_2": None and the litex_json2dts.py requires the constant "config_cpu_count".


enjoy-digital:
Thanks for reporting and the PR @dayjaby! There was a CI conflict with your change so I fixed directly it with https://github.com/enjoy-digital/litex/commit/16008d3f3a493b081bd25c97fc1f2d64b55016c4.

PR #768, opened 2021-01-05
Lattice NX: Allow up to 320KB of RAM

davidlattimore:
Before, the code was using `log2_int(depth_cascading)+1`, which only works with exact powers of 2


enjoy-digital:
Thanks @davidlattimore, looks good.

PR #772, opened 2021-01-12
Windows fixes for `litex_term`

cr1901:
* Avoid importing modules which don't exist on Windows.
* Set `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag when `litex_term` starts. This tells console processes which have not _already done so_ to interpret ANSI escape codes. Unfortunately it's inconsistent whether `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is enabled on any given Windows console.
* `getch` works at a [lower level](https://wiki.osdev.org/PS/2_Keyboard) than other methods to capture the keyboard. While many chars can be passed to the serial port unmodified, some keys, such as those beginning `0xe0` are worth converting to ANSI equivalents such as "up", "down", "home", etc.

### DIFF
--- a/litex/tools/litex_term.py
+++ b/litex/tools/litex_term.py
@@ -22,10 +22,14 @@ import telnetlib
 # Console ------------------------------------------------------------------------------------------
 
 if sys.platform == "win32":
+    import ctypes
     import msvcrt
     class Console:
         def configure(self):
-            pass
+            # https://stackoverflow.com/a/36760881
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks a lot @cr1901, this has been merged and simplified a bit with https://github.com/enjoy-digital/litex/commit/83fb79fb0cbc6e728f492eba3420fb18c0138e61, could you do a quick test to verify it's still working correctly?

PR #774, opened 2021-01-15
CPU: Vex: add debug slave for dbg cpu variant

kgugala:
This PR adds debug peripheral on a bus if the debug CPU variant is selected


enjoy-digital (approved):
Thanks @kgugala.

PR #775, opened 2021-01-15
I2c scan fixes and cleanups

geertu:


enjoy-digital:
Thanks @geertu, that's merged.

PR #777, opened 2021-01-15
soc/integration/export: Add CSR aliases for devices with suffix zero

geertu:
The LiteX BIOS checks for the presence of the various CSR_*_BASE
definitions to decide to include device support or not.  E.g. for I2C,
it checks for CSR_I2C_BASE.

However, linux-on-litex-vexriscv registers I2C masters using "i2c0"
instead of "i2c", in anticipation for future support for multiple I2C
busses.  As a consequence, CSR_I2C0_BASE is defined instead of
CSR_I2C_BASE, and LiteX BIOS is built without support for I2C.

Fix this by creating aliases for devices with suffix zero in the
generated CSR register definitions, so the LiteX BIOS will pick up the
presence of the corresponding device, and include support for it.

For now this is limited to the first instance of each type.

Signed-off-by: Geert Uytterhoeven <geert@linux-m68k.org>

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -19,6 +19,7 @@
 import os
 import json
 import inspect
+import re
 from shutil import which
 from sysconfig import get_platform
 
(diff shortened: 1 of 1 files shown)


geertu:
I've been thinking a bit about adding e.g. multiple I2C instances.
To keep impact on the BIOS commands minimal, I suggest adding an
"i2c_select" command to specify which i2c bus to talk too.  That way
most logic can be kept in libbase/i2c.c.

For libbase/i2c.c, I'm thinking along the lines of:

>     #define I2C_SINGLE
> 
>     static unsigned int i2c_current;
> 
>     static struct i2c_bus_params {
>             uint32_t (*w_read)(void);
>             void (*w_write)(uint32_t v);
>             ...
>     } i2c_bus_params[] __attribute__((__unused__)) = {
>     #ifdef CSR_I2C0_BASE
>             [0] = {
>                     .w_read = i2c0_w_read,
>                     .w_write = i2c0_w_write,
>                     ...
>             },
>     #endif
>     #ifdef CSR_I2C1_BASE
>     #undef I2C_SINGLE
>             [1] = {
>                     .w_read = i2c1_w_read,
>                     .w_write = i2c1_w_write,
>                     ...
>             },
>     #endif
>     #ifdef CSR_I2C2_BASE
>     #undef I2C_SINGLE
>             [2] = {
>                     .w_read = i2c2_w_read,
>                     .w_write = i2c2_w_write,
>                     ...
>             },
>     #endif
>     ...
>     };
> 
>     #ifndef I2C_SINGLE
>     #undef i2c_w_read
>     #define i2c_w_read  i2c_bus_params[i2c_current].w_read
>     #undef i2c_w_write
>     #define i2c_w_write i2c_bus_params[i2c_current].w_write
>     ...
>     #endif

Thanks for your comments!

enjoy-digital:
Thanks @geertu, I was planning to look at this soon, we can try to work on this together. 

The first thing is indeed the inconsistency between the names of peripherals: as you saw, some are named `periphxy` others `periphxy0` (ex: `timer0` in `SoCCore`, or `i2c0` in the Linux SoC, etc... while others peripherals generally don't have this numbering: `sdram`, `ethernet`, etc...). That would be good to have something consistent  but we should also try to avoid breaking retro-compatibility. As you suggest we could probably create aliases for the first peripheral, to access it with or without the `0` numbering (from the software but also from the gateware design: ex actual `sdram` could also be accessed with `sdram0`.)

The second thing is indeed the support of same peripherals that are instantiated multiple times and I'm still undecided if the BIOS should support it or not: maybe supporting only one instance in the BIOS is enough if the others instances are then supported by the OSes like Zephyr/Linux and declared in the `.dts` files. Allowing too much possibilities and supporting them natively in LiteX can make things very difficult to maintain and I would really like to focus on simplification in early 2021.

So we can maybe try to address the first point for now and decide later for the second point.

PR #780, opened 2021-01-22
integration/export: allow manually specifying toolchain triple.

garytwong:
If the environment variable TRIPLE is defined, use its value as the highest priority candidate.  Useful for testing new cross-compilers, or selecting among toolchains in a different priority than the built-in list.


enjoy-digital:
Thanks @garytwong, this has been merged but the environment variable has been changed to `LITEX_ENV_CC_TRIPLE` with https://github.com/enjoy-digital/litex/commit/b8bcbc522f2eb241cd52b0b8954e5b4ccfc9331f.

garytwong:
Thanks, that's a better name!

PR #781, opened 2021-01-22
build/lattice/trellis: Accept Inverted attribute when processing constraints.

garytwong:
Accept it, and ignore it (handling it is somebody else's job).  This is compatible with the xilinx/vivado class, which does exactly the same thing.


enjoy-digital:
As explain in https://github.com/enjoy-digital/litedram/pull/227#issuecomment-768455114, I'm not sure we should encourage the use of the `Inverted` attribute. I provided an alternative solution for your use case in https://github.com/enjoy-digital/litedram/commit/5a114be7e597767b4540a5292e3f13583f4624e3.

garytwong:
Agreed.  If the PHY generates the inverted clock without adding `Inverted`, then there's no need for the toolchain class to see the attribute at all. Thanks!

PR #788, opened 2021-01-28
liteeth: support hardware timestamping

lschuermann:
This is a companion PR to https://github.com/enjoy-digital/liteeth/pull/59, which introduces hardware timestamping functionality for RX and TX packets in LiteEth. This is useful for time-critical applications such as IEEE 1588 Precision Time Protocol and Time Sensitive Networking.

This PR
- makes the Timer's `uptime_cycles` signal an accessible member on the Timer class, as it might be a popular timestamp source for the LiteEth MAC
- supports passing a timestamp signal into an SoC's `add_ethernet` method. This makes it possible to wire up a hardware-timestamping ethernet interface in a single method call:

  ```python
  self.add_ethernet(phy=self.ethphy, timestamp_source=self.timer0.uptime_cycles)
  ```

### DIFF
--- a/litex/soc/cores/timer.py
+++ b/litex/soc/cores/timer.py
@@ -92,6 +92,6 @@ class Timer(Module, AutoCSR, ModuleDoc):
 
         # # #
 
-        uptime_cycles = Signal(width, reset_less=True)
-        self.sync += uptime_cycles.eq(uptime_cycles + 1)
-        self.sync += If(self._uptime_latch.re, self._uptime_cycles.status.eq(uptime_cycles))
+        self.uptime_cycles = Signal(width, reset_less=True)
+        self.sync += self.uptime_cycles.eq(self.uptime_cycles + 1)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @lschuermann, I just integrated this a bit differently with:
- https://github.com/enjoy-digital/litex/commit/7caed5679027743e16091e7b0481ca796394e636
- https://github.com/enjoy-digital/litex/commit/9bec0ce7a28afb498c7b7eb79943c270225c061a

I think it makes sense in to directly use uptime_cycles in LiteX `add_ethernet` as Timestamp source when Timestamp is enabled  for now. To use another Timestamp source, users can create customized `add_ethernet` methods or we could add another parameter in the future.

I hope this this is fine for you.

lschuermann:
@enjoy-digital Yeah, this looks perfectly fine, thanks for integrating this. I also agree it's more elegant to simply pass in a boolean to the `add_ethernet` helper method!

PR #790, opened 2021-01-29
software/liblitedram: support PHYs with more than 4 DFI phases

jedrzejboczar:
This a part of splitting features required for https://github.com/enjoy-digital/litedram/pull/224.

This PR adds support for more than 4 phases in software. I also added a compile time check that there are no more than 8 phases, because currently I was getting silent errors (was wrapping phase number to 4). Or maybe it would be better to generate these functions as a part of `sdram_phy.h`? Then we wouldn't have to hard-code the switch cases.


enjoy-digital (approved):
Thanks @jedrzejboczar, this looks good. We can merge this PR and as you suggest this could also make sense to auto-generate these functions directly in `sdram_phy.h` in the future.

PR #792, opened 2021-01-30
Add flash method to openFPGALoader class

zguig52:
This is needed to support with generic_programmer usage (needed for linux-on-litex-vexriscv) + add offset/address support for firmware load:
https://github.com/litex-hub/linux-on-litex-vexriscv/issues/189


enjoy-digital:
Thanks @zguig52, looks good, that's merged.

PR #795, opened 2021-02-01
Add automatic generator of GTKWave savefiles

jedrzejboczar:
Currently it happens sometimes that the signal names generated by Migen change from build to build due to some modification in code. When working with simulation it is often helpful to organize signals in groups and add colors, but any changes in the naming require recreating the layout (or modifying the `.gtkw` savefile in text editor). It also happens when a width of a signal changes. 

While working on LPDDR4 simulation I needed to change the signals often, so I added a class that allows to generate a savefile for a SoC from the signals we want to include. I thought it may be useful in general, so I extracted the code from https://github.com/enjoy-digital/litedram/pull/224. 

This PR adds `GTKWSave` class that can be used as show [here](https://github.com/antmicro/litex/blob/fc9ef4c255367addcd912be713368fafcbd5d237/litex/tools/litex_sim.py#L330-L364). This generates a savefile that can be opened with `gtkwave build/sim/gateware/sim.gtkw` and produces the following:
gtkwave

It requires the [pyvcd](https://pyvcd.readthedocs.io/en/latest/vcd.gtkw.html) module. I wasn't sure where would be a suitable place for the file. I placed it under `litex/build/sim/gtkwave.py` but I don't know if it's ok. What do you think about it?

I also added [some default sorters/colorers](https://github.com/antmicro/litex/blob/fc9ef4c255367addcd912be713368fafcbd5d237/litex/build/sim/gtkwave.py#L273-L318) for WishBone and DFI, but new ones could be later added for e.g. AXI or stream endpoints. Also, the [add](https://github.com/antmicro/litex/blob/fc9ef4c255367addcd912be713368fafcbd5d237/litex/build/sim/gtkwave.py#L144) method could be used to automatically add such signals with some default settings.

### DIFF
--- a/litex/build/sim/gtkwave.py
+++ b/litex/build/sim/gtkwave.py
@@ -1,0 +1,318 @@
+# This file is part of LiteX.
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+import os
+import re
+
+from typing import Optional, Sequence, Any, Callable, Generator, Dict, Tuple
+
+from migen import *
+from migen.fhdl.namer import Namespace
+
+
+class SigTrace:
+    """Trace configuration for a single signal"""
+    def __init__(self,
+            name: str,
+            signal: Signal,
+            alias: str = None,
+            color: str = None,
+            filter_file: str = None):
+        self.name = name
+        self.signal = signal
+        self.alias = alias
+        self.color = color
+        self.filter_file = filter_file
+
+
+# Type aliases
+Regex = str
+SigMapper = Callable[[Sequence[SigTrace]], Sequence[SigTrace]]
+
+
+class GTKWSave:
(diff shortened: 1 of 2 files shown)


enjoy-digital (approved):
Thanks @jedrzejboczar, this is a very nice feature and this looks good. This seems to require python version > 3.6 so I'll just move the `gtkw` import to the `generate_gtkw_savefile` to fix `litex_sim` with python 3.6.

enjoy-digital:
@jedrzejboczar: I added this on top https://github.com/enjoy-digital/litex/commit/ba7c503fb65e37cfc3f618e69003304f952dd5ff but I'm not sure if the minimal version is python 3.7 or 3.8 (I'm using 3.6). If you think this is 3.8, I could update the error message.

PR #805, opened 2021-02-08
build/xilinx/symbiflow: drop IDELAYCTRL if declared without IDELAYE2

kowalewskijan:
Signed-off-by: Jan Kowalewski <jkowalewski@antmicro.com>

In xilinx targets in litex-boards there is the `IDELAYCTRL` declared by default, in SymbiFlow it is assumed that the number of `IDELAYCTRL` and `IDELAYE2` is the same, thus single `IDELAYCTRL` without any `IDELAYE2` results in an error. However according to dcp generated from Vivado, declaring only `IDELAYCTRL` without any corresponding `IDELAYE2` results in dropping the `IDELAYCTRL` (it is not placed). This PR adds dropping functionality when we have a similar situation using SymbiFlow toolchain.

### DIFF
--- a/litex/build/xilinx/symbiflow.py
+++ b/litex/build/xilinx/symbiflow.py
@@ -208,9 +208,22 @@ class SymbiflowToolchain:
         platform.finalize(fragment)
 
         # Symbiflow-specific fixes
+        idelayctrl_num = 0
+        idelaye2_num = 0
+        idelayctrl = None
         for instance in fragment.specials:
+            if hasattr(instance, "name_override"):
+                if "IDELAYCTRL" == instance.name_override:
+                    idelayctrl_num += 1
+                    idelayctrl = instance
+                if "IDELAYE2" == instance.name_override:
+                    idelaye2_num += 1
             if isinstance(instance, Instance):
                 self._fix_instance(instance)
+        if idelayctrl_num == 1 and idelaye2_num == 0:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, this feature should probably be handled in Symbiflow directly (because here it only works with direct Migen code, not when the case will happen from re-imported code or in VHDL/Verilog/etc...), can you give a bit of context? are you planning to implement this in Symbiflow and remove the workaround later?

kowalewskijan:
Thanks for pointing out that it won't work with re-imported code, I'll check the symbiflow part how we can drop `IDELAYCTRL` while there is no corresponding `IDELAYE2`.

Previously there was a condition in some boards like `arty` to drop the `IDELAYCTRL`. There was a cleanup some time ago ([link](https://github.com/enjoy-digital/litex/pull/704/files#diff-7938e51b19129094b1c3bf5bbfd80566d2fb6ed9d719f78294e69e7a3c6b0081)) which removed the case in the `CRG` for symbiflow to not instantiate `IDELAYCTRL`. While I bumped litex submodule in symbiflow-arch-defs which apparently used litex code before workaround cleanups I had again a problem with `IDELAYCTRL`.

Let's hold this PR for now, I'll close it if I find out a way to handle this directly in SymbiFlow.

kowalewskijan:
We'll add a solution for this in the SymbiFlow part, closing.

PR #806, opened 2021-02-08
Bios improvements

geertu:


geertu:
Fixing the signed vs. unsigned char issue completely probably requires changing the return type of readchar() from char to int, cfr. glibc fputc ("fputc() writes the character c, cast to an unsigned char, to stream").  However, that would be more intrusive and risky, as this may require changing the uart code, too.

enjoy-digital:
Thanks @geertu! That's merged.

PR #807, opened 2021-02-08
build/xilinx/symbiflow: fix bitstream_device select

kowalewskijan:
Signed-off-by: Jan Kowalewski <jkowalewski@antmicro.com>

There was a change that was added mistakenly to xc7a200t support in the other PR. I reverted this change regarding bitstream_device and added a proper comment to avoid such mistake in future.


enjoy-digital:
Thanks @kowalewskijan, that's merged.

PR #809, opened 2021-02-08
cpu/mor1kx: Add initial SMP support to cpu core

stffrdhrn:
In order for mor1kx to run an SMP kernel shadow registers must be
enabled.  This patch adds two new variants:

 - linux+smp - basic linux + smp support
 - linux+smp+fpu - linux with FPU and smp support


enjoy-digital:
Thanks @stffrdhrn, that's merged.

PR #813, opened 2021-02-11
software/bios: add option to disable BIOS prompt

jedrzejboczar:
This allows to avoid a lot of unnecessary printing when running a Verilator simulation, saving a lot of time. It also omits the `crcbios()` check, which also seems unnecessary for simulation. It is useful for LPDDR4 simulation: https://github.com/antmicro/litedram/blob/9cb698360dc7f83f83a12eb64bed529d8802f973/litedram/phy/lpddr4/simsoc.py#L171

### DIFF
--- a/litex/soc/software/bios/main.c
+++ b/litex/soc/software/bios/main.c
@@ -91,6 +91,7 @@ int main(int i, char **c)
 	uart_init();
 #endif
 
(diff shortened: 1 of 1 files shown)


enjoy-digital (approved):
Thanks @jedrzejboczar, this can indeed be useful and we can merge it. Maybe at some point we could regroup the different `CONFIG_SIM_XY` options if we need to introduce others.

PR #818, opened 2021-02-14
Vexriscv "lite" uses "--mulDiv true", so enable "mul" instructions.

tcal-x:
This is the Makefile that builds VexRiscv_Lite.v:   
https://github.com/litex-hub/pythondata-cpu-vexriscv/blob/master/pythondata_cpu_vexriscv/verilog/Makefile#L11

`--mulDiv true` is specified, so the "Lite" CPU supports the "m" extension.  Same with "LiteDebug".

We will pick up some free performance by enabling the compiler to generate "mul" instructions.


enjoy-digital:
Thanks @tcal-x for catching this! This is merged.

PR #820, opened 2021-02-15
fixes for boot and (single-block, cmd17-only) Linux driver

gsomlo:
With the current sdcard clocking, when building a rocket-litex system at exactly 50 MHz, when 12.5 or 25 MHz are requested, it is possible to obtain those frequencies by using an exact power-of-two divisor for the sdclock (4 and 2, respectively). However, when driven at the full 25 MHz during `sdcardboot`, data transfer from the sdcard hangs. The first patch adds an additional halving of the sdclock frequency to ensure the sdclock frequency is never ***too*** fast for `sdcardboot`.

The second patch allows the current single-block-at-a-time (cmd17-only) Linux driver to avoid running into command timeouts when e.g., `dd`-ing a large file to `/dev/null` from the mounted sdcard. This doesn't help enable multi-block (cmd18) transfers without timeouts, but I will open a separate issue to keep track of those debugging efforts.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1480,7 +1480,7 @@ class LiteXSoC(SoC):
             sdcard_pads = self.platform.request(name)
 
         # Core
-        self.submodules.sdphy  = SDPHY(sdcard_pads, self.platform.device, self.clk_freq)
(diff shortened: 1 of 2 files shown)


enjoy-digital (approved):
Thanks @gsomlo, this looks fine and will indeed help stabilizing the Linux driver.

PR #821, opened 2021-02-15
build/xilinx/vivado.py: Allow a tcl script to be added as ip

jersey99:
These tcl scripts tend to generate .xci's on the fly. The tcl script is typcially a set of commands from vivado ip generation software (found in vivado.jou, or the vivado console).

Perhaps should be made available upstream to (n)-migen at some point.


enjoy-digital:
Thanks @jersey99, this is merged.

PR #830, opened 2021-02-25
VexRiscv: More general mem_map

dayjaby:
The same memory map works not just for linux, but also other OS like NuttX and Zephyr (e.g. https://github.com/apache/incubator-nuttx/pull/2844). Maybe we can generalize it?


enjoy-digital (approved):
Thanks @dayjaby, this makes senses and simplify things.

PR #832, opened 2021-02-26
WIP: Use jinja templates for code generation

dayjaby:
Beep. Boop. Beep. I've heard jinja templates are cool.

So far only csr.h is created, which looks like this: https://gist.github.com/dayjaby/8f14c4d67ba00c905555cfb2c780f76a


gsomlo:
@dayjaby : I'm not familiar enough with jinja to actually *review* this PR, but was intrigued by the idea, and decided to test it. When I build a LiteX+Rocket bitstream for my `nexys4ddr` board using the commits in this PR, with the following command:

```
litex-boards/litex_boards/targets/nexys4ddr.py --build \
   --cpu-type rocket --cpu-variant linux --sys-clk-freq 50e6 \
   --with-ethernet --with-sdcard --integrated-rom-size 0x10000
```

... the resulting bitstream prints a few garbled characters and hangs, never reaching the `litex >` bios prompt or attempting to boot anything. Without the commits in this PR, things work as expected. I haven't made it as far as diff-ing the two resulting sets of generated `csr.h` and `soc.h` files, but I'm hoping that doing exactly that might help narrow down what's actually going wrong.

dayjaby:
@gsomlo that's helpful. I will try to fix it! Tested it only with lxsim so far

enjoy-digital:
Thanks @dayjaby, I'm not familiar with Jinja but understand it can have interesting aspects but the dosage between Python code and Jinja templates in your implementation seems a bit extreme :):

The main idea in LiteX is to try to almost always use python/C when possible and only introduce other languages when the added value is largely superior to the complexity introduced. The code generation can probably be improved in LiteX and we could use Jinja templates with parsimony when really useful (I see more interest to use Jinja for the .svd  header or the build backends than for .h headers generation for example) and try to find a good dosage between the complexity of the native python code and Jinja template (knowing that most people are able to read easily Python code but not  Jinja templates). Your implementation improve some aspects, but also adds complexity on others aspects:
- Introduces a new language users are not necessarily familiar with.
- Splits the code which makes it more difficult to follow (When using Jinja temlplate, we should probably include the template in the python file to ease readability).
- Probably go too far with embedded python code which makes things difficult to understand: This for example: https://github.com/enjoy-digital/litex/pull/832/files#diff-30f1430d86c5877a04de12448097d638850eda7a94981d8f55a66f148652909aR9-R37 should be in native Python not embedded in a template.

So the idea is interesting, but the additional value vs added complexity is not well balanced here and we could probably start using Jinja but not that far :). This is just a first feedback with my initial feeling, I need to think about this a bit more, so I just wanted to say that some ideas are interesting, but that we'll not be able to merge as is, so don't spend too much time on it for now.

dayjaby:
Thanks for the review!

> Probably go too far with embedded python code which makes things difficult to understand

That's a fair point. I did not like the calculations etc. inside Jinja as well and I think the only solution would be to have the necessary calculations already done by LiteX (and not during the exports at all). E.g. calculating the start and end bits of a CSR could be part of the CSR classes which would simplify this template as well as other potential exports in the future. In https://github.com/enjoy-digital/litex/pull/832/files#diff-77b7d5ad59cd45ea4d211f472d51014df3c91e62d59eefc4ada91098479d9d4b I did something similar (maybe a `@property` would be better than cascading the arguments though..)

I could spend some more time on this, trying to nullify the calculations done in the templates.

> Splits the code which makes it more difficult to follow

To be honest, that's rather an advantage to have the template split from the code. If I am interested in changing the csr.svd for example, I would know exactly what to change: the csr.svd.jinja. Furthermore your code editor is able to do proper syntax highlighting.

> Introduces a new language users are not necessarily familiar with.

It's quite pythonic and it can be made even more pythonic, e.g. replacing `csr | hasattr("simple_csrs")` by `hasattr(csr, "simple_csrs")`.

enjoy-digital:
@dayjaby: The main advantage I can see using Jinja template is mostly to simplify the text formating that becomes really easy with it. Having the template directly embedded in the Python code would already simplify some part of the current code and ease readability: In the same python function you have the pre-processing done directly in Python and the simple template  (Eventually splitted in several templates when useful) we'll no longer have the formatting part. Maybe we should start with simple changes where Jinja is clearly improving things:
- Memory.x.
- C CSR Fields.
- git.h
- mem.h
- regions.ld
- soc.h
With embedded templates in the Python code first. 

Then for the other generation functions requiring more complex templates, we could indeed make sure that all the complex processing is done in native Python and that a simple Python structure is passed to the Jinja template for the generation.

And later, we could see if exporting the templates to separate files can make sense we could eventually do it.

If you want to  work on this, we could probably work on smaller PRs first following the steps I was describing. This would be easier to let our ideas converge and merge progressively things.

dayjaby:
@enjoy-digital I embedded the templates into export.py, except for the two template includes inside get_csr_header, if that's ok with you. I took the svd template out for now (the code still lives in https://github.com/dayjaby/litex/tree/jinja_templates_all so it's not lost :))

enjoy-digital:
Thanks a lot @dayjaby, I think we agree on the direction to go and your implementation should provide a good basis to start integrating things. I'll start re-integrating the changes manually this week or next week.

enjoy-digital:
Sorry I haven't been able to have a closer look at it. I'm not familiar with jinja so even if these changes seems nice on paper, it will increase maintenance burden. We can keep things are they are for now and will eventually rework in the future if needed.

PR #833, opened 2021-02-28
builder.py: Add -j 4 to make command

alanvgreen:
Execute make with -j 4, causing it to execute up to 4 compilations
concurrently. The decrease in build times is significant in cases where
no synthesis is required, such as when building a verilator simulation.

4 concurrent jobs was chosen as a conservative setting even for slightly
older machines.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -187,7 +187,7 @@ class Builder:
                 dst_dir = os.path.join(self.software_dir, name)
                 makefile = os.path.join(src_dir, "Makefile")
                 if self.compile_software:
-                    subprocess.check_call(["make", "-C", dst_dir, "-f", makefile])
+                    subprocess.check_call(["make", "-j", "4", "-C", dst_dir, "-f", makefile])
 
     def _initialize_rom_software(self):
         bios_file = os.path.join(self.software_dir, "bios", "bios.bin")


litex/soc/integration/builder.py
```diff
@@ -187,7 +187,7 @@ def _generate_rom_software(self, compile_bios=True):
                 dst_dir = os.path.join(self.software_dir, name)
                 makefile = os.path.join(src_dir, "Makefile")
                 if self.compile_software:
-                    subprocess.check_call(["make", "-C", dst_dir, "-f", makefile])
+                    subprocess.check_call(["make", "-j", "4", "-C", dst_dir, "-f", makefile])
```
mithro, line 190: I wonder if this should be something like `os.environ.get('MAKEFLAGS', '')`? See https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

mithro: Also https://www.gnu.org/software/make/manual/html_node/Parallel.html#Parallel

alanvgreen: I ended up proposing this change because MAKEFLAGS was apparently not being respected.

On a closer reading of the documentation, that's because litex is being started in a recursively executed makefile, and the outermost makefile is likely stripping -j from MAKEFLAGS. 

Maybe the better approach is to ensure that the MAKEFLAGS variable contains a -j flag.

mithro: Is the other Makefile telling make it needs to give the LiteX command the flags (via prefixing the command with a `+`)? https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable and https://www.gnu.org/software/make/manual/html_node/Job-Slots.html#Job-Slots

mithro: PS Anyone ever tried http://bashdb.sourceforge.net/remake/ ?

geertu: All Makefiles calling "make" should call "$(MAKE)" instead.

geertu:
> 4 concurrent jobs was chosen as a conservative setting even for slightly
> older machines.

Usually this is handled by specifying "-j $((nproc))".  But I'd use this only were MAKEFLAGS can't be propagated through another Makefile calling "$(MAKE)", i.e. only for e.g. Python files calling "make".

alanvgreen:
I'll close this PR for now. I got a good result for my own use case by starting the litex build with

MAKEFLAGS=-j8 python ...

PR #836, opened 2021-03-02
demo with c++ HW example

smosanu:
brings a hello world C and C++ as a minor addition to the demo, might be of help to users since c++ wasn't in the Makefile


enjoy-digital:
Thanks @smosanu, that's a good idea. It's merged.

Issue #698, opened 2020-11-12
Determine ROM size at build time.


enjoy-digital:
We are currently using a default ROM size (`integrated_rom_size`) that is large enough for the designs with the default peripherals enabled.

When enabling peripherals (ex SDCard, SATA), we often need to increase the default value. To avoid this, we could do the build in two passes: a first pass to determine the ROM size that will be required by the BIOS and a second where the ROM is automatically resized. This would avoid having to specify manually the ROM size and would also make sure the ROM size is always minimal.

enjoy-digital:
Implemented with https://github.com/enjoy-digital/litex/commit/3cbdc567ff4c50610e3b67c8e7b4ece9fe4da628.

PR #840, opened 2021-03-06
builder: estimate rom size

dayjaby:
This should handle the issue https://github.com/enjoy-digital/litex/issues/698. Not thorougly tested, e.g. need some safety checks for boards without rom. @enjoy-digital is this the kind of solution you had in mind?

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -65,8 +65,14 @@ class Builder:
         csr_svd          = None,
         memory_x         = None,
         bios_options     = [],
+        estimate_rom_size= False,
         generate_doc     = False):
         self.soc = soc
+
+        self.estimate_rom_size = estimate_rom_size
+        if estimate_rom_size:
+            # start estimation with very large rom size
+            self.soc.bus.regions["rom"].size = self.soc.integrated_rom_size * 256
 
         # From Python doc: makedirs() will become confused if the path elements to create include '..'
         self.output_dir    = os.path.abspath(output_dir    or os.path.join("build", soc.platform.name))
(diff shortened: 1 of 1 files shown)


dayjaby:
Some files need to be built twice if running the flag --build. With any other flag, we should be able to skip some file generations.

enjoy-digital:
@dayjaby: Thanks for looking at this! This seems similar to what I had in mind yes: Estimate the ROM size with a first build with a large ROM and then decrease it when actual size is known. I'll try to review and merge this very soon.

dayjaby:
I think there is still something missing. It does not boot well on Arty A7. Probably something like `self.soc.rom.mem.depth = self.soc.integrated_rom_size//(self.soc.rom.mem.width//8)` is still missing. I will test it in the evening.

enjoy-digital:
Thanks @dayjaby, I just looked at it and found an easier solution to do it (by modify the mem.depth at the end of the buiild, which avoid doing the two build iterations), this is implemented with https://github.com/enjoy-digital/litex/commit/3cbdc567ff4c50610e3b67c8e7b4ece9fe4da628. Looking at this on your side was definitely a motivation to also look at it on my side, so even if your implementation has not  been merged, thanks for looking at this, I hope you learned more about the internals in the process.

dayjaby:
@enjoy-digital That's a nice solution! I just tested my solution with the extra line `self.soc.rom.mem.depth = self.soc.integrated_rom_size//(self.soc.rom.mem.width//8)` which made the boot work as well, but with 2 iterations.

About your solution: I think some remnants still contain the old memory size, namely the rom region and thus the regions.ld. Are there no side effects if this regions size is estimated too high?

enjoy-digital:
I also wanted to go in your direction initially but wasn't able to find a simple solution. But since the `auto_resize` will only work in the case of the ROM used in Read Only mode (Read+Write can be used to be able to update  it from a bridge), reducing only the size in the FPGA will not be an issue. MMAP access for an address in the initial ROM range but > to the reduced size will still be redirected to it, but the ROM address will just wrap around (but such access will not occur from the BIOS).

PR #845, opened 2021-03-10
xics: Disable endianness swapping

meklort:
The endianess swapping code caused the core to diverge from microwatt resulting in:
    - The xics tests not working as-is: https://github.com/antonblanchard/microwatt/blob/master/tests/xics/xics.h
    - byte writes writing to the incorrect byte
    
    This removes endianswapping and minimizes the delta from upstream for the xics irq.h header.


enjoy-digital (approved):
Thanks @meklort, this makes sense. I was also wondering if the endianess swapping was really required while merging the XCIS PR. Simplicity always win :)

Issue #839, opened 2021-03-04
Problems reading CSRs on ZynqUS+ via PS AXI 


jedrzejboczar:
In the setup described here: https://github.com/antmicro/litex-rowhammer-tester/issues/22 we had issues with reading UART over EtherBone. The output was malformed, for example:
```
[m Bidyu adae aiym

 c oyih 0222 no-iia
BO ul nMr  011:33
LtXgtsa:fd442)

-========SC[m=========-
[mS:    2btdt-i  GB
[mRM[m  8iB
[mDA:   087KB6-i  00Ts(L9CL9

-===== [mntaiain[m======-
rt eeig osfwr oto.
  | | ||bs:-
 0 1dly kdlyt 1tp.
 m:|dly
 m:|dly
 m:|dly
 m:|dly
 m:|dly
 m:|0dly
 m:|dly
Ra eeig
 0 1 0011111111000000|dly:16-3
 m,b:|0000000000111111 eas 2+9
 0 5 0000000000000000|dly:
 m,b:|0000000000000000 eas
 1 0 1000000000000000|dly:0+0
 1 5 0000000000000000|dly:-0-19
 1 7 0000000000000000|dly:
 bs:m,b1dly:18-2
 m,b:|0000000000000000 eas
 2 3 0000000000000000|dly:0+10
 m,b:|000000000000000 eas
 m,b:|00000000000000 eas
 m,b:|0000000000000000 eas
 3 3 0000000000000000|dly:0+18
 m,b:|00000000000000 eas
 3 6 0000000000000000 eas
 4 1 1111111000000000|dly:9+9
 4 4 0000000000000000|dly:-5+13
 m,b:|0000000000000|dly:
 m,b:|0000000000000000 eas
 m,b:|00000000000000 eas
 5 0 0000000000000000|dly:
 5 3 0000000000000000|dly:4+18
 m,b:|0000000000000000 eas
 m,b:|0000000000000000 eas
 m,b:|000000000000000 eas
 m,b:|00000000000000 eas
 6 3 0000000000000001|dly:9+24
 m,b:|0000000000000000 eas
 6 6 0000000000000000|dly:
 m,b:|0000000000000000 eas
 7 1 1111100000000000|dly:6+6
 7 7 0000000000000000|dly:-2-02
 bs:m,b2dly:27-3
SicigSRMt adaecnrl
 rt:04000-x00002i   Wie x000004100 MB B
 Ra:04000-x0e001i   Ra:04000-x00002i
[...]

PR #846, opened 2021-03-10
interconnect/axi: Fix AXILiteDownverterWrite/Read base address.

enjoy-digital:
Downconverter should start on master's addr, not on aligned master's addr.


enjoy-digital:
As discussed in #839 we can now merge this PR.

PR #849, opened 2021-03-16
video: convenience method to add color bar pattern

hansfbaier:
This comes handy for testing. I kept the method signature the same as
the video terminal so it could be swapped out easily for a test pattern.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -18,7 +18,7 @@ from litex.soc.cores.identifier import Identifier
 from litex.soc.cores.timer import Timer
 from litex.soc.cores.spi_flash import SpiFlash
 from litex.soc.cores.spi import SPIMaster
-from litex.soc.cores.video import VideoTimingGenerator, VideoTerminal, VideoFrameBuffer
+from litex.soc.cores.video import VideoTimingGenerator, VideoTerminal, VideoFrameBuffer, ColorBarsPattern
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @hansfbaier, I indeed also had a need for this a few days ago and was thinking about adding this as a simple test to verify video generation but was wondering it would really be useful. You PR confirms this seems to be useful, so thanks :) I'll maybe try to merge some code later between `add_video_colorbars`, `add_video_terminal` and `add_video_framebuffer` later to avoid too much code duplication.

hansfbaier:
Yes I needed it for the DECA board, because for known reasons the other two would not run there.
That helped me find out that the HDMI chip on that board still needs initialization via I2C.
As far as I understand in litex that would have to go into the BIOS using the I2C core from bitbang.py?

enjoy-digital:
Yes, I started looking at an easy way to add a specific I2C config in this case but haven't finished the work yet. The idea would be to optionally add an I2C core to the Video core and pass an initialization config from the design to the BIOS (Addresses/Values dict) that would do the initialization.

hansfbaier:
I found some I2C commands to initialize it 
[here](https://github.com/intel/meta-de10-nano/blob/master/recipes-bsp/u-boot/files/v2017.03/0005-Add-DE10-Nano-HDMI-configuration-and-debug-apps.patch)

PR #852, opened 2021-03-18
litespi: fix cs handling for bulk transfers

mczerski:
For bulk transfers it is essential to keep cs asserted for all
transfers in a block. CS register must directly drive the cs pin
for the software to be able to control the CS pin.

Linux driver related changes are in https://github.com/litex-hub/linux/pull/4.

### DIFF
--- a/litex/soc/cores/spi.py
+++ b/litex/soc/cores/spi.py
@@ -35,7 +35,7 @@ class SPIMaster(Module, AutoCSR):
         self.irq         = Signal()
         self.mosi        = Signal(data_width)
         self.miso        = Signal(data_width)
-        self.cs          = Signal(len(pads.cs_n), reset=1)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @mczerski, I'll have a closer look at it in the next days.

enjoy-digital:
Sorry for the delay @mczerski, unfortunately, we can't merge the change since it would break several external design that are not using bulk transfers, so we have to keep things retro-compatible.

With https://github.com/enjoy-digital/litex/commit/8ce7c583e61d5973a3a7a12a741df0e53ffe0992, things are kept compatible and a new manual mode is added for CS, where CS will be a direct recopy of the CS set in the CSR:

https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/spi.py#L175

This can be used for bulk  transfers.

PR #853, opened 2021-03-18
liteeth: allow to specify nrxslots and ntxslots for liteeth

mczerski:


enjoy-digital (approved):
Thanks @mczerski, this looks good.

PR #858, opened 2021-03-23
gtkwave: fix error when prefix is empty, make treeopen optional

jedrzejboczar:
This fixes an error when e.g. generating savefiles from LiteScope dumps, where prefix should be empty.


enjoy-digital:
Thanks  @jedrzejboczar, that's merged.

PR #859, opened 2021-03-23
soc/cores/cpu/vexriscv_smp cpu per fpu ratio

Dolu1990:
add cpu_per_fpu option to change the ratio between core count and FPU count


enjoy-digital:
Thanks @Dolu1990, that's merged.

PR #860, opened 2021-03-24
Add high speed USB2 serial core based on luna/nmigen

hansfbaier:
ACM Serial from Luna adapted to migen/LiteX.
Weirdly enough,
as soon as I hook it up to an UART/UARTBone it stops working
(does not enumerate on the USB).
When I use it directly like this, it works:
```
        leds = platform.request_all("user_led")
        from litex.soc.cores.usb2 import USBHighSpeedACMSerialPHY
        acm = USBHighSpeedACMSerialPHY(self.platform)
        self.submodules.acm = acm
        self.comb += leds.eq(~acm.source.payload.data)
        self.sync.usb += [
            acm.source.connect(acm.sink),
        ]
```
I tried this in `soc.py`:
```
        # High Speed USB 2 via ULPI PHY
        elif name in ["ulpi_acm"]:
            from litex.soc.cores.usb2 import USBHighSpeedACMSerialPHY
            self.submodules.uart_phy = USBHighSpeedACMSerialPHY(self.platform)
            self.submodules.uart = uart.UART(self.uart_phy, phy_cd="usb")
```
And I tried this in the deca board constructor:
```
        from litex.soc.cores import uart
        self.submodules.uartbone_phy = acm
        self.csr.add("uartbone_phy")
        self.submodules.uartbone = uart.UARTBone(phy=self.uartbone_phy, clk_freq=sys_clk_freq, cd="usb")
        self.bus.add_master(name="uartbone", master=self.uartbone.wishbone)
```
But in either case the USB would not enumerate and/or the whole FPGA would freeze (JTAG terminal dead)
I really wonder why, because both UART and UARTBone actually should create proper clock
domain crossings, if I did not misunderstand the code.

Nevertheless, having a serial stream to the host is already quite useful, I think,
so I would like to submit this, hoping, maybe you have some suggestions.

Best regards,
Hans

### DIFF
--- a/litex/soc/cores/usb2.py
+++ b/litex/soc/cores/usb2.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python3
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2021 Hans Baier <hansfbaier@gmail.com>
+# SPDX-License-Identifier: BSD-2-Clause
+import os
+
+import nmigen
+from nmigen.back                import verilog
+from luna.full_devices          import USBSerialDevice
+
+from migen                      import Module, Signal, ClockSignal, Instance
+from migen.fhdl.specials        import Tristate
+from litex.soc.cores.uart       import UARTInterface
+
+from litex.soc.interconnect.csr import AutoCSR
+
+class _USBACMSerialDevice(nmigen.Elaboratable):
+    """ nmigen device that acts as a 'USB-to-serial' interface """
+    def __init__(self):
+        self.ulpi = nmigen.Record([
+            ("data",  [("i", 8), ("o", 8), ("oe", 8)]),
+            ("clk",   [("i", 1)]),
+            ("dir",   [("i", 1)]),
+            ("nxt",   [("i", 1)]),
+            ("stp",   [("o", 1)]),
+            ("reset", [("o", 1)]),
+        ], name="ulpi")
+
+        # source
+        self.source_valid   = nmigen.Signal()
+        self.source_ready   = nmigen.Signal()
+        self.source_first   = nmigen.Signal()
+        self.source_last    = nmigen.Signal()
+        self.source_payload = nmigen.Signal(8)
+
+        # sink
+        self.sink_valid     = nmigen.Signal()
+        self.sink_ready     = nmigen.Signal()
+        self.sink_first     = nmigen.Signal()
+        self.sink_last      = nmigen.Signal()
+        self.sink_payload   = nmigen.Signal(8)
+
+    def elaborate(self, platform):
+        m = nmigen.Module()
+
+        m.domains.usb = nmigen.ClockDomain()
+
+        m.submodules.usb_serial = usb_serial = \
+                USBSerialDevice(bus=self.ulpi, idVendor=0x16d0, idProduct=0x0f3b)
+
+        m.d.comb += [
+            # wire the tx/rx streams outside
+            usb_serial.tx.payload  .eq(self.sink_payload),
+            usb_serial.tx.valid    .eq(self.sink_valid),
+            usb_serial.tx.first    .eq(self.sink_first),
+            usb_serial.tx.last     .eq(self.sink_last),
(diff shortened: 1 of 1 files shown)


hansfbaier:
For this to work the [luna](https://pypi.org/project/luna/) python package needs to be installed, of course.

enjoy-digital:
Thanks @hansfbaier, that's really interesting and also something I wanted to try soon. zyp on IIRC was also experimenting with Luna/LiteX and got it working with the UART:
 
We can try to discuss both approaches here or on IRC to compare them and understand the issue with the UART in your case.

hansfbaier:
Do you have a link ro zyp's code?

hansfbaier:
I would prefer to discuss here because IRC is harder to keep track of

enjoy-digital:
I just pinged zyp on IRC about this and shared the link to this PR. You can find the IRC discussion here: https://freenode.irclog.whitequark.org/litex/2021-03-12#;zyp

zyp:
My code is still only in a «proof of concept» state, but here's the relevant parts: https://gist.github.com/zyp/0b5e5f1ebcc7307d0bb5a5f8c738f9cf

This will do both serial over ACM and memory read/writes compatible with ValentyUSB's USBWishboneBridge.

I need to be able to build devices with arbitrary endpoints, so I've attempted to build a wrapper layer that lets you instance endpoint objects on the litex/migen side and handles putting together the nmigen parts and bringing the signals over.

hansfbaier:
@zyp  That is already quite more advanced than what I have so far. So I consider what I have a nice exercise and close this request., to avoid the duplication of effort.

hansfbaier:
@enjoy-digital Thanks for the pointers. Will definitely have a look at that! @zyp I could help with some testing if you like.

mithro:
@zyp - Do you have Fomu and QuickFeather hardware?

zyp:
@mithro I do have a QuickFeather, but not Fomu.

mithro:
@zyp - If you send me an email at me@mith.ro with a shipping address I would be more than happy to send you some Fomu hardware to play with.

hansfbaier:
I actually got the approach working fine, if the sys domain runs on the same clock as the USB domain and if I don't use any clock domain crossings. As soon as I connect the clock domain crossing from the usb domain to the sys domain, the USB clock seems to freeze:

image

zyp:
What do you mean by «USB PLL»? How is your CRG set up?

hansfbaier:
@zyp: One PLL fed by the 60MHz ULPI clock, outputting a -120 degree shifted 60Mhz clock which feeds the "usb" clock domain.
Anothe Pll fed by the 50MHz reference clock, outputting a 50MHz clock for the "sys" clock domain.
So if I have two separate PLLs for the two domains, the system is not completely dead if the 60MHz clock from ULPI dies.
And it dies exactly when the commented code above is removed.
I have no idea why that is, I could not find anything suspicios in the verilog yet. It might be another nasty surprise of the Max10 synthesis. I really might want to try that with a different FPGA and PHY to see if it is actually a toolchain issue.

hansfbaier:
@zyp: BTW, what is all the AXI stuff in your implementation for. Why not just use uart.StreamToWishbone() or uart.UARTBone() ?

zyp:
Using native USB rather than UARTBone over ACM should in theory be faster, but I haven't benchmarked the difference. It also avoids the need for a dedicated ACM interface for the purpose, so it's easier to combine with other functionality.

The reason I'm using AXI-Lite rather than Wishbone is simply because it's easier to CDC and litex allows you to mix and match.

hansfbaier:
Thanks! Yes it is easy to see the benefits. Also the Linux ACM drivers seem a bit wonky. So a pure bulk interface is certainly preferrable.

hansfbaier:
Also I now realize you can't really CDC a Stream because the ready Signal runs in the opposite direction so the FIFO for that would need to be flipped too. And this also could not be in sync since one direction goes from fast to slow and the other from slow to fast.

PR #864, opened 2021-03-25
json2dts.py: fix mac0 reg property style for consistency

gsomlo:
Assuming "#[address|size]-cells = <1>", both of the following are
equivalent:

	reg = <start1 size1>, <start2 size2>, ..., <startN sizeN>;

	reg = <start1 size1 start2 size2 ... startN sizeN>;

The second form appears more widely used and popular, including in
the output of json2dts.py, with the exception of the mac0 node, which
uses the first form. This patch makes output generated for mac0
consistent with that for other DT nodes.

### DIFF
--- a/litex/tools/litex_json2dts.py
+++ b/litex/tools/litex_json2dts.py
@@ -198,9 +198,10 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
         dts += """
             mac0: mac@{ethmac_csr_base:x} {{
                 compatible = "litex,liteeth";
-                reg = <0x{ethmac_csr_base:x} 0x7c>,
-                      <0x{ethphy_csr_base:x} 0x0a>,
-                      <0x{ethmac_mem_base:x} 0x{ethmac_mem_size:x}>;
+                reg = <
+                    0x{ethmac_csr_base:x} 0x7c
+                    0x{ethphy_csr_base:x} 0x0a
+                    0x{ethmac_mem_base:x} 0x{ethmac_mem_size:x}>;
                 tx-fifo-depth = <{ethmac_tx_slots}>;
                 rx-fifo-depth = <{ethmac_rx_slots}>;
                 {ethmac_interrupt}


enjoy-digital:
Thanks @gsomlo for catching this, this is merged.

geertu:
While the generated DTB is the same, it is recommended to group tuples by angular brackets.  This improve human readability and enables (future) automatic validation.

enjoy-digital:
Thanks @geertu, so it seems we were using the two different approaches (on Ethernet and SDCard for example) and @gsomlo made it consistent, but from your comment we should maybe go in the other direction and follow the approach that seems recommended. Do you have links or examples of .dts to share that we could use as examples of recommended approach?

geertu:
The examples are in the DT bindings.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/devicetree/bindings

Examples of recent changes to group tuples are
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/log/?qt=grep&q=Group+tuples

There has been ongoing work to automate DTS validation, based on the YAML DT bindings ("make dtbs_check" in Linux).  This is a bit more complex with the generated ones, as they are obviously not part of the Linux source tree.
I plan to have a look at DTS generation soon, and fix the offenders...

enjoy-digital:
@geertu: OK thanks. Neiher @gsomlo or I are DT specialists, so well be happy to follow your advices, recommended changes.

gsomlo:
On Fri, Mar 26, 2021 at 02:14:26AM -0700, enjoy-digital wrote:
> @geertu: OK thanks. Neiher @gsomlo or I are DT specialists, so well be
> happy to follow your advices, recommended changes.

Right -- I was going for "let's do it the same way everywhere", and
after some cursory googling and digging around for examples it seems
that *currently* the `<start1 size1 start2 size2 ... startN sizeN>;`
form is more widely used. But if there's a concerted, on-going effort
to switch over to `<start1 size1>, <start2 size2>, ... <startN sizeN>;`
for well-thought-out reasons, I'm all in favor of doing it that way!

Right now, the only candidates I could find in json2dts.py are
`mac0` and `mmc0`. But I'll defer to @geertu for the update, as there
may well be other stylistic DTS fixes needed.

geertu:
Oh, seems I already had fixed this one before ;-)
d8b844bbda33afb0eae4272137ad4e21a6d35ebf

enjoy-digital:
Ooops... I'll revert this then (and will also apply it to the SDCard). Thanks for your help.

PR #866, opened 2021-03-27
Lattice Diamond/Windows build fixes

shuffle2:
* avoid always calling `get_data_mod` on some package that may not be wanted.
* revert to using `pnmainc` on windows as `diamondc` doesn't exist.


enjoy-digital (approved):
Thanks @shuffle2, this looks good and has been merged.

PR #872, opened 2021-04-04
Remove implicit dependency of LiteX BIOS on LiteX timer peripheral

madscientist159:
On certain higher-spec CPU architectures, such as POWER-compliant CPUs, the LiteX timer peripheral is redundant and uses FPGA die area that may be desired for other logic.  The LiteX BIOS however attempts to directly access the `timer0` peripheral, leading to build failures when the timer is removed.

Add a timer abstraction layer and enable timer-free builds of LiteX for POWER-based (e.g. Microwatt) projects.

### DIFF
--- a/litex/soc/cores/cpu/microwatt/irq.h
+++ b/litex/soc/cores/cpu/microwatt/irq.h
@@ -12,6 +12,8 @@ extern "C" {
 #include <generated/soc.h>
 #include <generated/mem.h>
 
+#include <ppc/ppc64_asm.h>
+
 #ifdef CONFIG_CPU_HAS_INTERRUPT
 
 // Address of exception / IRQ handler routine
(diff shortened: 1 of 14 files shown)


enjoy-digital:
Thanks @madscientist159 for the PR, I will need to have a closer look and need to think a bit more about the HAL introduced for the Timer and if/while doing such changes be sure to cover https://github.com/enjoy-digital/litex/pull/777 would also be interesting.  At a first impression, I think the PR is a bit too specific to your application so would have a preference for this code to be handled out of tree for now.

madscientist159:
@enjoy-digital One factor to consider here is that at the moment if you disable the timer (`with_timer=False`) for any target, the LiteX BIOS fails to compile.  If there's no desire to add support for timers other than the `timer0` peripheral, the `with_timer` option should be removed entirely (forced to the on state).

enjoy-digital:
The LiteX BIOS indeed requires the LiteX timer to compiles correctly, `with_timer` parameter allows disabling the Timer for users that are not using the LiteX BIOS. With https://github.com/enjoy-digital/litex/commit/02328e5236c21b43ce8b312ff6ad378cfc5433b6 an error will now be raised before doing the compilation if the SoC is using the BIOS but with the Timer disabled.

````
./digilent_arty.py --no-timer
````

Will for example raise:
````
ERROR:SoC:BIOS needs timer0 peripheral to be used.
````

We could eventually in the future allow building the BIOS without the Timer (doing the delays directly in the CPU) but supporting other Timer implementations in the BIOS is a bit out of scope for the project. The BIOS also often provides a software reference for users wanting to use the LiteX peripherals with direct access to MMAP/CSR registers, so introducing an HAL has to  be done very carefully to avoid compromising this and we should avoid adding unneeded complexity. So for now I would prefer only supporting the LiteX Timer in the BIOS. Applications that want to use a specific Timer can still disable it but have to use external software.

Issue #870, opened 2021-04-04
Yosys build scripts generate erroneous yosys command `read_systemverilog`


sthornington:
The three yosys build scripts (trellis, icestorm, oxide) generate an invalid yosys command `read_systemverilog` for files whose language is systemverilog.

The correct command is `read_verilog -sv`, the following patch to `trellis.py`  for example fixes the issue:

```
def _yosys_import_sources(platform):
    includes = ""
    reads = []
    for path in platform.verilog_include_paths:
        includes += " -I" + path
    for filename, language, library in platform.sources:
        # yosys has no such function read_systemverilog
        if language == "systemverilog":
            language = "verilog -sv"
        reads.append("read_{}{} {}".format(
            language, includes, filename))
    return "\n".join(reads)
```

I can prepare this as a PR once I confirm that I have permission to contribute to this project (at work).

enjoy-digital:
Thanks @sthornington, https://github.com/enjoy-digital/litex/pull/873 has been merged.

PR #873, opened 2021-04-05
Fix yosys read command for SystemVerilog sources

sthornington:
This is a fix for https://github.com/enjoy-digital/litex/issues/870 -- the generic map to create yosys commands generates an invalid command, `read_systemverilog` when importing SystemVerilog sources.  The correct command is `read_verilog -sv`.


enjoy-digital:
Thanks @sthornington for reporting this and fixing it. This is merged.

PR #874, opened 2021-04-05
[enh] ECP5 DIfferential input support

chmousset:
tested with both Trellis and Diamond


enjoy-digital:
Thanks @chmousset, this is merged.

PR #875, opened 2021-04-06
json2dts: add the framebuffer memory in the 'reserved-memory' entry, …

rdolbeau:
…so that Linux doesn't try to use it for something else.

As far as I can tell Linux doesn't automatically exclude the FB memory so will allocate it to processes, so display is garbled. Writing data to the FB itself cause the processes to crash, corrupt shared library and buffers, etc.

This simply adds a node in the 'reserved-memory' node of the DTS to tell Linux not to use that memory. It fixes all those issues for me, I can use `fbv` or `yaft` on the FB.


enjoy-digital:
Thanks @rdolbeau! The Video Framebuffer integration in Linux is indeed not completely finished and this was indeed missing,

PR #877, opened 2021-04-11
Fix interrupt issue with ethernet on recent Linux-On-Litex-Vexriscv/SMP

rdolbeau:
It seems an overreaching 'interrupt-parent' caused trouble to interrupt routing
This moves 'interrupt-parent' to devices with interrupts only.

Symptoms are `plic` saying 'cpu%d: parent irq not available' and ethernet not working (and the interrupt not shown in /proc/interrupts).

This might solve https://github.com/litex-hub/linux-on-litex-vexriscv/issues/178 and https://github.com/enjoy-digital/liteeth/issues/62 as they look very similar to the issue I encountered -> mentioning @zguig52 @geertu @Chgavilana @enjoy-digital

### DIFF
--- a/litex/tools/litex_json2dts.py
+++ b/litex/tools/litex_json2dts.py
@@ -28,7 +28,6 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
 / {
         #address-cells = <1>;
         #size-cells    = <1>;
-        interrupt-parent = <&intc0>;
 
 """
 
@@ -151,6 +150,7 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
             #size-cells    = <1>;
             bus-frequency  = <{sys_clk_freq}>;
             compatible = "simple-bus";
+            interrupt-parent = <&intc0>;
             ranges;
 """.format(sys_clk_freq=d["constants"]["config_clock_frequency"])
 


geertu:
So the plic is using the wrong interrupt parent, despite its device node not using interrupts, but interrupts-extended, which does include the right parent?
``
                interrupts-extended = <&L0 11 &L0 9>;
``
Looks like a bug in the Linux plic driver.

rdolbeau:
@geertu I'm not quite sure what was happening, as I did not 'bisect' the DTS - the first thing I tried worked after I noticed the PLIC was complaining, so I stuck with that solution. Another possibility is e.g. that the CPU itself must not have an interrupt-parent for things to work. SiFive's DTS I looked at also have a per-device interrupt-parent, which is what I replicated in the DTS generator.

geertu:
@rdolbeau What about moving the `interrupt-parent` property to the `soc` node?  The idea behind the `interrupt-parent` property is not having to specify it in each individual device node.

rdolbeau:
@geertu Could be an option - if @enjoy-digital thinks it's a better solution I'll try to give it a go. But in LIteX the DTS is auto-generated anyway so it's not a big deal to have a per-device entry.

enjoy-digital:
Thanks @rdolbeau for looking at this. @geertu is the expert here :) Even if LiteX .dts is auto-generated, it's easier to review it if the generation is based on usual good practices, so OK to go in @geertu's direction.

rdolbeau:
@geertu  @enjoy-digital OK I'll try to test with the interrupt-parent in the SoC to see what happens; if it works I'll update the PR.

rdolbeau:
@enjoy-digital the expert @geertu was absolutely right ; moving to the SoC entry is enough for Ethernet to work. Fixed the json2dts patch appropriately.

enjoy-digital:
Thanks @rdolbeau, that's merged.

stffrdhrn:
Just a note on this one, @shenki originally suggested moving the node from the soc out to the top level when I originally submitted the interrupt handling changes in: https://github.com/enjoy-digital/litex/pull/745

rdolbeau:
@stffrdhrn Does moving it to the SoC rather than top breaks some other SoC ? (mor1kx & Rocket were mentioned; I assume Rocket still works otherwise @gsomlo would have said something). Having it at the top-level definitely breaks interrupts in Linux on VexRiscv-SMP, as was already noted in #745.

stffrdhrn:
@rdolbeau , nope it works fine in either place on mor1kx. I am fine with this change. I just want to point out reason mentioned by @shenki  in #745 

  > The interrupt controller is global for the soc in both of our supported systems, so it should go at the top level. This way it is inherited by all nodes below it, ie, all of them.

Though, for mor1k, only the soc uses the interrupts so scoping it in soc is fine for us.  I am not sure about all use cases @shenki may have a more clear reason.

geertu:
The interrupt controller is global for the SoC, while the DT root node refers to the whole system.  That's why the interrupt-parent property should be in the /soc node. Yes, the CPU cores are also part of the SoC, but DT is CPU-centric.

PR #879, opened 2021-04-12
use AutoDoc for timer documentation

bunnie:
Not sure why nobody else saw this, but sometime in the last month's
patches sphinx started throwing an error when building docs for
the timer block. The problem is that the body and title are 'None'
and the doc code tries to invoke methods on None.

Changing the doc methodology to AutoDoc and explicitly creating the intro
section fixes this.


enjoy-digital:
Thanks @bunnie, this is merged.

Issue #862, opened 2021-03-25
Add `description` field to soc.svd


xobs:
The SVD description format appears to require a `description` tag that litex does not currently supply. This means that tools such as Renode do not accept it:

```
There was an error executing command 'sysbus ApplySVD @D:\Code\Xous\Core\emulation\renode.svd'
Field 'description' in 'SOC' is required.
```

Litex should generate a `description` field when it produces the SVD file, even if it's blank.

enjoy-digital:
Hi @xobs, @bunnie,

I think this has been implemented through https://github.com/enjoy-digital/litex/pull/880, can we close this issue?

xobs:
Yes, it works for me thanks!

PR #880, opened 2021-04-12
resolve issue #862 add description to soc.svd

bunnie:
This fixes issue #862.

The root cause is that with no description provided it simply would
not put out a description tag, which breaks compatibility with
other programs.

Insert a somewhat useful default description including a timestamp
and the words "LiteX SoC".


litex/soc/integration/export.py
```diff
@@ -397,6 +397,12 @@ def print_svd_register(csr, csr_address, description, length, svd):
     svd.append('    <name>{}</name>'.format(name.upper()))
     if description is not None:
         svd.append('    <description><![CDATA[{}]]></description>'.format(reflow(description)))
+    else:
+        import datetime
```
mithro, line 401: Imports should generally be at the top of the file.

enjoy-digital:
Thanks @bunnie, this is merged.

PR #885, opened 2021-04-18
software/liblitesdcard: Tell the controller when to wait while the card is busy

paulusmack:
Bit 2 of the command register now tells the controller to wait while
the card is indicating that it is busy (by pulling the DAT0 line low).
The card can do this for commands 7 and 12 and app command 41 (and
also for commands 20, 28, 29, 38 and 43, but we don't use those here.)

This sets bit 2 for those commands.

Signed-off-by: Paul Mackerras <paulus@ozlabs.org>

### DIFF
--- a/litex/soc/software/liblitesdcard/sdcard.c
+++ b/litex/soc/software/liblitesdcard/sdcard.c
@@ -164,7 +164,7 @@ int sdcard_app_send_op_cond(int hcs) {
 #ifdef SDCARD_DEBUG
 	printf("ACMD41: APP_SEND_OP_COND, arg: %08x\n", arg);
 #endif
-	return sdcard_send_command(arg, 41, SDCARD_CTRL_RESPONSE_SHORT);
+	return sdcard_send_command(arg, 41, SDCARD_CTRL_RESPONSE_SHORT_BUSY);
 }
 
 int sdcard_all_send_cid(void) {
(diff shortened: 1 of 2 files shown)


mithro:
@gsomlo + @mateusz-holenko - As people working on the sdcard stuff could you review this?

mithro:
@paulusmack - QQ: Does this need a specific version of `litesdcard`?

paulusmack:
> @paulusmack - QQ: Does this need a specific version of `litesdcard`?

In order to get the behaviour of waiting while the card is indicating it is busy, yes, you need my "core/phy: Add ability to wait while card signals that it is busy" commit in litesdcard (see PR 25 in the litesdcard github).

Without that, things will work exactly as they did before, since bit 2 of the command register was previously unassigned. In other words this commit shouldn't break things if you don't have the litesdcard patch.

paulusmack:
Deferring this until https://github.com/enjoy-digital/litesdcard/pull/25 has been accepted would be reasonable.

enjoy-digital:
Thanks @paulusmack, I integrated the changes to handle the busy SDCard state in LiteSDCard and also did some changes to it. I'm going to merge this PR and do modifications and will write a comment in https://github.com/enjoy-digital/litesdcard/pull/25.

PR #886, opened 2021-04-19
soc/integration: generate and add list of peripherals

AleksandraSwierkowska:
This PR adds `--generate-peripherals` parameter. Using that parameter causes generating a list of peripherals in the csv format which is then changed into an array in C (using xxd). The array is being added by the linker script but instead of just adding it to the ROM, there's being added peripherals_rom right after the ROM (`self.soc.mem_regions['rom'].origin + 2**log2_int(self.soc.mem_regions['rom'].size, False)`) to put the array in. The peripherals_rom size matches the array size. In future, there could be an option to put that new peripherals_rom in a specific place but currently, setting an address was causing problems with the ROM resizing. 

The main aim for now is checking if a DTS is correct if it comes from other build then the binary itself but there are also possible other use cases.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -18,9 +18,10 @@ import struct
 import shutil
 
 from litex import get_data_mod
-from litex.build.tools import write_to_file
+from litex.build.tools import write_to_file, replace_in_file
 from litex.soc.integration import export, soc_core
 from litex.soc.cores import cpu
+from migen.fhdl.bitcontainer import log2_int
 
 # Helpers ------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks @AleksandraSwierkowska, sorry I want to review this carefully and I'm a bit busy currently. Will try to look at it soon.

enjoy-digital:
I've been able to have a first look, but for now I'm not sure this the right solution. Another solution could be to embed directly a `.dtb `(full or fragment) that would have the advantage to be directly generated by `litex_json2dts` tool that we are already using on the different Linux projects (and could also probably used for Zephyr). This then be directly used by the OSes when running Linux/Zephyr or to discover capabilities of bridged SoCs. For your usecase, the `.dtb` could just be read and de-compiled for the check. What do you think about this?

AleksandraSwierkowska:
Thank you for having a look, that's an interesting suggestion. I'll see if there's a quick way of implementing it.

enjoy-digital:
Closing since no update and seems to be a duplicate to JSON export.

PR #887, opened 2021-04-20
integration/soc/add_sdcard: Add an interrupt for command completion

paulusmack:
This is useful for long-running commands generally and in particular
for those without any data transfer, such as erase.  It is a
level-sensitive interrupt because that makes it a little harder to
lose interrupts due to incorrect programming.

Signed-off-by: Paul Mackerras <paulus@ozlabs.org>


enjoy-digital:
Thanks @paulusmack, this is merged.

PR #892, opened 2021-04-25
bios: support passing tftp filename to the 'netboot' command

jluebbe:
Combined with the dynamic IP configuration, this makes it possible to use a shared TFTP server, as every developer can use their own boot.json file.

```
litex> eth_remote_ip 192.168.8.13

Remote IP: 192.168.8.13
litex> eth_local_ip 192.168.8.254

Local IP: 192.168.8.254
litex> netboot jlu-boot.json

Booting from network...
Local IP: 192.168.8.254
Remote IP: 192.168.8.13
Booting from jlu-boot.json (JSON)...
Copying jlu-Image-ecpix5.bin to 0x40000000... (6903848 bytes)
Copying jlu-rv32.dtb to 0x40ef0000... (2401 bytes)
Copying jlu-ecpix5.cpio to 0x41000000... (6135296 bytes)
Copying jlu-fw_jump.bin to 0x40f00000... (49636 bytes)
Executing booted program at 0x40f00000
```

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -487,9 +487,13 @@ static void netboot_from_bin(const char * filename, unsigned int ip, unsigned sh
 	boot(0, 0, 0, MAIN_RAM_BASE);
 }
 
-void netboot(void)
+void netboot(int nb_params, char **params)
 {
 	unsigned int ip;
+	char * filename = NULL;
+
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @jluebbe, this is indeed an very interesting feature, when the TFTP Server it shared between developers but also to be able to boot different software easily, this is merged.

Issue #890, opened 2021-04-25
Pull request #884 seems to break nexys video sdram


andrewb1999:
On the current head of master I get the following issue with the digilent_nexys_video.py target in litex-boards:
```
--=============== SoC ==================--                                                                                                                                                                   
CPU:            VexRiscv @ 100MHz                                                                                                                                                                            
BUS:            WISHBONE 32-bit @ 4GiB                                                                                                                                                                       
CSR:            32-bit data                                                                                                                                                                                  
ROM:            64KiB                                                                                                                                                                                        
SRAM:           8KiB                                                                                                                                                                                         
L2:             8KiB                                                                                                                                                                                         
SDRAM:          524288KiB 16-bit @ 800MT/s (CL-6 CWL-5)                                                                                                                                                      
                                                                                                                                                                                                             
--========== Initialization ============--                                                                                                                                                                   
Initializing SDRAM @0x40000000...                                                                                                                                                                            
Switching SDRAM to software control.                                                                                                                                                                         
Write latency calibration:                                                                                                                                                                                   
m0:0 m1:0                                                                                                                                                                                                    
Read leveling:                                                                                                                                                                                               
  m0, b0: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b1: |00000000000000000011111111111111| delays: 09+-22                                                                                                                                                  
  m0, b2: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b3: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b4: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b5: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b6: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m0, b7: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  best: m0, b01 delays: 09+-22                                                                                                                                                                               
  m1, b0: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m1, b1: |00000000000000000011111111111111| delays: 09+-22                                                                                                                                                  
  m1, b2: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m1, b3: |00000000000000000000000000000000| delays: -                                                                                                                                                       
  m1, b4: |00000000000000000000000000000000| delays: -
  m1, b5: |00000000000000000000000000000000| delays: -
[...]

PR #895, opened 2021-04-26
software/liblitedram: fix error in window centering

jedrzejboczar:
This should fix the problem in https://github.com/enjoy-digital/litex/issues/890.
The problem was that max delay was ending up exactly on 32 and the wrapping logic was incorrect in that case.

I did some tests outside of Litex using the following implementation 
[sdram_leveling_center_module.zip](https://github.com/enjoy-digital/litex/files/6375271/sdram_leveling_center_module.zip) and with the fix included I now get the following results:
```
Range 0-9:
m0: |11111111100000000000000000000000| delays: 04+-04
Range 5-25:
m0: |00000111111111111111111110000000| delays: 15+-10
Range 19-32:
m0: |00000000000000000001111111111111| delays: 25+-06
Range 18-31:
m0: |00000000000000000011111111111110| delays: 24+-06
Range 30-6:
m0: |11111100000000000000000000000011| delays: 02+-04
```

### DIFF
--- a/litex/soc/software/liblitedram/sdram.c
+++ b/litex/soc/software/liblitedram/sdram.c
@@ -878,7 +878,7 @@ static void sdram_leveling_center_module(
 	}
 
 	/* Extend the range if it wraps around */
-	if (delay_min_next > 0) {
+	if (delay_min_next > delay_max) {
 		delay_min = delay_min_next;
 		delay_max += SDRAM_PHY_DELAYS;
 	}


enjoy-digital:
@jedrzejboczar: For now I removed the wrap around, since I don't think all aspects were taken into account in the current implementation  (for example no relation between the total delay that can be compensated and the SDRAM period). We'll discuss it together.

PR #897, opened 2021-04-27
software/liblitedram: use single iteration of dq-dqs training

jedrzejboczar:
@enjoy-digital As discussed earlier, I tested with your fix applied and it looks like the 2nd iteration is now not needed. Or at least I haven't observed any improves when using 2 iterations now so this will speed up the training process.


enjoy-digital:

PR #899, opened 2021-04-28
software/bios: Add option to obtain IP address from DHCP server

Meet909:
That PR allows to obtain IP address automatically from DHCP server using `eth_dhcp` command.
That command is only available when with_dhcp argument is set in litex/soc/integration/soc.py.
`dhcp_get_ip` function in litex/soc/software/bios/boot.c invokes `dhcp_resolve` function with MAC address as parameter.
That PR is associated with litex-hub/litex-boards#206.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1367,7 +1367,7 @@ class LiteXSoC(SoC):
                 base_address = self.bus.regions["main_ram"].origin)
 
     # Add Ethernet ---------------------------------------------------------------------------------
-    def add_ethernet(self, name="ethmac", phy=None, phy_cd="eth", dynamic_ip=False, software_debug=False,
+    def add_ethernet(self, name="ethmac", phy=None, phy_cd="eth", dynamic_ip=False, with_dhcp=False, software_debug=False,
         nrxslots       = 2,
         ntxslots       = 2,
         with_timestamp = False):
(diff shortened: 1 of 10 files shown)


jluebbe:
Is there a way to generate a reproducible MAC address based on some information in the FPGA or SPI-NOR? For example a serial number?

mithro:
@jluebbe -- Depends on the FPGA and board, some FPGAs have a "device DNA" that could potentially be used to randomly generate a MAC address. Some FPGA boards have an EEPROM with a MAC address encoded in them (like the Numato Opsis), others have an EEPROM embedded in the Ethernet PHY.

enjoy-digital:
Thanks @MateuszKosmala, this is a very interesting feature. I'll try to review it very soon.

enjoy-digital:
Thanks again for this PR. The DHCP support is useful and the idea still makes sense.

  Since the branch is now quite old and conflicts with current LiteX, I reworked the implementation in a new PR: https://github.com/enjoy-digital/litex/pull/2483.

  The new version keeps the same goal but is a bit more conservative:
  - it uses the existing `ETH_UDP_BROADCAST` path instead of changing the normal UDP receive filtering;
  - it adds a small `libliteeth` DHCP client for the DISCOVER/OFFER/REQUEST/ACK flow;
  - it exposes this through `add_ethernet(with_dhcp=True)`;
  - it adds the BIOS `eth_dhcp` command and can run DHCP before netboot when enabled.

  The original contribution is credited in the new commit with a `Based-on:` reference and `Co-authored-by:` trailer.

PR #902, opened 2021-05-02
Add 'lse' as a --synth-mode alternative with Radiant toolchain.

tcal-x:
Lattice Radiant comes with two synthesis options, 'synplify' and 'lse'.

The LiteX Radiant toolchain uses 'synplify' by default, and allows it to be replaced with
'yosys' using option `--synth-mode 'yosys'`.   Since the `--synth-mode` option already exists, 
it seems to make sense to allow it to specify 'lse' as well, to give a third alternative.


enjoy-digital:
Thanks @tcal-x, this is merged.

PR #904, opened 2021-05-06
I2S fix: sample SYNC on the correct edge

bunnie:
The original Tx path implementation samples SYNC on the falling
    edge, out of convenience with the fact that teh data must also
    change on the falling edge.
    
    This works OK, until you have a CODEC which has a ~40ns max
    delay spec on the SYNC, and also has a slightly asymmetric
    SYNC edge (the SYNC signal is also the WCLK or LRCLK depending on
    which docs you read). The SYNC by spec is supposed to change
    on the falling edge, and this extra delay is enough to cause
    the SYNC to introduce occassional bit or frame shifts into
    the audio.
    
    This fix samples the SYNC on the rising edge, but still
    changes the data on the falling edge, thus allowing for
    implementations where SYNC has quite loose timings relative
    to everything else (as is the case on the TLV320AIC3200)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,75 @@
+[> 2021.04, released on May 3th 2021
+------------------------------------
+
+	[> Issues resolved
+	------------------
+	- litex_term:         Fix Windows/OS-X support.
+	- soc/USB-ACM:        Fix reset clock domain.
+	- litex_json2dts:     Various fixes/improvements.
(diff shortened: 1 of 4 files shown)


bunnie:
I don't understand why github shows four commits when I'm trying to merge just one -- but it looks like the diff is correct, only 6f132d7 has changes relative to master.

enjoy-digital:
Thanks @bunnie, this is merged as a single commit (squash and merge).

PR #905, opened 2021-05-07
lattice/programmer.py: Add iCESugar programmer

hansfbaier:
In preparation for the iCESugar board.
Load and Flash tested: works.


enjoy-digital:
Thanks @hansfbaier, this is merged.

Issue #883, opened 2021-04-15
Custom Function Unit (CFU) integration


tcal-x:
I have been experimenting with custom function units (CFUs) added to VexRiscv (using VexRiscv since it provides a CfuPlugin option, and it's also awesome, although eventually we hope more CPUs support the interface).   The CFU connects ONLY to the CPU, not to the system bus, and it has no CSRs of its own.    I've been using CPU+CFU in LiteX systems for quite a while now, but I've been hiding the CFU from LiteX --- I make a wrapper containing the CPU+CFU, and the wrapper exports just the normal CPU signals, so LiteX doesn't even know the CFU is there.

But now I think it makes sense to hook up the CFU in LiteX, so I'd like to ask your advice.   I've mocked up a couple of ways of doing it.  I have the code for connecting the CFU to the VexRiscv in cores/cpu/vexriscv/core.py --- it creates and hooks the signals, creates the CFU instance, and optionally adds the CFU source.   But it might make more sense to have the CPU just create its interface when needed, and move the CFU instantiation and CFU<->CPU hookup out of the CPU core.py.

I've had a suggestion to create a Record for the CFU<->CPU connections. 

The interface will be fairly stable.   It is basically one `cmd` stream CPU-->CFU with a payload of an opcode and two data arguments, and a `rsp` stream CFU-->CPU with a payload of one data result and a `ok` status bit.   There will likely also be a context_id field in the `cmd` payload, although that is still being finalized.

enjoy-digital:
Thanks @tcal-x, sorry I've not yet been able to look at it, but just saw https://github.com/google/CFU-Playground and now have a better understanding of the need.

tcal-x:
Thank you @enjoy-digital -- I can see how many issues you're handling.   In fact, the liteSPI work is also something I'm happy to see getting attention.

rdolbeau:
@tcal-x Out of curiosity, as you seem interested in performance, have you evaluated upcoming RISC-V extensions as a preliminary step to the full-custom functional unit ?

Your example includes byte-reverse & bit-reverse, both of which are part of bitmanip 'B' as specific sub-cases of `grevi` (`rev8` and `rev`), which can be added to VexRiscv. But the choice of those might be for the sake of simplicity.

[...]

PR #908, opened 2021-05-13
Add CFU integration

robertszczepanski:
Following issue #883 I've added `add_cfu()` function that provides support for [CfuPlugin](https://github.com/SpinalHDL/VexRiscv/blob/master/src/main/scala/vexriscv/plugin/CfuPlugin.scala) provided by SpinalHDL for VexRiscv CPU.

You can add CFU verilog filename via argument `--cfu-filename`. I have tested it on Arty A7 with software generated in [CFU-Playground](https://github.com/google/CFU-Playground/tree/main/proj/example_cfu_v) by executing `make software`.

### DIFF
--- a/litex/soc/integration/soc_core.py
+++ b/litex/soc/integration/soc_core.py
@@ -70,6 +70,8 @@ class SoCCore(LiteXSoC):
         cpu_reset_address        = None,
         cpu_variant              = None,
         cpu_cls                  = None,
+        # CFU parameters
+        cfu_filename             = None,
         # ROM parameters
         integrated_rom_size      = 0,
         integrated_rom_mode      = "r",
@@ -190,6 +192,11 @@ class SoCCore(LiteXSoC):
             if timer_uptime:
                 self.timer0.add_uptime()
 
+        # Add CFU
+        if cfu_filename:
+            assert(cpu_type == "vexriscv")
+            self.cpu.add_cfu(cfu_filename=cfu_filename)
+
     # Methods --------------------------------------------------------------------------------------
 
     def add_interrupt(self, interrupt_name, interrupt_id=None, use_loc_if_exists=False):
@@ -301,6 +308,9 @@ def soc_core_args(parser):
 
     # L2 Cache
     parser.add_argument("--l2-size",           default=8192, type=auto_int, help="L2 cache size (default=8192).")
+
+    # CFU
+    parser.add_argument("--cfu-filename",      default=None,                help="CFU verilog filename.")
 
 def soc_core_argdict(args):
     r = dict()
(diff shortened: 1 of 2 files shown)


mithro:
FYI - @tcal-x

tcal-x:
Hi @robertszczepanski , this is great!   It's perfect as far as the incomplete information that I passed on.  However, I left out a couple of details.

* The CFU also has inputs 'clk' and 'reset'.     These do not come _from_ the VexRiscv, although the VexRiscv also has those pins (as inputs).   Those signals come from the surrounding system (clock region?).   So the 'clk' pins of both the Cfu and the VexRiscv are driven by the same signal, same for 'reset'.   These need to get connected to the Cfu in add_cfu( ).
* The 'io_bus_' prefix is an artifact of the fist Cfu that I generated, from SpinalHDL.   I would like to delete that string from the port names at this point.  You probably have an existing cfu.v as an example for testing with port names with the prefix --- so just delete that prefix from the ports/wires for testing.

robertszczepanski:
Hi, I see that there are indeed these signals in `CfuPlugin`. 
Since `clk` and `reset` are connected to the same signals as `VexRiscv` I connected these to `ClockSignal()` and `ResetSignal()` now.
I also removed `io_bus_` prefix.

litex/soc/integration/soc_core.py
```diff
@@ -190,6 +192,11 @@ def __init__(self, platform, clk_freq,
             if timer_uptime:
                 self.timer0.add_uptime()
 
+        # Add CFU
+        if cfu_filename:
+            assert(cpu_type == "vexriscv")
+            self.cpu.add_cfu(cfu_filename=cfu_filename)
```
tcal-x, line 198: Also assert `("cfu" in cpu_variant)`.   The assertion will be less confusing than getting "`named port connection 'CfuPlugin_bus_cmd_ready' does not exist`" from Vivado.

Eventually when I've build a +cfu version of every VexRiscv variant, the code can modify the variant here if needed (infer the "+cfu").

enjoy-digital:
Thanks @robertszczepanski for the CFU support and @tcal-x for the review. I'm going to merge and address the last @tcal-x comments.

enjoy-digital:
I just did small changes after the merge:
- https://github.com/enjoy-digital/litex/commit/dd72b1acfe8f817693aa414a0398d91edfb2cb90
- https://github.com/enjoy-digital/litex/commit/78bdde04246710afde40c7e814e2946f02ca02bf

The CFU is now specified by `--cpu-cfu` argument and can be a Verilog or VHDL file. It will also defaults to a local `Cfu.v` is not specified.

Idealy, we should only provide the `--cpu-cfu` argument to the target when a CPU CFU support is selected but this problem is also present for other peripherals/extensions, we'll try to address this with https://github.com/enjoy-digital/litex/issues/912.

tcal-x:
Thanks @enjoy-digital!  --- I had been struggling with exactly the issue in #912 trying to figure out how to add the CFU option in a manner consistent with LiteX.

Issue #322, opened 2019-12-21
printf formatting for 64bit integers is broken on 32bit cpus


gsomlo:
I started with applying the following patch (function to test uint64_t printf formatting:
```
diff --git a/litex/soc/software/bios/main.c b/litex/soc/software/bios/main.c
index 621d377a..64c5af06 100644
--- a/litex/soc/software/bios/main.c
+++ b/litex/soc/software/bios/main.c
@@ -552,6 +552,24 @@ static void boot_sequence(void)
        }
 }
 
+static void _do_test_uint64(void)
+{
+       int i;
+       union {
+               uint8_t  foo_chr[8];
+               uint64_t foo_int;
+       } foo;
+
+       foo.foo_int = 0x0123456789abcdef;
+
+       for (i = 0; i < 8; i++)
+               printf(" %02x", foo.foo_chr[i]);
+       printf("\n");
+       printf("%016x\n", foo.foo_int);
+       printf("%016lx\n", foo.foo_int);
+       printf("%016Lx\n", foo.foo_int);
+}
+
 int main(int i, char **c)
 {
        char buffer[64];
@@ -627,6 +645,7 @@ int main(int i, char **c)
                printf("\n");
        }
 
+       _do_test_uint64();
        printf("--============= \e[1mConsole\e[0m ================--\n");
        while(1) {
                putsnonl("\e[92;1mlitex\e[0m> ");
```
I got the following results:
1. with `litex/litex/tools/litex_sim.py --cpu-type mor1kx`:
```
 01 23 45 67 89 ab cd ef
0000000001234567
0000000001234567
0000000089abcdef
```
2. with `litex/litex/tools/litex_sim.py --cpu-type vexriscv`:
```
 ef cd ab 89 67 45 23 01
0000000000000002
0000000000000002
0000000000000002
```
3. with `litex/litex/tools/litex_sim.py --cpu-type rocket`:
```
 ef cd ab 89 67 45 23 01
0000000089abcdef
0123456789abcdef
0123456789abcdef
```

I'd expect `%016Lx` to work on all three CPU models (the explicit 64bit formatting), but it's broken on 32 bit CPUs. I'll investigate when I get some spare time, but in the mean time, if anyone has any ideas for an easy fix (e.g. refresh whatever code we are using for printf in the bios), I wouldn't mind being preempted :)

geertu:
> \+       printf("%016x\n", foo.foo_int);

[...]

PR #910, opened 2021-05-13
software/libbase: fix %lld and %llu in printf functions

danc86:
The vsnprintf() function would correctly interpret %lld and %llu
arguments as "long long" width, but the number() helper function only
accept "long" so the value would be truncated and then printed
incorrectly.

Widen the argument to number() to accommodate up to "long long".

### DIFF
--- a/litex/soc/software/include/base/stdlib.h
+++ b/litex/soc/software/include/base/stdlib.h
@@ -57,7 +57,7 @@ static inline int atoi(const char *nptr) {
 static inline long atol(const char *nptr) {
 	return (long)atoi(nptr);
 }
-char *number(char *buf, char *end, unsigned long num, int base, int size, int precision, int type);
+char *number(char *buf, char *end, unsigned long long num, int base, int size, int precision, int type);
 
 #define   RAND_MAX        2147483647
 
--- a/litex/soc/software/libbase/libc.c
+++ b/litex/soc/software/libbase/libc.c
@@ -424,7 +424,7 @@ int skip_atoi(const char **s)
 	return i;
 }
 
-char *number(char *buf, char *end, unsigned long num, int base, int size, int precision, int type)
+char *number(char *buf, char *end, unsigned long long num, int base, int size, int precision, int type)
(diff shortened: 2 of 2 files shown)


danc86:
Hmm. This still prints garbage for %llu on my Vexriscv SoC under some circumstances. I guess there is still something else wrong here...

gsomlo:
I think this is related to https://github.com/enjoy-digital/litex/issues/322 (which should be closed once you figure out how to cover all the angles with this patch).

The patch in its current form fixes the `printf("%016Lx")` case for me on mor1kx, but still prints out garbage (`89abcdef00000001` instead of the expected `0123456789abcdef`) on vexriscv. Still works OK on 64-bit (rocket), but that was already the case.

geertu:
I gave it a try on vexriscv (which is little-endian, while mor1kx is big endian!).
Before this commit, using the test code from @gsomlo in #22, I got:

```
  ef cd ab 89 67 45 23 01
0000000000000100
0000000010001f0c
0000000010001f0c

```

After this commit, and after adding a few more integer test formats, I got:
```
 +1000000000000000000000000000011101111 1000000000000000000000000000011001101 1000000000000000000000000000010101011 1000000000000000000000000000010001001 1000000000000000000000000000001100111 1000000000000000000000000000001000101 1000000000000000000000000000000100011 1000000000000000000000000000000000001
x:   1000000100
lx:  1000000020
Lx:  1000000020
llc: 1000000020
u:   1000000020
lu:  1000000020
llu: 1000000020
main:1000000020
```
Which is even more garbage.  Is this a sign of stack corruption or overflow?

enjoy-digital:
Thanks @danc86, now that's merged now that the the mysterious behaviour is fixed with #988 (https://github.com/enjoy-digital/litex/pull/988#issuecomment-896688637).

enjoy-digital:
@danc86: Sorry I  indeed had to revert it since it was having the same behavior seen by @geertu. I'll have a closer look at it.

PR #911, opened 2021-05-15
software/liblitespi: Fix speed test.

zyp:
Before this patch, the loop would finish with `lowest_div` either set to the first failing value
or 0 even if it succeeded with 0. Fix it so that if all tests pass, it’ll end up being -1 before
the incrementation.

This patch also skips retesting the original value. If the retest failed, `lowest_div` would be incremented past the original value and could potentially wrap around.


enjoy-digital:
Thanks @zyp for the fix, this is merged.

PR #915, opened 2021-05-17
software/liblitespi: Remove manual mode control.

zyp:
An upcoming PR for https://github.com/litex-hub/litespi/issues/46 is adding automatic arbitration between MMIO and master cores, so the CSR for manual control is being removed.

In existing gateware, mode defaults to MMIO, so in most cases this check does not matter anyway.


enjoy-digital (approved):
Thanks @zyp, this https://github.com/litex-hub/litespi/pull/47 has been merged so we can also merge it.

PR #920, opened 2021-05-19
Display more precision in BIOS mem_speed test.

tcal-x:
Rather than jumping from "999KiB/s" to "1MiB/s",
where 1MiB/s could be anywhere from 1.00 to 1.99 MiB/s,
instead go up to "9999KiB/s" and then "10MiB/s", to always
provide at least two significant digits of accuracy.

### DIFF
--- a/litex/soc/software/libbase/memtest.c
+++ b/litex/soc/software/libbase/memtest.c
@@ -162,11 +162,11 @@ int memtest_addr(unsigned int *addr, unsigned long size, int random)
 }
 
 static void print_size(unsigned long size) {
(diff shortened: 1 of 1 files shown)


tcal-x:
I noticed that "10*GIB" overflows the 32b that we have with "long int".

tcal-x:
As I mention in the commit, I don't like the asymmetry in my updated change, but I think it's ok here since the GiB level will rarely be used.

tcal-x:
That's great @enjoy-digital, I like it.

PR #923, opened 2021-05-26
software/lib*: address gcc warnings

gsomlo:
Fix gcc warnings: use 'unsigned long' to represent memory addresses,
and remove 'static' from the definition of 'cdelay()', as it is called
from multiple C files.


enjoy-digital (approved):
Thanks @gsomlo.

PR #926, opened 2021-05-31
RFC: Power optimization: multiple gated clocks

bunnie:
@enjoy-digital --

While implementing power management, I found a need to make multiple gated clocks for various domains. 

I propose supporting the specification of multiple gated clocks with a dictionary that consists of the clock signal name, and the gating condition.

Furthermore, according to UG472 I've modified the gating primitive away from a BUFGCE to an explicit BUFGTRL set up like a BUFGMUX. Apparently, the "CE" mode is deprecated and the "mux" mode is preferred because it's guaranteed to be glitchless.

There's an additional platform command that is necessary to prevent the place&route from working overtime to get the gating signal to meet the setup/hold of the clock. Reading in detail to the datasheet, it seems this "setup/hold" time is only necessary if your design requires all clocks to stop on exactly the same edge for a clock gate. i didn't add this constraint here, but for power savings it's not a big deal if an extra clock pulse slips through, so long as the clock transition is glitchless (which it is, hence the false path is ok).

This is the effective platform command that I use to ignore the gating path as a timing constraint:
```
platform.add_platform_command('set_false_path -through [get_pins BUFGCTRL*/S*]')
```

I was able to get about 30% power savings using an OS-driven clock-stopping routine where when the thread scheduler has nothing to run, the clocks are gated off. In this implementation the MMCME2 is still powered and running.

Further digging indicated a couple of interesting factoids, maybe of use to you in the future:

1. Gating clocks is instantaneous, glitchless, and fairly easy to do, but you don't get the full power savings.
2. You can use the `RST` input to turn off the MMCME2 for a faster power-on cycle. From what I can tell of some scrounging on the web for answers, RST will stop your clocks, but it will keep the VCO running and the regulators powered up, so that some power is saved, but your resume time is medium-fast
3. The 'PWRDWN` input will fully power off the MMCME2, including turning off the VCO and the regulators for the block. This leads to a much longer power-on time.

The difference between `RST` and `PWRDWN` are not characterized in the Xilinx datasheets, unfortunately.

I could not get these more aggressive modes to work I *think* in part because shutting down the PLL also kills the IDELAYCTRL block, and I'd have to introduce some extra logic to prevent any fetches to the SPINOR as it relies on this block to correctly phase-align the DQS strobe to the data. I decided it wasn't worth it to try to push this hard for power savings at this point, and stuck with clock gating only.

In addition to introducing logic to hold off SPINOR access until the IDELAYCTRL is ready, probably another thing that would need to be introduced is an *actual* clock MUX so certain "always on" domains keep clocking a little bit during the low power mode, even if the PLL is off. This is to deal with wake-up signal sources that e.g. run through synchronizers that require multiple clock domains for the signal to propagate to its destination properly.

So, I could see an extension to the currently proposed "dictionary of gates" to also create an option for a "dictionary of multiplex options", or something like that, where a gating is selecting between a `clock` and `0`, and a mux is picking between a `clock` and a `clock`, but that may be a bit down the road...


enjoy-digital:
Thanks for the PR and details @bunnie, this is interesting and looks good. I'll do a closer review very soon.

PR #936, opened 2021-06-09
Sync ROM_BOOT_ADDRESS with main_ram location

developandplay:
With this change the simulation will adjust automatically to the memory layout.

Rocket and Blackparrot main_ram starts at 0x80000000
Vexriscv main_ram starts at 0x40000000


enjoy-digital:
Thanks @developandplay, this makes sense.

PR #941, opened 2021-06-11
soc/cores/ram: allow populating initial values in Nexus LRAM

danc86:
On designs which use Nexus parts without any external memory, it can be
difficult to fit an embedded ROM program larger than a few KiB. Radiant
cannot infer LRAM, and refuses to infer EBRAM under many circumstances
too, so large memories tend to just consume a huge number of LUTs.

This patch makes it possible to explicitly wire up an LRAM as a ROM,
populate its initial values with a program, and execute directly from
it. That lets us embed programs up to 64KiB.

This implementation is a bit ugly and is not consistent with regular Memory .init attribute, so I am open to suggestions of how to improve this.

Confirmed this produces the expected values in LRAM with Radiant.

### DIFF
--- a/litex/soc/cores/ram/lattice_nx.py
+++ b/litex/soc/cores/ram/lattice_nx.py
@@ -21,37 +21,68 @@ Note that this memory is dual port, but we only use a single port in this
 instantiation.
 """
 
+
+def initval_parameters(contents, width):
+    """
+    In Radiant, initial values for LRAM are passed a sequence of parameters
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @danc86, this has been merged, we could maybe improve it in the future, but for now it should be OK. I just did minor changes to the API for now with https://github.com/enjoy-digital/litex/commit/7179d88e8ca273c02215ee4ab956fbf0c45cec22 to have something more consistent with Migen's Memory and avoid breaking it later.

Issue #757, opened 2021-01-03
Move software libs to cores


enjoy-digital:
Now that the BIOS has been refactored/cleaned-up to with specific libs for each core, we should move these libs to the cores. This would ease adding hardware/software features to the cores, following changes (or finding regressions) and is one of the required step to make the cores pluggable.

PR #942, opened 2021-06-13
Allow external software packages to be linked into the BIOS

zyp:
Hi, this is a set of two commits to allow external code to be linked into the BIOS, as a proposed starting point for #757.

The first commit modifies the builder and the BIOS Makefile to allow packages/libraries to be registered with the builder and passed through to the Makefile to get linked into the binary.

The second commit adds a registration mechanism for initialiation functions to the BIOS, following the same pattern used for command registration.

This has been successfully tested to link in hyperram initialization code without having to modify the litex repo.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -46,9 +46,6 @@ soc_software_packages = [
     "libliteeth",
     "liblitesdcard",
     "liblitesata",
-
-    # BIOS.
-    "bios"
 ]
 
 # Builder ------------------------------------------------------------------------------------------
@@ -105,15 +102,24 @@ class Builder:
         # Documentation
         self.generate_doc = generate_doc
 
-        # List software packages.
+        # List software packages and libraries.
         self.software_packages = []
+        self.software_libraries = []
+
         for name in soc_software_packages:
             self.add_software_package(name)
+
+            if name == "libbase":
+                name += "-nofloat"
+            self.add_software_library(name)
 
     def add_software_package(self, name, src_dir=None):
         if src_dir is None:
             src_dir = os.path.join(soc_directory, "software", name)
         self.software_packages.append((name, src_dir))
+
+    def add_software_library(self, name):
+        self.software_libraries.append(name)
 
     def _generate_includes(self):
         # Generate Include/Generated directories.
@@ -123,11 +129,17 @@ class Builder:
         # Generate BIOS files when the SoC uses it.
         with_bios = self.soc.cpu_type not in [None, "zynq7000"]
         if with_bios:
+            self.add_software_package("bios")
 
             # Generate Variables to variables.mak.
             variables_contents = []
             def define(k, v):
                 variables_contents.append("{}={}".format(k, _makefile_escape(v)))
+
+            # Define packages and libraries.
+            define("PACKAGES", " ".join(name for name, src_dir in self.software_packages))
+            define("PACKAGE_DIRS", " ".join(src_dir for name, src_dir in self.software_packages))
+            define("LIBS", " ".join(self.software_libraries))
 
             # Define the CPU variables.
             for k, v in export.get_cpu_mak(self.soc.cpu, self.compile_software):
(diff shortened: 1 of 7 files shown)


enjoy-digital:
Thanks a lot @zyp, this is indeed a very good first step for #757. I also did some test here with it without troubles, so we can merge it.

bunnie:
This patch breaks our build system, because we don't use any Litex BIOS at all. Our BIOS is written in Rust (uses `cargo`, not `make`), so we can't integrate into the Litex C-based build system.

Previously, the fix to this was to specify `compile_software=False` in the arguments to `Builder`, and we would shim in the BIOS file by running our build script and placing the `bios.bin` into the `run/software/bios/` directory. 

I think this approach still works, but we don't have a good way of telling Litex to build no BIOS at all with this patch... It seems that the `with_bios` flag is computed based on the CPU type, and there is no other way to override it now:

```
        # Generate BIOS files when the SoC uses it.
        with_bios = self.soc.cpu_type not in [None, "zynq7000"]
```

I think if there could be a separate flag that just turns off BIOS building altogether, that would allow us to co-exist with this patch, but maybe there are other ways to do this, too?

bunnie:
To be clear, our method this means we do a two-pass build: once to create the SVD files that we need to create the BIOS, then we compile the BIOS and place the artifact in bios.bin, and then another pass to actually incorporate the BIOS. This is our full flow:

```
    # pass #1 -- make the SVD
    platform = Platform(dutio)
    soc = Dut(platform, spiboot=boot_from_spi)

    os.system("mkdir -p ../../target")  # this doesn't exist on the first run
    builder = Builder(soc, output_dir="./run", csr_svd="../../target/soc.svd", compile_gateware=False, compile_software=False)
    vns = builder.build(run=False)
    soc.do_exit(vns)

    os.system("rm -rf testbench/sha2-pac")  # nuke the old PAC if it exists
    os.system("mkdir -p testbench/sha2-pac") # rebuild it from scratch every time
    os.system("cp pac-cargo-template testbench/sha2-pac/Cargo.toml")
    os.system("cd testbench/sha2-pac && svd2rust --target riscv -i ../../../../target/soc.svd && rm -rf src; form -i lib.rs -o src/; rm lib.rs")
    BiosHelper(soc, boot_from_spi) # marshals cargo to generate the BIOS from Rust files

    # pass #2 -- generate the SoC, incorporating the now-built BIOS
    platform = Platform(dutio)
    soc = Dut(platform, spiboot=boot_from_spi)

    builder = Builder(soc, output_dir="./run")
    builder.software_packages = [  # Point to a dummy Makefile, so Litex pulls in bios.bin but doesn't try building over it
        ("bios", os.path.abspath(os.path.join(os.path.dirname(__file__), "testbench")))
    ]
    vns = builder.build(run=False)
    soc.do_exit(vns)
 ```

And inside the BiosHelper the key line is:
```
        ret += os.system("riscv64-unknown-elf-objcopy -j .text -j .rodata -j .data -O binary ../../target/{}/release/{} run/software/bios/bios.bin".format(target, sim_name))
```

zyp:
Instead of relying on a dummy bios package to trick the builder to call `_initialize_rom_software()` to get your `bios.bin` pulled in, I think a more proper solution would be to add a way to explicitly tell the builder to initialize the rom from an external file rather than building the bios.

bunnie:
yah, I'd be fine with that too; but I do need the SVD file in order to build the BIOS, so the first phase has to be with no `bios.bin` at all (or maybe could just be an array with zeros, some dummy stand-in is also fine), and the other phase would actually have the final `bios.bin`.

The solution we arrived at was mainly so we didn't have to muck with the Litex core to accommodate our eclectic use case of a Rust BIOS.

zyp:
Apparently this feature already exists: https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc_core.py#L156

I guess running it with a dummy file first to generate the SVD, and then again with the actual binary to build should work.

bunnie:
Ah yep, I can use that feature to work around this. I still have to propagate this change to all our testbenches and do a full regression but at least it's not throwing errors. 

Didn't know it existed!

PR #943, opened 2021-06-15
Replace deprecated inspect.getargspec with inspect.getfullargspec.

JosephBushagour:
The [inspect.getargspec](https://docs.python.org/3/library/inspect.html#inspect.getargspec) function has been deprecated since Python 3.0. For the use case in LiteX (referencing the .args part), [inspec.getfullargspec](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) is a non-deprecated, drop-in replacement (shown by the source code [here](https://github.com/python/cpython/blob/3.9/Lib/inspect.py#L1120)).

### DIFF
--- a/litex/compat/soc_sdram.py
+++ b/litex/compat/soc_sdram.py
@@ -58,7 +58,7 @@ def soc_sdram_args(parser):
 
 def soc_sdram_argdict(args):
     r = soc_core_argdict(args)
-    for a in inspect.getargspec(SoCSDRAM.__init__).args:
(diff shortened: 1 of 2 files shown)


developandplay:
Oh yes I stumbled upon this as well.
Some arguments don't seem to work anymore due to this.
Like `--l2-size` but probably some others as well...

enjoy-digital:
Thanks @JosephBushagour for catching it! This is merged.

PR #947, opened 2021-06-17
Add options to save and print environment variables to Flash persistent storage 

MateuszKosmala:
That Merge Request allows to:
- Print the environment variable from Flash persistent storage using **flash_show_env** bios command. Without argument, command lists all available environment variables. 
- Save the environment variables to the specified memory location using **flash_save_env** bios command. Without argument in command, saves variables to the memory address based on the default FLASH_OFFSET macro.
-  Sets any number greater than 0 as the port of the tftp server using **tftp_server_port** command.

Adds to functionality:
- get_env_params function in bios/boot.c to convert environment variables(data types) to token boundaries in JSON string.
- json_token_check function in libbase/jsmn_helpers.c to check presence of given token in JSON format string.
- erase_flash_subsector function in libbase/spiflash.c to erase given subsector.
- init_env_from_flash function in bios/boot.c to initialization of our environment variables at startup.

That PR is associated with litex-hub/litex-boards#228

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -34,6 +34,8 @@
 #include <liblitesdcard/sdcard.h>
 #include <liblitesata/sata.h>
 #include <libfatfs/ff.h>
+#include <spiflash.h>
+#include <jsmn_helpers.h>
 
 /*-----------------------------------------------------------------------*/
 /* Helpers                                                               */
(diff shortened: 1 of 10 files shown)


enjoy-digital:
Thanks @MateuszKosmala, we are indeed not using the SPI Flash to store BIOS environment variables for now. Adding this has to be done carefully to allow extending it easily in the future. For such features, I would first recommend creating an issue to discuss it and allow discussing it and allow other to discuss it. I'll think about it.

I understand the general idea of the implementation, but I'm not sure to understand why `flash_read `and `flash memory_dump` are added since we already have direct access to the SPIFlash through MMAP?

MateuszKosmala:
Thanks @enjoy-digital for your reply, indeed `flash_read` and `memory_dump` are redundant in that particular case.
I agree that there is a need to discuss comprehensively that feature.

enjoy-digital:
Closing, this would need to be discussed and agreed first.

PR #948, opened 2021-06-18
Expose non-interactive mode

developandplay:
Let's GitHub Actions users specify the `interactive` mode.
Follow up to #881


enjoy-digital:

PR #952, opened 2021-06-21
build/DFUProg: Allow to specify alt interface and to not reboot

smunaut:
Some targets have multiple alt settings for DFU for different zone
of the flash. Allowing to specify which one to flash and not
rebooting immediately allows to flash several of them at once.


enjoy-digital:
Thanks @smunaut, that's merged.

PR #953, opened 2021-06-22
Blackparrot 32bit csr

### LINKED ISSUE #938: How does Litex expect 32bit reads/writes to be handled by 64bit CPUs?
Unfortunately I'm still stuck on https://github.com/enjoy-digital/litex/commit/4f82a36a when updating Blackparrot support to the most recent Litex version.

The current wishbone adapter in Blackparrot only handles 64bit reads/writes which means `sel` (corresponding to AXis `strb`) will always be 0xFF. If I do understand what is breaking https://github.com/enjoy-digital/litex/commit/4f82a36a for Blackparrot is the reliance on `sel` in https://github.com/enjoy-digital/litex/commit/75936775 which requires `sel` to be set to either 0x0F or 0xF0 when doing 32bit accesses.

However I'm not sure what Litex expects exactly from the CPUs Wishbone/AXI adapter. I also read the [discussion](https://github.com/enjoy-digital/litex/issues/544) behind https://github.com/enjoy-digital/litex/commit/4f82a36a and the [CSR bus wiki](https://github.com/enjoy-digital/litex/wiki/CSR-Bus) but when it comes to 64bit CPUs it doesn't seem very clear to me.

Maybe @gsomlo can chime as Rocket does support 32bit read/writes for CSR access and I saw he asked how Rocket handles 32bit read/writes [here](https://github.com/chipsalliance/rocket-chip/issues/2185).

developandplay:
Fixes #938

### DIFF
--- a/litex/soc/cores/cpu/blackparrot/bp_litex/bp2wb_convertor.v
+++ b/litex/soc/cores/cpu/blackparrot/bp_litex/bp2wb_convertor.v
@@ -77,7 +77,7 @@ module bp2wb_convertor
   assign mem_resp_v_o =  v_li;
   assign stb_o = (set_stb) && !stb_justgotack; 
   assign cyc_o = stb_o; 
-  assign sel_o = 8'b11111111; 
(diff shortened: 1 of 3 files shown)


enjoy-digital:

PR #957, opened 2021-06-27
Fix compile error when FST traces are enabled on macOS.

pftbest:
Compile options should be the same for all platforms.


pftbest:
Or maybe even better would be to remove this duplicated options, and just use `VM_TRACE_FST` in the code which is already defined by verilator. Also it seems `OPT_LEVEL` switch has no effect because it is overwritten by verilator later.

enjoy-digital:
Thanks @pftbest, this is merged. I'll have a closer look at your other suggestions.

PR #958, opened 2021-06-28
soc/interconnect/axi: fix valid signal in connect_to_pads for axi lite

sergachev:
The code here https://github.com/enjoy-digital/litex/blob/master/litex/soc/interconnect/axi.py#L96 is obscure enough to disable the 'valid' signal when axi_full is False:

```
        for name, width in (
            [("valid", 1)] +
            [("last",  1)] if (ch in ["w", "r"] and axi_full) else [] +
            ch.description.payload_layout):
```


enjoy-digital:
Thanks for catching this Ilia, this is merged.

PR #961, opened 2021-06-29
litex/soc/integration/builder.py: allow overriding rom contents

wsipak:
This allows you to initialize ROM with a file specified as an argument, instead of using LiteX Bios.
It can be used with `litex-boards/litex_boards/targets/digilent_arty.py --build --rom-override=./baremetal_app.bin`

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -73,6 +73,7 @@ class Builder:
 
         # BIOS Options.
         bios_options     = [],
+        rom_override     = None,
 
         # Documentation.
         generate_doc     = False):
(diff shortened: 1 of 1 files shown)


danc86:
FYI there is currently also an `--integrated-rom-init` option in `soc_core_args()` which seems to serve a similar purpose.

enjoy-digital:
Thanks @wsipak, as said by @danc86, this indeed seems similar to` --integrated-rom-init`. 

Can you do a test with:
`litex-boards/litex_boards/targets/digilent_arty.py --build --integrated-rom-init=./baremetal_app.bin` and confirms it works for you?

enjoy-digital:
@wsipak: I'm closing this since as discussed `--integrated-rom-init` should already allow you to override the ROM contents but please re-open if you have troubles using it.

PR #962, opened 2021-06-30
Spartan 6 Support

DaveRichmond:
I was trying to build for a spartan 6 board and discovered that a patch a while back added in a flag to the clocking that just doesn't exist on the S6PLL. I feel like just removing the line linking it through seems a little excessive, but I can't see anything in any of the datasheets that would suggest that there's support for this.
With that said, this change allows the s6 boards to build again. Tested with building using ISE 14.7 against the panologic g2 target.


enjoy-digital:
Thanks @DaveRichmond for catching this and fixing it. This is merged.

PR #949, opened 2021-06-19
cores/clock/ecp5: Add dynamic phase adjustment signals.

zyp:
I'd like to upstream this change by @gregdavill, extracted from https://github.com/gregdavill/DiVA-firmware/blob/main/gateware/rtl/ecp5_dynamic_pll.py.

PR #965, opened 2021-07-07
ECP5PLL: implement 4-output solver

thirtythreeforty:
Reimplement the configuration loop to allow all 4 outputs to be used by the user, if one of them is suitable for use as VCO feedback.

The new strategy is to first iterate over requested outputs to see if any of them can be used as a feedback source.  If one can, it is selected, and if no output is suitable, it attempts to instantiate one.  Once the feedback path is selected, the VCO frequency is known and it attempts to calculate the remaining outputs' settings.

In addition, this implementation now respects datasheet limits in two new ways:

- It respects the post-input-divider minimum frequency of 10MHz
- It respects the max output frequency of 400MHz for instantiated feedback outputs

I am slightly unhappy with the seemingly-repetitive for loops. However each one has slightly different semantics and I don't see a way to combine them that doesn't hinder readability.

I've also rolled in a very small unrelated change that exposes the PLL's STANDBY signal.

Comments welcome, thanks for the useful framework!

### DIFF
--- a/litex/soc/cores/clock/lattice_ecp5.py
+++ b/litex/soc/cores/clock/lattice_ecp5.py
@@ -25,6 +25,7 @@ class ECP5PLL(Module):
         self.logger.info("Creating ECP5PLL.")
         self.reset      = Signal()
         self.locked     = Signal()
+        self.stdby      = Signal()
         self.clkin_freq = None
         self.vcxo_freq  = None
         self.nclkouts   = 0
@@ -117,6 +118,7 @@ class ECP5PLL(Module):
                 ("MFG_GMCREF_SEL",         "2")],
             i_RST           = self.reset,
             i_CLKI          = self.clkin,
+            i_STDBY         = self.stdby,
             o_LOCK          = locked,
             p_FEEDBK_PATH   = "INT_OS3", # CLKOS3 reserved for feedback with div=1.
             p_CLKOS3_ENABLE = "ENABLED",
--- a/litex/soc/cores/clock/lattice_ecp5.py
+++ b/litex/soc/cores/clock/lattice_ecp5.py
@@ -30,6 +30,7 @@ class ECP5PLL(Module):
         self.stdby      = Signal()
         self.clkin_freq = None
         self.vcxo_freq  = None
+        self.dpa_en     = False
         self.nclkouts   = 0
         self.clkouts    = {}
         self.config     = {}
@@ -47,13 +48,13 @@ class ECP5PLL(Module):
         self.clkin_freq = freq
         register_clkin_log(self.logger, clkin, freq)
 
-    def create_clkout(self, cd, freq, phase=0, margin=1e-2, with_reset=True):
+    def create_clkout(self, cd, freq, phase=0, margin=1e-2, with_reset=True, uses_dpa=True):
         (clko_freq_min, clko_freq_max) = self.clko_freq_range
         assert freq >= clko_freq_min
         assert freq <= clko_freq_max
         assert self.nclkouts < self.nclkouts_max
         clkout = Signal()
-        self.clkouts[self.nclkouts] = (clkout, freq, phase, margin)
+        self.clkouts[self.nclkouts] = (clkout, freq, phase, margin, uses_dpa)
         if with_reset:
             self.specials += AsyncResetSynchronizer(cd, ~self.locked)
         self.comb += cd.clk.eq(clkout)
@@ -80,7 +81,10 @@ class ECP5PLL(Module):
             for clkfb_div in range(*self.clkfb_div_range):
                 # pick a suitable feedback clock
                 found_fb = None
-                for n, (clk, f, p, m) in sorted(self.clkouts.items()):
+                for n, (clk, f, p, m, dpa) in sorted(self.clkouts.items()):
+                    if dpa and self.dpa_en:
+                        # cannot use clocks whose phase the user will change
+                        continue
                     for d in range(*self.clko_div_range):
                         vco_freq = self.clkin_freq/clki_div*clkfb_div*d
                         clk_freq = vco_freq/d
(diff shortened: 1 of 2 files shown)


litex/soc/cores/clock/lattice_ecp5.py
```diff
@@ -12,19 +13,21 @@
...
 class ECP5PLL(Module):
-    nclkouts_max    = 3
+    nclkouts_max    = 4
     clki_div_range  = (1, 128+1)
     clkfb_div_range = (1, 128+1)
     clko_div_range  = (1, 128+1)
     clki_freq_range = (    8e6,  400e6)
     clko_freq_range = (3.125e6,  400e6)
     vco_freq_range  = (  400e6,  800e6)
+    pfd_freq_range  = (   10e6,  400e6)
```
zyp, line 23: I believe the `clki_freq_range` above is supposed to be the same as this. They should probably be combined.

zyp:
This looks good to me.

I was about to comment that the margin for the selected output should probably be set to 0 to keep any deviation from affecting other outputs before I realized that it doesn't factor into the calculation of the VCO frequency that the other outputs are derived from. Setting a large margin on outputs that doesn't need to be very specific would actually make it easier to find a valid feedback path.

One thing I'm a little bit concerned about is dynamic phase adjustment. The documentation states:

> The clock output selected as the feedback cannot use the dynamic phase adjustment feature since it will cause the PLL to unlock. For example if the FB_MODE is INT_OP or CLKOP, then there should be no phase shift on CLKOP.

It'd be good to add a way to disallow specific outputs from being selected for the feedback path.

thirtythreeforty:
> One thing I'm a little bit concerned about is dynamic phase adjustment. The documentation states:
>
> > The clock output selected as the feedback cannot use the dynamic phase adjustment feature since it will cause the PLL to unlock. For example if the FB_MODE is INT_OP or CLKOP, then there should be no phase shift on CLKOP.
>
> It'd be good to add a way to disallow specific outputs from being selected for the feedback path.

This is a great point. It was a little unclear based on their block diagram where phase shifting is performed, some things suggest it's post-output divider. However this language unambiguously disallows using such outputs.  Since the output is automatically selected, we either:

- need the user to indicate which clocks are intended to be dynamically controlled
- the clock-reuse scheme to be disabled when dynamic control is enabled

The first is probably better, since the user could still technically monkey with the feedback clock in the second scheme.

API suggestions are welcome. The current `expose_dpa()` API simply exposes the control signals; maybe a new `create_clkout` parameter?

```
def create_clkout(self, cd, freq, phase=0, margin=1e-2, with_reset=True, uses_dpa=False):
```
Then, the algorithm can only consider `uses_dpa=False` channels for feedback.

litex/soc/cores/clock/lattice_ecp5.py
thirtythreeforty: There are actually two separate specs for clki/clkfb and pfd:

Screenshot_20210707_090816

The footnote says:
>Period jitter and cycle-to-cycle jitter numbers are guaranteed for fPFD > 10 MHz. For fPFD < 10 MHz, the jitter numbers may not be met in certain conditions

This is a "soft" spec. It sounds like it will work without meeting it, however Diamond refuses to solve PLL configs where this spec is violated. I can be persuaded to leave it out.

zyp: Ah, fair. Still seems a bit unintuitive that the minimum is higher for f_PFD than f_IN, considering the former is derived from the latter through a divisor.

I can also confirm that for my sample size of one, it works at 5 MHz in practice. :)

zyp:
Sounds reasonable to me.

thirtythreeforty:
Actually, the backward-compatible way would be a `uses_dpa=True` default, and this constraint applies only if the user calls `expose_dpa()`.

zyp:
I don't think backwards compatibility for this is a big issue considering `expose_dpa()` were merged only two weeks ago in #949.

thirtythreeforty:
> I don't think backwards compatibility for this is a big issue considering expose_dpa() were added only two weeks ago in #949.

Fair enough, but I have a secondary reason for having the default be True, which is just principle of least surprise. At the instantiation site with `uses_dpa=False`, it is now clearer that the user has a contract to uphold. This implementation will refuse to solve it if the user specifies four outputs, enables DPA, and forgets to specify `uses_dpa=False` - but this didn't work before because one output was reserved, so I think this is okay.

enjoy-digital:
Thanks a lot @thirtythreeforty, @zyp  these are very interesting improvements! (the initial ECP5PLL wrapper has mostly been done to simplify clocking on the different ECP5 target and was clearly very limited). I'll try to review it more closely soon and also do tests with some ECP5 boards.

enjoy-digital:
@thirtythreeforty This also looks good to me and all the ECP5 targets from LiteX-Boards also seem to build fine so we can merge.

thirtythreeforty:
Thank you for the review!

enjoy-digital:
Hi @thirtythreeforty,

there was in fact a regression with the implementation: Verilog code was generated, but some parameters were apparently not correct. I just had a look at this and reverted the code to an approach that is closer to the one I had in the initial version (not necessarily better, but at least easier for me to maintain :)): https://github.com/enjoy-digital/litex/commit/04c2c45d79addb2ed74f0a7e1149e2d9acf4e3b3 

This now seems back to normal while testing the build on some boards from LiteX-Boards (previously I just tested the verilog code generation). Can you check the features you added are still handled correctly?

thirtythreeforty:
Hey sorry for the trouble! Thanks for fixing... for my own edification, do you have a snippet of the invalid parameters? Would like to know what went wrong.

enjoy-digital:
@thirtythreeforty: No worries, the PR has been very useful and removes the specific case for the feedback. I'm not sure to remember the exact issue, but some of the coefficients were not generated correctly since when building a target, clock constraints were not propagated correctly. If you compare the generated code with/without the last change you should be able to spot the incorrect(s) parameter(s).

Issue #966, opened 2021-07-09
Auto-generated soc.h incompatible with assembly files.


JosephBushagour:
Issue: The auto-generated `soc.h` file cannot be included in any assembly code because the access functions aren't hidden behind a pre-processor conditional. This makes conditional assembling based in things like `CONFIG_CPU_HAS_INTERRUPT` and `CONFIG_CPU_VARIANT` currently not possible.
```C
// in file asssembly.S
#include "generated/soc.h"
// results in: Error: unrecognized opcode `static inline int config_clock_frequency_read(void){'
```

Proposed change: Place access functions behind a `#ifndef LITEX_SOC_DEFINES_ONLY`. Users can then include the file in any assembly code by defining (and then undefining) `LITEX_SOC_DEFINES_ONLY`. Because this is an "opt-out" system, it won't break existing code.
```C
// in file asssembly.S
#define LITEX_SOC_DEFINES_ONLY
#include "generated/soc.h"
#undef LITEX_SOC_DEFINES_ONLY
// results in: all defines visible in the assembly
```

enjoy-digital:
Thanks @JosephBushagour for identifying and removing this limitation, your #967 PR has been merged.

PR #967, opened 2021-07-09
Make "generated/soc.h" able to be included in assembly files.

JosephBushagour:
Related Issue:  #966

This PR implements the proposed change from the related issue, this allows the `generated/soc.h` file to be safely included in any assembly files.

It does change the layout of `soc.h` from the following:
```C
//--------------------------------------------------------------------------------
// Auto-generated by Migen (3ffd64c) & LiteX (8276857c) on 2021-07-09 10:46:33
//--------------------------------------------------------------------------------
#ifndef __GENERATED_SOC_H
#define __GENERATED_SOC_H
#define CONFIG_CLOCK_FREQUENCY 24000000
static inline int config_clock_frequency_read(void) {
	return 24000000;
}
#define CONFIG_CPU_HAS_INTERRUPT
#define CONFIG_CPU_RESET_ADDR 2147680256
static inline int config_cpu_reset_addr_read(void) {
	return 2147680256;
}
...
#endif
```

to:

```C
//--------------------------------------------------------------------------------
// Auto-generated by Migen (3ffd64c) & LiteX (8276857c) on 2021-07-09 10:46:33
//--------------------------------------------------------------------------------
#ifndef __GENERATED_SOC_H
#define __GENERATED_SOC_H
#define CONFIG_CLOCK_FREQUENCY 24000000
#define CONFIG_CPU_HAS_INTERRUPT
#define CONFIG_CPU_RESET_ADDR 2147680256
...
#ifndef __ASSEMBLER__
static inline int config_clock_frequency_read(void) {
	return 24000000;
}
static inline int config_cpu_reset_addr_read(void) {
	return 2147680256;
}
...
#endif // !__ASSEMBLER__

#endif
```

I'm more than happy to update this PR with any suggestions or changes.

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -147,7 +147,7 @@ def get_mem_header(regions):
 def get_soc_header(constants, with_access_functions=True):
     r = generated_banner("//")
     r += "#ifndef __GENERATED_SOC_H\n#define __GENERATED_SOC_H\n"
-
+    funcs = ""
 
     for name, value in constants.items():
         if value is None:
@@ -161,8 +161,13 @@ def get_soc_header(constants, with_access_functions=True):
             ctype = "int"
         r += "#define "+name+" "+value+"\n"
         if with_access_functions:
-            r += "static inline "+ctype+" "+name.lower()+"_read(void) {\n"
-            r += "\treturn "+value+";\n}\n"
+            funcs += "static inline "+ctype+" "+name.lower()+"_read(void) {\n"
+            funcs += "\treturn "+value+";\n}\n"
+
+    if with_access_functions:
+        r += "\n#ifndef __ASSEMBLER__\n"
+        r += funcs
+        r += "#endif // !__ASSEMBLER__\n"
 
     r += "\n#endif\n"
     return r


geertu:
What about using
```
#ifndef __ASSEMBLY__
...
#endif
```
instead of a custom new definition?

JosephBushagour:
> What about using
> 
> ```
> #ifndef __ASSEMBLY__
> ...
> #endif
> ```
> 
> instead of a custom new definition?

You're right that seems like a better approach here :). I think `__ASSEMBLER__` is the one predefined by [GCC](https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html) but I'm pretty new to this area so if `__ASSEMBLY__` is common practice I'll go ahead and use that.

geertu:
@JosephBushagour You are right, `__ASSEMBLER__ `seems to be the standard one, cfr.
https://lore.kernel.org/lkml/9C47C740-86CF-48F1-8DB6-B547E5D098FF@mac.com/

JosephBushagour:
I've pushed the change to `__ASSEMBLER__`, this should be a better implementation. Thanks for the great tip @geertu.

PR #969, opened 2021-07-14
openfpgaloader: add cable and freq options

trabucayre:
Not all boards have an onboard cable (FTDI, DFU, CMSIS-DAP, ...) it's the case for instance for acorn CLE 215+ or colorlight 5A-75B.
In this situation `--cable` is required or default `ft2232` cable is used.
Without `--freq` the cable is configured with a 6MHz clock frequency. It's a good compromise between speed and stability but with current master branch we have demonstrate *openFPGALoader* is able with arty to work at 30MHz (max FT2232 clock frequency).

This PR add two optional constructor params to pass cable name and frequency and build cmd line accordingly.

### DIFF
--- a/litex/build/openfpgaloader.py
+++ b/litex/build/openfpgaloader.py
@@ -12,16 +12,22 @@ from litex.build.generic_programmer import GenericProgrammer
 class OpenFPGALoader(GenericProgrammer):
     needs_bitreverse = False
 
-    def __init__(self, board):
-        self.board = board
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre, this is merged. I'm currently playing with a Trenz TEC0117 and enjoying OpenFPGALoader, thanks for this very nice tool that supports very exotic FPGA boards :)

trabucayre:
Thanks you!

PR #970, opened 2021-07-17
openfpgaloader: improves freq conversion type

trabucayre:
`freq` param may be a string, an int but in all cases it's usualy easier to use engineering notation to represent big number.
This PR add a conversion to float to avoid exception like:
`
ValueError: invalid literal for int() with base 10: '10e6'
`


enjoy-digital:

PR #975, opened 2021-07-20
soc/add_sdcard: Fix IRQ code, only add IRQs depending on mode

cklarhorst:
The current code only works for the default "read+write" mode.
This fixes the gateware for the "read" and "write" mode, but the bios doesn't build for mode "write" (sdcard init depends on read).


enjoy-digital:
Good catch, thanks @cklarhorst, this is merged.

PR #977, opened 2021-07-23
GTKWSave: improved support for Endpoint

jedrzejboczar:
This is a small improvement for the GTKWave savefile generator to automatically handle Endpoint objects - signals will be sorted and colored automatically. The `.add()` method can also handle other objects from litex.soc.interconnect.stream. This will happen if an object contains sink/source attribute that is of type Endpoint, and the result will look like this:
screenshot_2021-07-23_11-20-59
This should be useful for pipelines of multiple stream modules.


enjoy-digital:

PR #979, opened 2021-07-26
WIP: enable QPI mode in liblitespi

fkokosinski:
This PR introduces a way to support QPI mode when using LiteSPI on the example of MX25L12833F. The chip-specific code is included when the SoC has the following constants defined:
```
soc.constants['FLASH_CHIP_MX25L12833F_QUAD'] = 1  # to use wp and hold as data lines
soc.constants['FLASH_CHIP_MX25L12833F_QPI'] = 1   # to also send commands using 4 lines
```

We'd probably want to put common instructions, i.e. `WREN`, into a separate header file + it'd be nice to have a separate header file per chip, but I'm not sure if that approach won't unnecessarily restrict us later.

This PR also depends on litex-hub/litespi#53.

### DIFF
--- a/litex/soc/software/liblitespi/spiflash.c
+++ b/litex/soc/software/liblitespi/spiflash.c
@@ -56,6 +56,22 @@ void spiflash_dummy_bits_setup(unsigned int dummy_bits)
 #endif
 }
 
+static void spiflash_master_write(uint32_t val, size_t len, size_t width, uint32_t mask)
+{
+	/* empty rx queue */
+	while (spiflash_mmap_master_status_rx_ready_read())
+		spiflash_mmap_master_rxtx_read();
+
+	spiflash_mmap_master_cs_write(1);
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @fkokosinski, this is integrated and I did minor changes on top of it with https://github.com/enjoy-digital/litex/commit/60c4a1b667b81c5d4d3dab6b8d63b457e05ce18c.

I gave more generic names to the constants (so you'll have to adapt your integration code) and we could imagine generating/providing the instructions to the software during the LiteX integration (as #define or header).

enjoy-digital:
@fkokosinski With https://github.com/enjoy-digital/litex/commit/ab48461d6a1933f26671d6d006af0329b81976a1 the software constants will now be automatically genrated for the software. For example when using the `MX25L12833F `module, `SPIFLASH_MODULE_QPI_CAPABLE` will be automatically defined.

PR #980, opened 2021-07-27
Update blackparrot imafd

developandplay:
BlackParrot's latest version brings support for floating point operations


enjoy-digital:

PR #988, opened 2021-08-10
Fix stack alignment

### LINKED ISSUE #322: printf formatting for 64bit integers is broken on 32bit cpus

michalsieron:
RISC-V requires stack to be aligned to 16 bytes.
https://github.com/riscv/riscv-elf-psabi-doc/blob/1d5384e6690ea39b5395ddc6943882ed2830dbee/riscv-elf.md?plain=1#L183

Right now, in bios/linker.ld, `_fstack` is being set to 8 bytes before the end of sram region.
```
PROVIDE(_fstack = ORIGIN(sram) + LENGTH(sram) - 8);
```

Removing ` - 8` makes it aligned to 16.

Also there are changes in crt0.S for vexriscv, vexriscv_smp and cv32e40p.

Code that was setting up stack, was adding 4 to its address for some reason.

Removing it makes it aligned to 8 bytes, and with change in bios/linker.ld to 16 bytes.

It also fixes `printf` with long long integers on 32bit CPUs ([relevant issue](https://github.com/riscv/riscv-gcc/issues/63)).

Should close #322 and #910

### DIFF
--- a/litex/soc/cores/cpu/cv32e40p/crt0.S
+++ b/litex/soc/cores/cpu/cv32e40p/crt0.S
@@ -90,7 +90,7 @@ trap_entry:
 
 
 crt_init:
-  la sp, _fstack + 4
+  la sp, _fstack
   la a0, vector_table
   csrw mtvec, a0
 
--- a/litex/soc/cores/cpu/vexriscv/crt0.S
+++ b/litex/soc/cores/cpu/vexriscv/crt0.S
@@ -54,7 +54,7 @@ trap_entry:
 
 
 crt_init:
-  la sp, _fstack + 4
+  la sp, _fstack
   la a0, trap_entry
   csrw mtvec, a0
 
--- a/litex/soc/cores/cpu/vexriscv_smp/crt0.S
+++ b/litex/soc/cores/cpu/vexriscv_smp/crt0.S
@@ -58,7 +58,7 @@ trap_entry:
   .text
 
 crt_init:
-  la sp, _fstack + 4
+  la sp, _fstack
   la a0, trap_entry
   csrw mtvec, a0
   sw x0, smp_lottery_lock, a1
--- a/litex/soc/software/bios/linker.ld
+++ b/litex/soc/software/bios/linker.ld
@@ -85,7 +85,7 @@ SECTIONS
 	}
 }
 
-PROVIDE(_fstack = ORIGIN(sram) + LENGTH(sram) - 8);
+PROVIDE(_fstack = ORIGIN(sram) + LENGTH(sram));
 
 PROVIDE(_fdata_rom = LOADADDR(.data));
 PROVIDE(_edata_rom = LOADADDR(.data) + SIZEOF(.data));


enjoy-digital:
Thanks @michalsieron, good catch! Before merging we should just make sure this is not breaking LM32/OR1K and that it indeed fixes #322 and #910. I'm not able to do it before the end of the month but if someone has time to do it beforme and confirm it, we could merge it (cc @kgugala, @mateusz-holenko, @gsomlo, @geertu)

Dolu1990:
I have seen by the past that some of the standard c library functions where relying on the alignement of the stack.

enjoy-digital:
Thanks for the feedback @danc86, @Dolu1990 . The modified linker is not breaking LM32/OR1K, so we can merge. I'll apply the changes to the other RISC-V CPUs.

enjoy-digital:
This has been updated on Rocket/BlackParrot with https://github.com/enjoy-digital/litex/commit/e3a88c324e91ee819efd36ca7f2845a28480edd2.

PR #992, opened 2021-08-11
cores/vexriscv_smp: Fix vexriscv_smp doesn't build without a memory bus

cklarhorst:
The current code only works with a memory bus because otherwise
"generate_cluster_name" doesn't get called.
Cluster_name is only needed in the finalize phase.
Therefore, the name will now be generated just before its usage.

Verifiable with: 
litex_sim --cpu-type vexriscv_smp (should be broken before this commit)


enjoy-digital:
Thanks @cklarhorst, this is merged.

PR #993, opened 2021-08-11
Add Renode platform/script and Zephyr DT overlay generators

mateusz-holenko:
This PR adds scripts used to dynamically generate:
* platform definition and simulation script for Renode,
* DT overlay and configuration for Zephyr

from the LiteX configuration `csr.json` file.

Those were hosted originally in https://github.com/litex-hub/litex-renode, but having them directly in LiteX will help people discover and use them easier.


enjoy-digital:
Thanks @mateusz-holenko, I'm fine having these scripts directly in LiteX, but to simplify maintenance of https://github.com/litex-hub/litex-renode, are you going to remove them from there? Maybe we could also create Wiki page in LiteX with the contents of the litex-renode README?

mateusz-holenko:
Thanks, @enjoy-digital!

I plan to drop `litex-hub/renode` completely in favor of `enjoy-digital/litex`, but I need to update all repos referencing scripts first. This means that there will be a window when both repos will contain copies of scripts.

I can surely also transform (and update) the litex-renode README content to a LiteX Wiki page.

enjoy-digital:
Thanks for confirming, that's what I was thinking but just wanted to be sure. So we could:
- Merge your PR.
- Add a Wiki page to run Renode (and maintain only this to simplify things), same for Zephyr.
- Add a link in the main README.md to point to these Wiki pages to allow users to find it more easily.

We could also expose these scripts directly as it's done with the other tools to simplify things for the user.

Just regarding the naming, we also currently have a `litex_json2dts.py` script that is used for Linux. What about renaming the different scripts to:
- litex_json2linuxdts.py (instead of litex_json2dts.py).
- litex_json2renode.py  (instead of litex_renode_generator.py).
- litex_json2zephyrdts.py (instead of litex_zephyr_dts_generator.py)

?

mithro:
How about?
* `litex_json2dts_linux.py` (instead of `litex_json2dts.py`).
* `litex_json2dts_zephyr.py` (instead of `litex_zephyr_dts_generator.py`)
* `litex_json2renode.py` (instead of `litex_renode_generator.py`).

enjoy-digital:
@mithro: Thanks, that's also an alternative I was thinking of, so if you have a preference for it, I'm fine with it too.

mithro:
@enjoy-digital - I think that makes the most sense. We should also look at eventually merging `litex_json2dts_linux.py` and `litex_json2dts_zephyer.py` so that the device tree just works with both.

mithro:
(Getting both files into LiteX would be the first steps towards doing this merge.)

enjoy-digital:
@mithro: Sure, this would be nice to be able to have a common dts generator, we can keep two generator for now and we'll try to merge them when the similarities/differences will be clearly identified.

enjoy-digital:
This has been merged and the scripts have been renamed as discussed with https://github.com/enjoy-digital/litex/commit/dceed5f7e795f3d5b964a743f8b8d1b4d88d5f1d. @mateusz-holenko You can probably use https://github.com/enjoy-digital/litex/wiki/Run-Zephyr-On-Your-SoC to describe how to run Zephyr with LiteX. Feel also free to create a similar Wiki page for Renode.

geertu:
> @enjoy-digital - I think that makes the most sense. We should also look at eventually merging `litex_json2dts_linux.py` and `litex_json2dts_zephyer.py` so that the device tree just works with both.

Exactly. Why do we have two different scripts in the first place? DT describes hardware, not software policy, so they should be identical.

mithro:
@geertu - It's a case of "theory verse practice"
> _In theory_ - theory and practice is identical, _in practice_ that isn't true.

mateusz-holenko:
Thanks for merging. I'll update documentation soon.

mateusz-holenko:
FYI: Added a wiki page describing the purpose and use of the litex_json2renode.py script: https://github.com/enjoy-digital/litex/wiki/Generate-Renode-simulation

enjoy-digital:
Great, thanks @mateusz-holenko.

mateusz-holenko:
Created a basic https://github.com/enjoy-digital/litex/wiki/Run-Zephyr-On-Your-SoC wiki page with some basic commands and references to the litex-buildenv Wiki for details.

enjoy-digital:
Thanks @mateusz-holenko! I'm not yet familiar with Renode/Zephyr and will try to follow your wiki pages soon and will provide feedback.

Issue #876, opened 2021-04-09
litex/vexriscv boot problem


andreapas-ap:
Hello, I am working on a custom board based on Artix xc7a100tcsg324-1, Litex/vexriscv, code flash memory and DDR RAM. System is properly working, but in some rare builds I am getting the following behavior (from debug UART):

``` Booting from flash...
Copying 0x90500000 to 0x40000000 (59241 bytes)...
[########################################]
Executing booted program at 0x40000000

--============= Liftoff! ===============--
```

Nothing wrong about that, but unluckily user software is not starting at all, as something wrong is happening just after BIOS liftoff.
Additional note: I have a static initialized variable in my C code, something like ```static uint8_t my_var = 1;```. If I change it to uninitialized, so ```static uint8_t my_var;```, then boot problem disappears. No problem at all also if I change variable from initialized uint8_t to initialized uint32_t, while initialized uint16_t bevavior is same as for initialized uint8_t. 

Thanks in advance for the help

Andrea

[EDIT] just noticed many similarities to #856: changing variable from initialized to uninitialized is actually moving it from .DATA to .BSS, and this solves the issue. Anyway, I would still not understand why changing from initialized uint8_t to initialized uint32_t is also working

enjoy-digital:
Hello @andreapas-ap,

thanks for the feedback. The issue could be similar to #856 but could also be related to an issue with the DRAM calibration. Do you  notice changes in the calibration logs between builds that are working and build that aren't? Could you eventually share the calibration logs for both cases?

andreapas-ap:
Hi, here both console logs:

- first from firmware defining uint8_t initialized static variable, user software not running after boot:

```        __   _ __      _  __
       / /  (_) /____ | |/_/
      / /__/ / __/ -_)>  <
     /____/_/\__/\__/_/|_|
   Build your hardware, easily!

 (c) Copyright 2012-2020 Enjoy-Digital
 (c) Copyright 2007-2015 M-Labs

 BIOS built on Apr  9 2021 11:11:13
 BIOS CRC passed (743fc585)

 Migen git sha1: --------
 LiteX git sha1: --------

--=============== SoC ==================--
CPU:        VexRiscv @ 100MHz
BUS:        WISHBONE 32-bit @ 4GiB
CSR:        8-bit data
ROM:        1024KiB
SRAM:       8KiB
L2:     8KiB
SDRAM:      262144KiB 16-bit @ 800Mbps/pin

--========== Initialization ============--
[...]

PR #1007, opened 2021-08-18
Fix alignments in demo/linker.ld

### LINKED ISSUE #856: demo hangs on boot if it contains more than 16 bytes of .data
In the demo application, when defining a global array like this:
```
#define CRASHME 1
uint8_t some_random_data[] = {
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f,
	0xd0, 0x0f
#if CRASHME
	,0xd0
#endif
};
```
Then the application will hang on boot if CRASHME == 1 and successfully run, if CRASHME == 0.
In other words, if there are more than 16 bytes of global data allocated, the application will fail to boot (freeze at liftoff).
Reproduction: DECA board with `integrated_main_ram_size = 0x4000`

michalsieron:
Without this change, when `.data` section size wasn't multiple of word size, `data_loop` in crt0 was jumping over `_edata` and continued looping. As it works on words and right now 64 bit CPUs are biggest ones supported - alignment is now 8 bytes.

Also removed `- 4` from stack address, as it needs to be aligned to 16 bytes on RISC-V.

Should fix #856 and at least explain why #876 wasn't working.

### DIFF
--- a/litex/soc/software/demo/linker.ld
+++ b/litex/soc/software/demo/linker.ld
@@ -16,27 +16,29 @@ SECTIONS
 
 	.rodata :
 	{
-		. = ALIGN(4);
+		. = ALIGN(8);
 		_frodata = .;
 		*(.rodata .rodata.* .gnu.linkonce.r.*)
 		*(.rodata1)
+		. = ALIGN(8);
 		_erodata = .;
 	} > main_ram
 
 	.data :
 	{
-		. = ALIGN(4);
+		. = ALIGN(8);
 		_fdata = .;
 		*(.data .data.* .gnu.linkonce.d.*)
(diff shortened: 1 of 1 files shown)


michalsieron:
One possibility why sometimes OS won't boot may be that `_start` wasn't at the beginning of `main_ram`. Last commit fixes it.

enjoy-digital:
Thanks @michalsieron, good catch. This is merged.

hansfbaier:
@michalsieron Awesome, thanks!

PR #1008, opened 2021-08-19
Depacketizer: properly handle last_be overflow onto next bus word

lschuermann:
While the Depacketizer did correctly calculate a new last_be value for the data with the header removed, it may happen that the last_be overflows and thus relates to the current, non-delayed sink value. This introduces logic in form of a simple FSM to handle these cases and properly output last_be on the last valid bus word.


enjoy-digital:
Thanks @lschuermann, would you mind adding a simple test to reproduce the issue to https://github.com/enjoy-digital/litex/blob/master/test/test_packet.py? This would allow me to review more carefully the PR and would also ensure we catch future potential regressions on this in the future.

lschuermann:
Absolutely. I currently don't have access to my LiteX working environment, but should be back next week latest.

enjoy-digital:
Perfect, thanks.

lschuermann:
Actually these changes here don't really work yet as shown by the tests, still working on it. Marking as draft.

lschuermann:
Alright, the changes pushed now seem to work reliably and properly handle `last_be` for various data widths including 8 bit. As promised I will be adding test cases as well. A big thanks to @david-sawatzke for helping me debug these changes.

enjoy-digital:
Thanks @lschuermann, this looks fine. As discussed together and since this modules can have non-trivial edge cases, I'll be able to review it more carefully and merge it with a test exhibiting the issue with the previous version of the code.

gregdavill:
Just some user feedback, this is working nicely for me on ECP5 hardware.

I've got a `HyperRAM > DMA > FIFO > Packetizer > LiteEth Hybrid MAC` pipeline all running at 32bit dw.
Where the Paketizer is responsible for forming UDP packets in hardware.

Thanks!

lschuermann:
@gregdavill awesome, thanks! You wouldn't believe me how many hours of staring at GTKWave this took :smile:. In the version you've tested there still was a bug where the Packetizer/Depacketizer didn't handle transactions of one (partial) bus word, but that presumably doesn't occur with ARP/ICMP/your UDP protocol.

I have now rewritten the `test_packet.py` file to test for various combinations of data widths, with enabled or disabled `last_be`. This also rewrites `test_stream.py` as it introduces a new, extremely generic `stream_inserter` and `stream_collector` infrastructure, which translates between byte-list Packets with optional parameters to LiteX streams. This should be a significant improvement over the manual tests written before, as this is a quite solid way to mass-inject and read from LiteX streams. It also takes care of the very tricky part of setting and respecting `last_be` -- if so desired -- and properly deals with ready/valid states.

@enjoy-digital The new `test_stream.py` could be very useful for larger integration tests with the LiteEth MAC pipeline. Do you think we can export it in LiteX such that it's usable in LiteEth?

This new test infrastructure revealed countless edge-cases in the Packetizer/Depacketizer, some of which even occur in the last_be-less implementation of upstream, if I'm not mistaken. However, a large number of packets is required to fuzz all edge-cases, I usually test with approx. 1024 packets locally. I have observed errors in the 261th packet, that goes to show how complex these things are. :smile: 

@enjoy-digital I'd appreciate if we can get this reviewed quickly, as much of my and @david-sawatzke's LiteEth work uses these components, including the `last_be` functionality. Executing these tests on the old version of the Packetizer/Depacketizer will block indefinitely because there are some issues with transactions <= a single bus word, but I think that's fine for now and also expected to fail. One can think about a check whether any signals other than the sys_clk change to abort tests in the future.

test/test_packet.py
```diff
@@ -53,77 +67,81 @@ def loopback_test(self, dw):
...
+        # will then fill up these values with garbage again. Thus we
+        # also have to remove the proper amount of bytes from the sent
+        # packets so the comparison will work.
+        if not with_last_be and dw != 8:
+            for (packet, recvd_packet) in zip(packets, recvd_packets):
+                invalid_recvd_bytes = packet_header_length % (dw // 8)
+                recvd_packet.data = recvd_packet.data[:-invalid_recvd_bytes]
+                packet.data = packet.data[:len(recvd_packet.data)]
+
+        self.assertTrue(compare_packets(packets, recvd_packets))
```
lschuermann, line 123: One thing to note is that the Packetizer, without any `last_be` signal, will throw away the last partial bus word. For instance, when sending `0xDEAD BEEF BADC 0DED` over a 32-bit data path with a 3-byte header of `0xABC`, the following packet will be sent: `0xABCD EADB EEFB ADC0`. This can be quite unintuitive. The Depacketizer will then restore the `0xABC` header, producing the data `0xDEAD BEEF BADC 0XXX` where X are unrelated signal states.

Whether this should change (so that all data plus some trailing garbage would be encoded) is a different question, but the above logic handles these cases accordingly.

test/test_packet.py
```diff
@@ -12,39 +12,53 @@
...
-    swap_field_bytes = True)
-
-def packet_description(dw):
-    param_layout = packet_header.get_layout()
-    payload_layout = [("data", dw)]
-    return EndpointDescription(payload_layout, param_layout)
-
-def raw_description(dw):
-    payload_layout = [("data", dw)]
-    return EndpointDescription(payload_layout)
```
lschuermann, line 35: All of this is moved into the test method as keeping it being a global variable caused the tests to fail depending on the order in which they were executed. Presumably some of these variables keep some test-modified state, but I didn't investigate this further.

gregdavill:
> @gregdavill awesome, thanks! You wouldn't believe me how many ~hours~ days of staring at GTKWave this took :smile:. In the version you've tested there still was a bug where the Packetizer/Depacketizer didn't handle transactions of one (partial) bus word, but that presumably doesn't occur with ARP/ICMP/your UDP protocol.

That's correct. I'm merging a header and payload that both aren't fully aligned to 32bits. I've been able to construct my data such that I avoid a lot of the more nuanced edge cases, like a header/payload less than dw.

test/test_packet.py
lschuermann: Adding randomly generated header fields to the test suite revealed that this statement is not true unconditionally. If the `header_leftover` (number of bytes of the header which are interleaved with data in a single bus word) plus the number of bytes in the last aligned data word is below or equal to the number of bytes in a bus word, all data (plus some trailing garbage) will be transmitted. Example:

Header: `0xABCD E`
Data: `0xBADD EADD AD`
Packetized: `0xABCD EBAD DEAD DADx` (where x is are unrelated bytes)
Depacketized: Header `0xABCD E`, Data `0xBADD EADD ADxx` (where x are unrelated bytes)

Hence I've adjusted the test logic.

gregdavill: Just an idea,. Instead of trying to guess/fake the last_be signal, could this generate a warning/error if the last_be signal isn't present? That way the user can determine the behavior they want?

lschuermann: I think that, yes, it might be reasonable to refuse operation without a `last_be` signal with dw > 8. However, that would be a breaking change and I think it'd be preferable to have this in its own PR. I also don't know whether any subsystems rely on the current behavior -- which this PR doesn't change, except for 1-2 bug fixes.

Also, I really don't have much more time for this, and given it does now work correctly as far as I can tell I don't see a reason per sé to deliberate disable otherwise functional code paths. :smile: Generating a warning does sound very reasonable though, @enjoy-digital what do you think?

enjoy-digital:
Thanks @lschuermann, I just had a first look at the code and think most the complexity comes from a bad architecture for the handling of the unaligned-data (added for 10GBE and already there when you started looking at the code) and the use of `last_be` instead of a more simple `be` data qualifier. I'll spend some time next week looking at it and try to improve these points to reduce the complexity. Sorry for providing my input just now, I was a busy the last weeks with other things.

lschuermann:
Thanks for the quick review @enjoy-digital. I appreciate trying to reduce complexity, however these changes do seem to work with the way `last_be` works currently. If the redesign will take a while, can we maybe get this merged in the meantime? It presumably wouldn't break anything and is an improvement over the current state.

Also, if you're thinking about an entirely different data qualifier, I'd be happy to brainstorm on that (and also what that would mean for the changes already done in LiteEth). Unfortunately, I really only have time getting the 10G data path working for the next month or so, and so I'd rather have something working which can be improved later on, and I think it'd be better for us and others if we can get these changes into mainline LiteX instead of trying to maintain a fork.

enjoy-digital:
@lschuermann: I'll have time next week to look at it, sorry I need to be careful on such PRs. I also have a preference to just merge the Depacketizer/Packetizer changes without the stream/packet sim changes that we could eventually discuss later (I also have plans on this and already deprecated some old sim code: https://github.com/enjoy-digital/litex/blob/master/litex/compat/stream_sim.py)  I'm well aware you spent quite some time on this and I'm sorry not being able to just merge the PR, but this part of the code is important for the various cores and I want to be sure to make things correctly.

lschuermann:
Absolutely. I wasn't trying to pressure you into merging, sorry if my comment read this way. I'm by now well aware this requires careful review and has a large impact. I was just a little wary of throwing all over board in favor of a complete redesign and/or switch to some different data qualifier. 😀

enjoy-digital:
@lschuermann: No worries, but yes I'm trying to be careful on these things now, because I have to maintain them in the long term :) In fact the idea I have in mind is to use the same data qualifier are use on PCIe TLP, at least in the Packetizer/Depacketizer used in LiteX. In LitePCIe, we already have Packetizer/Depacketizer:
- https://github.com/enjoy-digital/litepcie/blob/master/litepcie/tlp/packetizer.py
- https://github.com/enjoy-digital/litepcie/blob/master/litepcie/tlp/depacketizer.py

But as you can see, the code is hardcoded for each datapath width (because the first version has been created _before_ the Depacketizer/Packetizer in LiteX). So I had plan to switch LitePCIe to LiteX's Depacketizer/Packetizer, but this requires proper handling of un-aligned data/data qualifier, so I this probably a good occasion to do it. Note that we could probably easily add a function to the code to easily switch from regular `be` qualifier to `last_be`.

Happy to discuss all this while I'll be looking at this.

PR #1012, opened 2021-08-24
json2dts_linux: update ethmac [tx|rx]-fifo-depth

gsomlo:
The LiteETH driver currently under review for upstream Linux expects
total byte values for [tx|rx]-fifo-depth. Instead, the linux-specific
json2dts generator currently sets those values to the number of
respective [tx|rx] slots.

Modify json2dts_linux to multiply slots by ethmac_slot_size, in order
to generate the correct values expected by the Linux driver.


gsomlo:
PS. I rebased [`litex-rebase`](https://github.com/litex-hub/linux/tree/litex-rebase) on top of @shenki's upstream proposed driver (currently under review on lkml), with a much simplified fixup after-patch (which mostly just adds back in Antony Pavlov's polling mode).

As such, this PR is currently required for the latest litex-rebase to work with `json2dts_linux`.

mithro (commented):
Should the DT include information about the slot size?

gsomlo:
On Tue, Aug 24, 2021 at 01:39:50PM -0700, Tim Ansell wrote:
> Should the DT include information about the slot size?

At the moment, it's hard-coded to 0x800 in the Linux driver side.

After some LiteX archaeology (and grepping), we seem to come by it via:

  liteeth/mac/__init__.py: self.slot_size = CSRConstant(2**bits_for(eth_mtu))

Since eth_mtu is set to 1530, rounding up to the next power of 2 we
end up with 2048.

Whether we should add a DTS property for the Linux driver to pick up
dynamically, or whether we should rather leave well enough alone and
assume it'll stay put as a constant on both sides of the gateware/kernel
divide, I'll leave that decision up to Joel... :)

Cheers,
--Gabe

mithro:
Well, there seems to be a high likelihood of wanting to support things like jumbo frames? Also I think we do still need the number of slots?

litex/tools/litex_json2dts_linux.py
```diff
@@ -285,8 +285,8 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
     ethmac_csr_base  = d["csr_bases"]["ethmac"],
     ethmac_mem_base  = d["memories"]["ethmac"]["base"],
     ethmac_mem_size  = d["memories"]["ethmac"]["size"],
-    ethmac_tx_slots  = d["constants"]["ethmac_tx_slots"],
-    ethmac_rx_slots  = d["constants"]["ethmac_rx_slots"],
+    ethmac_tx_depth  = d["constants"]["ethmac_tx_slots"] * d["constants"]["ethmac_slot_size"],
```
shenki, line 288: This shouldn't be the slot size, but the size of the fifo. So for the default case where we have two slots, the fifo-depth should be 0x1000.

shenki (changes requested):

shenki:
> Well, there seems to be a high likelihood of wanting to support things like jumbo frames? 

Jumbo frames will be dropped by the hardware as it stands. It will reject any packet over eth_mtu, which is defined as 1530.

enjoy-digital:
@shenki, @gsomlo: Thanks for upstreaming the LiteEth driver. I'm not sure to understand why 1) we are passing total tx-depth and rx-depth instead of 2) number of tx-slots,  rx-slots and the slot size? Is it because it's something standardized in Linux for Ethernet devices? With 2) we would just pass raw LiteEth's parameters to the driver and would just have to reuse them directly in the driver instead of re-generating them locally with fixed constants for the slot size. This would also cover the case of future possible Jumbo frames where LiteEth would be configured with a larger slot size (that would be then directly passed with the .dts).

(Maybe I'm missing one point that has already been discussed sorry if so).

shenki:
> @shenki, @gsomlo: Thanks for upstreaming the LiteEth driver. I'm not sure to understand why 1) we are passing total tx-depth and rx-depth instead of 2) number of tx-slots, rx-slots and the slot size? Is it because it's something standardized in Linux for Ethernet devices? 

I agree that this would be better. I went with the rx-fifo-depth as they were standard device tree properties.

>With 2) we would just pass raw LiteEth's parameters to the driver and would just have to reuse them directly in the driver instead of re-generating them locally with fixed constants for the slot size. This would also cover the case of future possible Jumbo frames where LiteEth would be configured with a larger slot size (that would be then directly passed with the .dts).

I sent an email off list to @gsomlo with a similar conclusion to you. It would be better for us if we described the fifo depth and slot size in the bindings. I will propose this in v3 of the patches.

shenki:
v3 of the patches can be seen here: https://lore.kernel.org/netdev/20210825124655.3104348-1-joel@jms.id.au/

From the bindings example:

```
    mac: ethernet@8020000 {
        compatible = "litex,liteeth";
        reg = <0x8021000 0x100>,
              <0x8020800 0x100>,
              <0x8030000 0x2000>;
        reg-names = "mac", "mdio", "buffer";
        litex,rx-slots = <2>;
        litex,tx-slots = <2>;
        litex,slot-size = <0x800>;
        interrupts = <0x11 0x1>;
    }
```

If you're using the default settings in litex, this is the minimal device tree description you will need:

```
    mac: ethernet@8020000 {
        compatible = "litex,liteeth";
        reg = <0x8021000 0x100>,
              <0x8020800 0x100>,
              <0x8030000 0x2000>;
        reg-names = "mac", "mdio", "buffer";
        interrupts = <0x11 0x1>;
   }
```

This is because the slots numbers and size defaults to the settings we use in litex.

enjoy-digital:
@shenki: Thanks, LGTM.

gsomlo:
I'll update the PR to matchi this later today -- hopefully it passes
upstream linux review, since this does indeed make the most sense as
far as what first-order information *should* be passed between
gateware and kernel.

PR #986, opened 2021-08-04
soc/software/liblitedram: fix pattern checking for low DFI databits

jedrzejboczar:
This a fix described at the end of https://github.com/enjoy-digital/litedram/pull/258#issuecomment-882435501.

To my understanding this should be correct, but please check if you think it is. I've tested this on xilinx_zcu104, digilent_arty and antmicro_lpddr4_test_board and it worked correctly on these.

I also used the following script to verify the error counts:
```python
#!/usr/bin/env python

VERBOSE = False

def check_test_pattern_popcounts(module, *, nphases, dfii_pix_data_bytes, nmodules):
    per_phase = []
    for p in range(nphases):
        this_phase = []
        for i in range(dfii_pix_data_bytes):
            j = p * dfii_pix_data_bytes + i
            if j % nmodules == (nmodules - 1 - module):
                this_phase.append(i)
        per_phase.append(this_phase)
    # max 8 per popcount as we always compare unsigned char
    max_errors = sum(8*len(popcounts) for popcounts in per_phase)
    # calculated analytically
    xdr = 2
    max_errors_calculated = 8 * nphases * xdr
    return per_phase, max_errors, max_errors_calculated

for nphases in [1, 2, 4, 8]:
    for nmodules in [1, 2, 4, 8]:
        xdr = 2
        pix_databits = nmodules*8 * xdr
        pix_databytes = pix_databits//8
        for module in range(nmodules):
            per_phase, max_errors, max_errors_calculated = check_test_pattern_popcounts(
                module, nphases=nphases, dfii_pix_data_bytes=pix_databytes, nmodules=nmodules)

            variables = f'{nphases=} {pix_databytes=:2} {nmodules=} {module=} {max_errors=:4} {max_errors_calculated=:4}'
            if max_errors != max_errors_calculated:
                print(f'!MISMATCH! {variables}')
            else:
                print(f' ok        {variables}')
            if VERBOSE:
                # lines = [f'    prs[{p}][{i}] ^ tst[{i}]' for (p, i) in enumerate(per_phase)]
                lines = []
                for p, phase in enumerate(per_phase):
                    for i in phase:
                        lines.append(f'    prs[{p}][{i}] ^ tst[{i}]')
                print('\n'.join(lines))
```
which yields:
```
 ok        nphases=1 pix_databytes= 2 nmodules=1 module=0 max_errors=  16 max_errors_calculated=  16
[...]

PR #1016, opened 2021-08-30
soc/software/liblitedram: fix max error count computation

jedrzejboczar:
There is an error in READ_CHECK_TEST_PATTERN_MAX_ERRORS computation. It is being computed based on nphases and data rate (x1/x2), but this gives incorrect result when using DFIRateConverter which increases number of data bytes per phase. This may results in integer underflow when subtracting `(max_errors - errors)` and selecting wrong bitslip during read leveling.

With this change we make use of DFII_PIX_DATA_BYTES directly, which should be more reliable as sdram_write_read_check_test_pattern should always scan 1/SDRAM_PHY_MODULES of all data bytes on DFI. So max errors is nphases (outer for loop) times dfi data bytes (inner loop) divided by nmodules (scanning 1 module) times 8 bits in a byte ([relevant code](https://github.com/antmicro/litex/blob/f943092bb5a3aae1bfda08e2ffbe927d048e32be/litex/soc/software/liblitedram/sdram.c#L345-L356)).

I tested it with the script from https://github.com/enjoy-digital/litex/pull/986 and it doesn't modify the results for these cases. I also successfully tested on hardware (Arty and ZCU104).

### DIFF
--- a/litex/soc/software/liblitedram/sdram.c
+++ b/litex/soc/software/liblitedram/sdram.c
@@ -300,7 +300,7 @@ static void print_scan_errors(unsigned int errors) {
 #endif
 }
 
-#define READ_CHECK_TEST_PATTERN_MAX_ERRORS (8*SDRAM_PHY_PHASES*SDRAM_PHY_XDR)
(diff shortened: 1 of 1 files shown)


enjoy-digital:

PR #1017, opened 2021-08-30
customize main ram size from command line argument

asadaleem-rs:
litex_sim.py is edited to customize main ram size from command line arguments, if main_ram_size is set to zero then it won't show in generated SoC logs.


enjoy-digital:
Thanks @asadaleem-rs, this is merged. I did other improvements on top of it to make increase similarities between LiteX-Boards targets and litex_sim with https://github.com/enjoy-digital/litex/commit/7c50f52a575269efc4676ac5fa5998c533cb50c6.

PR #1019, opened 2021-09-01
Add docs static files to package_data in setup.py

G33KatWork:
This patch adds the static files needed for the documentation to the
`package_data` field in `setup.py`.

I ran into failures when generating documentation after installing litex
using pipenv which didn't copy these files which in turn caused the
documentation generation to fail.

This change causes all files in the `static` subfolder of the
litex.soc.doc module to be installed as well.


enjoy-digital:
Thanks @G33KatWork! That's merged.

PR #1022, opened 2021-09-01
Restructure config flags for dcache/icache presence in Vex.

tcal-x:
This change uses new explicit config flags CONFIG_CPU_NO_DCACHE and CONFIG_CPU_NO_ICACHE to control whether special instructions are emitted to flush these caches when present.

This makes it more convenient when adding new VexRiscv variants without one or both of the caches.   For example, currently we need to lie and set CONFIG_CPU_VARIANT_LITE to suppress the dcache flush instruction.

This change centralizes knowledge about the different variants into vexriscv/core.py.


enjoy-digital:
Thanks @tcal-x, this is merged, I just did minor changes with https://github.com/enjoy-digital/litex/commit/e257d91d46e8b750eabae596ec7adb506e7514e8 to try to keep consistency with the other software config flags.

PR #1030, opened 2021-09-09
Fixes to allow crossover uart over PCIe with lxterm and litex_server

teknoman117:
Add a command line option for lxterm to allow specification of the path to the CSR register CSV file, rather than just requiring csr.csv to be in the directory lxterm is executed from

Add a command line option to specify the base address of whatever addresses lxterm is sending to lxserver. In the case of PCIe, these are negative values (e.g. -0xF000_0000)

If the base address was not specified and the design uses PCIe, automatically remap the base address to the CSR start

```
# Generate bitstream for SQRL Acorn
./sqrl_acorn.py --with-pcie --build --driver --uart-name=crossover --csr-csv=build/sqrl_acorn/csr.csv

# flash the board using your programmer

# connect to crossover UART
litex_server --pcie --pcie-bar 48:00.0 &
lxterm --csr-csv=build/sqrl_acorn/csr.csv bridge
```

### DIFF
--- a/litex/tools/litex_term.py
+++ b/litex/tools/litex_term.py
@@ -87,8 +87,8 @@ else:
 from litex import RemoteClient
 
 class BridgeUART:
-    def __init__(self, name="uart_xover", host="localhost", base_address=0): # FIXME: add command line arguments
-        self.bus = RemoteClient(host=host, base_address=base_address)
+    def __init__(self, name="uart_xover", host="localhost", base_address=None, csr_csv=None): # FIXME: add command line arguments
+        self.bus = RemoteClient(host=host, base_address=base_address, csr_csv=csr_csv)
         present = False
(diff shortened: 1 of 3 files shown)


litex/tools/litex_term.py
```diff
@@ -97,6 +97,10 @@ def __init__(self, name="uart_xover", host="localhost", base_address=0): # FIXME
         if not present:
             raise ValueError(f"CrossoverUART {name} not present in design.")
 
+        # On PCIe designs, CSR is remapped to 0 to limit BAR0 size.
```
teknoman117, line 100: I'm not sure what the default for this should be. Technically, it's also entirely possible for a uart to wishbone bridge to exist as well as a PCIe phy (maybe without CSR mapping). It's also possible someone changed the size of the BAR so that the entire memory map is visible via a BAR.

PR #1032, opened 2021-09-12
Json2dts sdcard booting

stffrdhrn:
This is a series of 3 patches to allow setting up the DTS for booting from an sdcard.
  1. We want to be able to disable the initrd loading from ram.  This also was pointed out by @shenki as being something that keeps him from using the dts generator as it means we cannot use kernels with built in rootfs. (option `--without-initrd`)
  2. We want to be able to set a different root device, for my SD card I use `mmcblk0p3` (option `--root-device mmcblk0p3`)
  3. I setup earlycon to use liteuart rather than the risc-v only `sbi`, no option, this is now the hard coded default.  This means we get console output right away rather than waiting for the console drive to be initialized (usually after loading initrd).

### DIFF
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -10,9 +10,9 @@
 import sys
 import json
 import argparse
+import os
 
-
-def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
+def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, polling=False):
 
     kB = 1024
     mB = kB*1024
@@ -44,18 +44,32 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
     if initrd_size is None:
         initrd_size = default_initrd_size
 
+    if initrd == "enabled" or initrd is None:
+        initrd_enabled = True
+    elif initrd == "disabled":
+        initrd_enabled = False
+    else:
+        initrd_enabled = True
+        initrd_size = os.path.getsize(initrd)
+
     dts += """
         chosen {{
-            bootargs = "mem={main_ram_size_mb}M@0x{main_ram_base:x} rootwait console=liteuart earlycon=sbi root=/dev/ram0 init=/sbin/init swiotlb=32";
-            linux,initrd-start = <0x{linux_initrd_start:x}>;
-            linux,initrd-end   = <0x{linux_initrd_end:x}>;
-        }};
-""".format(
+            bootargs = "mem={main_ram_size_mb}M@0x{main_ram_base:x} {console} {rootfs} init=/sbin/init swiotlb=32";""".format(
     main_ram_base      = d["memories"]["main_ram"]["base"],
-    main_ram_size      = d["memories"]["main_ram"]["size"],
     main_ram_size_mb   = d["memories"]["main_ram"]["size"] // mB,
-    linux_initrd_start = d["memories"]["main_ram"]["base"] + initrd_start,
-    linux_initrd_end   = d["memories"]["main_ram"]["base"] + initrd_start + initrd_size)
+    console            = "console=liteuart earlycon=sbi",
+    rootfs             = "rootwait root=/dev/ram0")
+
+    if initrd_enabled is True:
+        dts += """
+            linux,initrd-start = <0x{linux_initrd_start:x}>;
+            linux,initrd-end   = <0x{linux_initrd_end:x}>;""".format(
+        linux_initrd_start = d["memories"]["main_ram"]["base"] + initrd_start,
+        linux_initrd_end   = d["memories"]["main_ram"]["base"] + initrd_start + initrd_size)
+
+    dts += """
+        };
+"""
 
     # CPU ------------------------------------------------------------------------------------------
 
@@ -624,10 +638,12 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
     return dts
 
 def main():
+
     parser = argparse.ArgumentParser(description="LiteX's CSR JSON to Linux DTS generator")
     parser.add_argument("csr_json", help="CSR JSON file")
     parser.add_argument("--initrd-start", type=int,            help="Location of initrd in RAM (relative, default depends on CPU)")
     parser.add_argument("--initrd-size",  type=int,            help="Size of initrd (default=8MB)")
+    parser.add_argument("--initrd",       type=str,            help="Supports arguments 'enabled', 'disabled' or a file name. Set to 'disabled' if you use a kernel built in rootfs or have your rootfs on an SD card partition. If a file name is provied the size of the file will be used instead of --initrd-size. (default=enabled)")
     parser.add_argument("--polling",      action="store_true", help="Force polling mode on peripherals")
     args = parser.parse_args()
 
@@ -635,6 +651,7 @@ def main():
     r = generate_dts(d,
         initrd_start = args.initrd_start,
         initrd_size  = args.initrd_size,
+        initrd       = args.initrd,
         polling      = args.polling,
     )
     print(r)
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -57,7 +57,7 @@ def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, polling=Fa
             bootargs = "mem={main_ram_size_mb}M@0x{main_ram_base:x} {console} {rootfs} init=/sbin/init swiotlb=32";""".format(
     main_ram_base      = d["memories"]["main_ram"]["base"],
     main_ram_size_mb   = d["memories"]["main_ram"]["size"] // mB,
-    console            = "console=liteuart earlycon=sbi",
+    console            = "console=liteuart earlycon=liteuart,0x{:x}".format(d["csr_bases"]["uart"]),
     rootfs             = "rootwait root=/dev/ram0")
 
     if initrd_enabled is True:
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -12,7 +12,7 @@ import json
 import argparse
 import os
 
-def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, polling=False):
+def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, root_device=None, polling=False):
 
     kB = 1024
     mB = kB*1024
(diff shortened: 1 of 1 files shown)


litex/tools/litex_json2dts_linux.py
```diff
@@ -624,17 +633,27 @@ def add_clkout(clkout_nr, clk_f, clk_p, clk_dn, clk_dd, clk_margin, clk_margin_e
     return dts
 
 def main():
+    initrd_enabled = True
+
     parser = argparse.ArgumentParser(description="LiteX's CSR JSON to Linux DTS generator")
     parser.add_argument("csr_json", help="CSR JSON file")
     parser.add_argument("--initrd-start", type=int,            help="Location of initrd in RAM (relative, default depends on CPU)")
     parser.add_argument("--initrd-size",  type=int,            help="Size of initrd (default=8MB)")
```
mithro, line 641: https://docs.python.org/3/library/argparse.html#:~:text=same%20interface.%20The-,BooleanOptionalAction,-is%20available%20in
> You may also specify an arbitrary action by passing an Action subclass or other object that implements the same interface. The BooleanOptionalAction is available in argparse and adds support for boolean actions such as --foo and --no-foo:
> ```
> >>>
> >>> import argparse
> >>> parser = argparse.ArgumentParser()
> >>> parser.add_argument('--foo', action=argparse.BooleanOptionalAction)
> >>> parser.parse_args(['--no-foo'])
> Namespace(foo=False)
> ```

mithro:
I wonder if you should be able to give something like `disabled`, `internal` or a filename reference to an `--initrd` argument?

mithro:
I agree that the default option should be the litex serial device which is cross architecture rather than SBI (which only works on RISC-V). Maybe we do want to support something like `--serial-mode=(litex|sbi)`?

litex/tools/litex_json2dts_linux.py
stffrdhrn: Right, I saw that was available, bit I went with this implementation add it's what I found in existing litex code.  I'll change it though.

stffrdhrn:
> I wonder if you should be able to give something like `disabled`, `internal` or a filename reference to an `--initrd` argument?

I like this option, it means we can replace `--initrd-size` with `--initrd path/to/rootfs.cpio` the generator can calculate the size for us. 
I don't think there is a difference between `disabled` and `internal`.  

Currently with my patches we would have:

  - Internal: `--without-initrd`
  - SD card: `--without-initrd  --root-device mmcblk0p3`
  - Initrd: `--initrd-size 8123456`  (provide the right size to load the .cpio image if bigger than the default)

One option **1** is we have, an all encompassing argument like:

  - Internal: `--rootfs internal`
  - SD card: `--rootfs device,mmcblk0p3`
  - Initrd: `--rootfs initrd,path/to/rootfs.cpio`   (file path used to get size)

Though the parsed arguments are a bit confusing.

Another option **2** is using --initrd, it could work like this

  - Internal: `--initrd disabled`
  - SD card: `--initrd disabled --root-device mmcblk0p3` (note:root device is default `ram0` for internal and initrd rootfs)
  - Initrd: `--initrd path/to/rootfs.cpio`   (file path used to get size, alternative is still `--initrd-size 8123456`)

I like option **2**

stffrdhrn:
I have implemented the `--initrd` argument as per option **2** described above.  For earlycon I am not sure what would be a good name for it `--serial-mode` is not really right as this is only earlycon all devices use the liteuart serial console as per the `console=liteuart` param.  So the option could be `--earlycon-mode` but that is maybe too specific.

On the other hand...

The reason to use earlycon=liteuart is for the case where litex bios boots linux (isn't this most of the time?).  The litex bios uses lituart as per [libbase](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/uart.c).   Earlycon works because the liteuart device is already initialized by litex bios so linux can just write to the registers without doing a driver probe, that is what earlycon assumes.  That assumption should be true if we are booting linux from litex bios.

So I am wondering if the `earlycon=sbi` argument actually ever worked.  Similarly, I wonder if `swiotlb=32` does anything.

geertu:
> So I am wondering if the `earlycon=sbi` argument actually ever worked. Similarly, I wonder if `swiotlb=32` does anything.

I have been wondering that myself, too, when I noticed CONFIG_HVC_RISCV_SBI and CONFIG_SERIAL_EARLYCON_RISCV_SBI depend on RISCV_SBI_V01.

litex/tools/litex_json2dts_linux.py
```diff
@@ -44,18 +44,35 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
     if initrd_size is None:
         initrd_size = default_initrd_size
 
+    if initrd == "enabled" or initrd is None:
```
mateusz-holenko, line 47: Doesn't `initrd == None` intuitively mean "disabled"?

mateusz-holenko: Perhaps `initrd_size` should have priority - meaning that if it's set then `initrd_enabled` is set to `True`?

litex/tools/litex_json2dts_linux.py
```diff
@@ -44,18 +44,35 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
...
+        initrd_enabled = True
+    elif initrd == "disabled":
```
mateusz-holenko, line 49: Not a great fan of using `enabled` and `disabled` as special values.

I guess it won't be a very common scenario but still those **could** be a valid initrd file names.

litex/tools/litex_json2dts_linux.py
```diff
@@ -44,18 +44,35 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
...
+        initrd_enabled = False
+    else:
+        initrd_enabled = True
+        initrd_size = os.path.getsize(initrd)
+
+    if root_device is None:
+        root_device = "ram0"
```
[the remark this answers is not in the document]
stffrdhrn: I don't think it works, I tried it, if we call `generate_dts()` like below then without providing `--root_device` the argument passed is None.

```
   generate_dts(d,
         initrd_start = args.initrd_start,
         initrd_size  = args.initrd_size,
         initrd       = args.initrd,
         root_device  = args.root_device,
         polling      = args.polling,
     )
```

litex/tools/litex_json2dts_linux.py
stffrdhrn: Currently we have:
  - No args - means `--initrd enabled --initrd_start 8mb --initrd_size 8mb`
  - `--initrd disabled` - no initrd
  - `--initrd filename.xyz` means - `--initrd enabled --initrd_start 8mb --initrd_size sizeof(filename.xyz)`

I didn't want to change the defaults but we could change it to have initrd disabled by default i.e.:
  - No args -> Initrd disabled
  - `--initrd_size 8mb` -> initrd enabled
  - `--initrd filname.xyz` -> initrd enabled and `--initrd_size sizeof(filename.xyz)`

litex/tools/litex_json2dts_linux.py
stffrdhrn: That right, but this suggested, it seems pretty reasonable.

litex/tools/litex_json2dts_linux.py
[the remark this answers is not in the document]
mateusz-holenko: ok, that was a minor remark; if the suggested way doesn't work, let's keep it as it is right now

litex/tools/litex_json2dts_linux.py
mateusz-holenko: I understand the argument about keeping the defaults.

I guess @enjoy-digital has a better intuition how destructive changing defaults/signature would be.

litex/tools/litex_json2dts_linux.py
mateusz-holenko: Sure, I don't have a nice solution to this myself. Having said that I don't think this is a blocker.

litex/tools/litex_json2dts_linux.py
shenki, line 56: I don't think there's any value in providing this - the kernel will try to boot from the initrd as /dev/ram0 if you provide it.

litex/tools/litex_json2dts_linux.py
```diff
@@ -44,18 +44,35 @@ def generate_dts(d, initrd_start=None, initrd_size=None, polling=False):
...
+
     dts += """
         chosen {{
-            bootargs = "mem={main_ram_size_mb}M@0x{main_ram_base:x} rootwait console=liteuart earlycon=sbi root=/dev/ram0 init=/sbin/init swiotlb=32";
-            linux,initrd-start = <0x{linux_initrd_start:x}>;
-            linux,initrd-end   = <0x{linux_initrd_end:x}>;
-        }};
-""".format(
+            bootargs = "mem={main_ram_size_mb}M@0x{main_ram_base:x} {console} {rootfs} init=/sbin/init swiotlb=32";""".format(
```
shenki, line 60: Does anyone know why the `mem=` parameter is specified? Unless you're restricting the kernel to less RAM than it expects, I don't think we need it.

The device tree describes the size and location of RAM.

shenki, line 60: There's no need to specify `init=/sbin/init`, the kernel will try it first:

https://elixir.bootlin.com/linux/v5.14/source/init/main.c#L1543
```
        if (!try_to_run_init_process("/sbin/init") ||
            !try_to_run_init_process("/etc/init") ||
            !try_to_run_init_process("/bin/init") ||
            !try_to_run_init_process("/bin/sh"))
                return 0;
```

shenki, line 60: Unless we have a bug, I don't think we need to force the number of IO TLB slabs (`swiotlb=32`). If it is required, we should open a bug against the CPU that needs it and document it here.

shenki (approved):
A great cleanup! I think we can remove more things, but I'll leave it to you as to weather you do that in this PR or a future one.

litex/tools/litex_json2dts_linux.py
stffrdhrn: I was thinking the same thing for both `mem=` and `init` I dont think its needed.

stffrdhrn: Yes, I don't think its needed.  But I am not running the risc-v setup to know what they are doing.

stffrdhrn:
In terms of most of these `mem=` `init=`, `swiotlb=`, `root=`, they all originate from commit fafa844aa.  I am thinking that the were copied from some dts file somewhere without much thought.  But I can't find the source, the initial commit was by @mateusz-holenko , but I think @mithro had an earlier version.

I was just trying to maintain them for now for backward compatibility in case some arch wanted/needed them.  But I am happy to remove them. (I dont have the setup to test all arch's though).

enjoy-digital:
Thanks @stffrdhrn, this looks good. `mem=` `init=`, `swiotlb=`, `root=`in fact originate from this: 
https://github.com/litex-hub/linux-on-litex-vexriscv/commit/526a6840c8f1e1e022bb643ddb90fbf07b1fe89f#diff-108c71c3be2eb44308485e75f91b380f71bad247ef1ca5172e06dbcdab5d1d7d and were maybe already not required at time but we were just using a "known working" config to integrate the other parts and probably kept this since then.... so we can probably remove them, I'll do a test on hardware tomorrow and will merge.

enjoy-digital:
@stffrdhrn @shenki: This is merged and `mem=`, `init=`, `swiotlb= ` bootargs have been removed by https://github.com/enjoy-digital/litex/commit/84f1afd6d4fbd2643d9b71e40f7974ffcc5c208d. Thanks!

PR #1033, opened 2021-09-13
Linker fix for initialized global variables

caverar:
I found that in some cases, initialized global variables don't work with user libraries, so a little change to the linker that I use, taken from the demo file, seems to solve the problem. I think it makes more sense to put the global variables in sram and their initial values in the main_ram, similar to the bios linker.


enjoy-digital (approved):
Thanks @caverar, this makes sense.

PR #1035, opened 2021-09-13
litex_sim: optionally add GPIOTristate core

lschuermann:
This changeset encompasses two distinct commits:

- litex/soc/cores/gpio: support external tristate buffer

  Support exposing tristate GPIOs with tristate pads, by avoiding instantiation of tristate buffers directly in the module. This gives the developers more flexibility in how they want to implement their tristate IOs (for example with level shifters behind the IOs), and allows to use the GPIOTristate core in the Verilated simulation as Verilator does not support top-level inout signals.

- litex_sim: add optional GPIOTristate core

  Adds a switch `--with-gpio`, which will add a 32 pin GPIOTristate core, with the GPIOTristate signals exposed on the top-level module. This can be used to add a custom GPIO module in the Verilated simulation.

Both of these changes combined have allowed me to write a GPIO Verilator module, controlled via ZeroMQ, which external (CI) tools can use to interact with the simulation as if it were a typical development board. The plan is to use this for a CI workflow in Tock OS (tock/tock#2801). I will create a PR for the ZeroMQ/JSON-based simulation control module later.

The first changeset is also required by boards such as the NetFPGA-SUME, which have bidirectional level shifters attached to their PMOD GPIOs. By exposing the raw tristate signals, implementations can attach such chips and be transparent about the level shifters and other hardware details to software.

### DIFF
--- a/litex/soc/cores/gpio.py
+++ b/litex/soc/cores/gpio.py
@@ -73,20 +73,35 @@ class GPIOInOut(Module):
 
 class GPIOTristate(_GPIOIRQ, Module, AutoCSR):
     def __init__(self, pads, with_irq=False):
-        assert isinstance(pads, Signal)
-        nbits     = len(pads)
-        self._oe  = CSRStorage(nbits, description="GPIO Tristate(s) Control.")
-        self._in  = CSRStatus(nbits,  description="GPIO Input(s) Status.")
-        self._out = CSRStorage(nbits, description="GPIO Ouptut(s) Control.")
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @lschuermann, this looks good and has been merged.

lschuermann:
Awesome, thank you @enjoy-digital!

enjoy-digital:
@lschuermann: I just did a minor simplification on top of it: https://github.com/enjoy-digital/litex/commit/1e24fd87d1dc899ec059154e168a09421cb87dd6.

PR #1037, opened 2021-09-15
Fix premature selection of full PLL config with no feedback

thirtythreeforty:
Two changes here:

- clock/lattice_ecp5/ECP5PLL: ensure feedback path selected before exiting search

The test case I added is the one I originally hit. The loop would initially select clkfb=1, see that all 4 outputs were correct, and exit the search. Then it would try to generate a feedback clock and fail because all were already allocated.

Fix this by adding the feedback path to the search success criteria.

- clock/lattice_ecp5/ECP5PLL: emit frequency annotations to help Diamond

Unlike nextpnr, Diamond appears not to infer the frequency of PLL outputs.  Emit the same attributes that Diamond's PLL tool does.


enjoy-digital:
Thanks @thirtythreeforty, this looks good and has been merged.

PR #1040, opened 2021-09-21
Rocket SMP support

gsomlo:
Works, ~~but needs a bit of clean-up and streamlining before being pulled.~~

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -60,15 +60,15 @@ GCC_FLAGS = {
     "full":     "-march=rv64imafdc -mabi=lp64 ",
 }
 
-# AXI Data-Widths ----------------------------------------------------------------------------------
+# CPU Size Params ----------------------------------------------------------------------------------
 
-AXI_DATA_WIDTHS = {
-    # Variant : (mem, mmio)
-    "standard": ( 64,  64),
-    "linux":    ( 64,  64),
-    "linuxd":   (128,  64),
-    "linuxq":   (256,  64),
-    "full":     ( 64,  64),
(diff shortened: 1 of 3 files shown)


litex/soc/cores/cpu/rocket/core.py
```diff
@@ -141,7 +141,11 @@ def __init__(self, platform, variant="standard"):
             i_reset = ResetSignal("sys") | self.reset,
 
             # Debug (ignored).
-            i_resetctrl_hartIsInReset_0           = Open(),
+            #FIXME: dynamically add one of these for each core!!!
+            #i_resetctrl_hartIsInReset_0           = Open(),
+            #i_resetctrl_hartIsInReset_1           = Open(),
+            #i_resetctrl_hartIsInReset_2           = Open(),
+            #i_resetctrl_hartIsInReset_3           = Open(),
```
gsomlo, line 148: @enjoy-digital: the rest of the rocket verilog interface stays the same, except for additional per-core instances of `i_resetctrl_hartIsInReset_X`.

Now, originally, this signal was commented out and left open *implicitly* until commit 116c2f154 where it was made explicit by being assigned with `Open()`.

What's a good way to handle a variable number of these signals (one per core)? Maybe something like using `self.cpu_params.update()` from a loop (`for i in range(num_cores)`)? Or simply commenting them back out and leaving them all open implicitly again?

gsomlo: I fixed this in commit 3cb89bc by dynamically generating one signal per core. LMK what you think.

gsomlo:
I think this is ready to merge (subject to review, of course).

enjoy-digital:
Thanks @gsomlo, this looks good. We can merge it.

PR #1041, opened 2021-09-22
cpu/vexriscv_smp/crt0.S: only boot core should run data_init

gsomlo:
Also, no need to for non-boot cores to `call smp_slave`, it's the
immediately following instruction for them already.

***NOT*** tested, only following up on an email conversation to ensure we don't forget. @Dolu1990 @enjoy-digital please test and review -- thanks!


Dolu1990:
Thanks :D

Dolu1990:
Works well for me.
@enjoy-digital  I guess it can be merged

enjoy-digital:
Thanks @gsomlo, @Dolu1990.

PR #1043, opened 2021-09-27
cpu/rocket/core: Remove ResetInserter on adapters.

enjoy-digital:
Previously, the SoCController was only reseting the CPU, which required adding
these ResetInserters. Now that the SoCController resets both CPU and peripherals
these ResetInserters are redundant and no longer useful.

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -294,12 +294,10 @@ class RocketRV64(CPU):
         self.cpu_params.update({'i_resetctrl_hartIsInReset_%s'%i : Open() for i in range(num_cores)})
 
         # Adapt AXI interfaces to Wishbone.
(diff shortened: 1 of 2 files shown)


gsomlo:
looks fine (and worked fine when I tested it on nexys4ddr: writing to the reset register from the bios resulted in a well-behaved reset, and so did hitting the designated hardware reset button).

One question: should https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L1337 also be addressed as part of this change?

enjoy-digital:
Thanks @gsomlo for the feedback. I was indeed almost sure there was something another location for this :) We should indeed also remove it. I updated the PR, can you do a test with it?

gsomlo:
On Mon, Sep 27, 2021 at 06:47:24AM -0700, enjoy-digital wrote:
> I updated the PR, can you do a test with it?

Tested with linuxd (128-bit mem width) on nexys4ddr (64-bit litedram
width), still works fine. I think we're good to go!

enjoy-digital:
Thanks @gsomlo, that's merged.

PR #1053, opened 2021-09-30
Add 16-bits, RGB565 FB support in simple-framebuffer

rdolbeau:
This adds support for a 16-bits (RGB565) version of the framebuffer, the intent is to save on DRAM bandwidth (although that's probably only true if the DRAM interface is 16 bits or lower, as the code doesn't load multiple pixels at once if the memory interface is wider than the pixel depth).


rdolbeau:
my bad, the code does load multiple pixels at once, it's 'hidden' in the stream.Converter if I understand correctly.

rdolbeau:
force-pushed a missing bits in the DTS, the reserved memory was still 4 bytes per pixel.

enjoy-digital:
@rdolbeau: I just changed `depth `parameter to `format`. This should be more explicit and will allow supporting more video formats in the future. I hope it's fine for you?

rdolbeau:
Fine with me :-)
(but those two are the only two supported by Linux' simple-framebuffer, so other formats would need a different DTS entry / driver to get Linux support ; other OSes - or lack thereof ! - may differ of course)

PR #1054, opened 2021-09-30
FIX: arg name is changed.

niw:
**Problem**

In commit 01a906add7cddcb63bc93813f30e3fd39336eec6, argument name is changed from `sector` to `block`, but not changed the one used in the code.

**Solution**

Use `block` instead.


enjoy-digital:
Sorry a copy/paste error from sdcard.c to spisdcard.c, Thanks!

niw:
Thank you so much for quick reply...!!!

PR #1055, opened 2021-10-01
64-bit follow-up for picolibc warning fixes

gsomlo:
Providing "uint32_t" to printf's "%ld" results in warnings on 64-bit
builds: use "unsigned long" instead.

This is a fixup on top of 77283d3d.


enjoy-digital:
Thanks @gsomlo!

Issue #1049, opened 2021-09-28
picolibc: Rocket support broken?


enjoy-digital:
Reported by @gsomlo: https://libera.irclog.whitequark.org/litex/2021-09-27#30904583

cc @kgugala, @michalsieron.

enjoy-digital:
The issue seems related to Atomics support: Changing the architecture from `rv64imac` to r`v64imc` [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/rocket/core.py#L59) makes it work.

It can be reproduced by: `litex_sim --threads 4 --opt-level Ofast --cpu-type rocket`. The UART RX IRQ are behaving correctly but it seems that the call to picolibc's `getchar`() never returns.

enjoy-digital:
https://github.com/enjoy-digital/litex/commit/a588c3b830d92c2d0525a0489c5e6b87ec25e883 fixes the issue (at least provides a workaround).

gsomlo:
While a588c3b does fix it in the simulator, on the FPGA I still get no console output at all. Using the nexys4ddr as an example:

```
litex-boards/litex_boards/targets/digilent_nexys4ddr.py --build --cpu-type rocket --cpu-variant linux --sys-clk-freq 50e6 --with-ethernet --with-sdcard
```

enjoy-digital:
As discussed together @gsomlo, the issue can be reproduced in simulation when enabling ethernet:
````
litex_sim --threads 4 --opt-level Ofast --cpu-type rocket --integrated-main-ram-size=0x10000 (work)
litex_sim --threads 4 --opt-level Ofast --cpu-type rocket --integrated-main-ram-size=0x10000 --with-ethernet  (no longer work).
````
Removing `-flto` [here](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/common.mak#L56) fix the `--with-ethernet` case but break the other...

gsomlo:
Removing `-flto` allows the FPGA deployment (on nexys4ddr, with *both* ethernet *and* sdcard) to start, load linux, and work as well as before the picolibc switch-over.

It also results in no output whatsoever on the FPGA system's console when built without ethernet and sdcard (just the cpu) -- same tradeoff as shown by @enjoy-digital in simulation.

enjoy-digital:
Thanks @gsomlo, I'll try to understand the issue tomorrow.

kgugala:
[...]

PR #1057, opened 2021-10-04
cpu/rocket: naturally align data defined in crt0.S

j-piecuch:
This issue came up while investigating #1049. This PR does not completely fix the problem described there, but it's a step in the right direction.
The startup code accesses this data using `sd/ld` instructions, which require that the address being accessed is 8-byte aligned.
The `.dword` asm directive does NOT imply any alignment, so we need to force it using the `.align` directive.

### DIFF
--- a/litex/soc/cores/cpu/rocket/crt0.S
+++ b/litex/soc/cores/cpu/rocket/crt0.S
@@ -112,11 +112,14 @@ inf_loop:
   j inf_loop
 
 .bss
+  .align 8
 smp_ap_args:
   .dword 0
   .dword 0
   .dword 0
+  .align 8
 smp_ap_target:
   .dword 0
+  .align 8
 smp_ap_ready:
   .dword 0


gsomlo:
> This PR does not completely fix the problem described there, but it's a step in the right direction.

So I tested this both in the simulator, with and without ethernet:

```
litex/litex/tools/litex_sim.py --threads 4 --opt-level Ofast --cpu-type rocket \
        --with-ethernet --integrated-main-ram-size=0x10000
...
litex/litex/tools/litex_sim.py --threads 4 --opt-level Ofast --cpu-type rocket
```

and on the `nexys4ddr`, also with just the rocket cpu and no peripherals on one hand, and both ethernet and sdcard on the other:

```
litex-boards/litex_boards/targets/digilent_nexys4ddr.py --build \
        --cpu-type rocket --cpu-variant linux4 --sys-clk-freq 50e6
...
litex-boards/litex_boards/targets/digilent_nexys4ddr.py --build \
        --cpu-type rocket --cpu-variant linux4 --sys-clk-freq 50e6 \
        --with-ethernet --with-sdcard
```

Worked fine in all 4 cases. @j-piecuch -- I'd be curious what part of #1049 this does *not* fix (besides the inability to accept input at the `litex >` prompt which was already addressed by @enjoy-digital in commit a588c3b)? AFAICT, this takes care of the `-flto` portion, so we should apply it -- thanks for figuring it out!

j-piecuch:
@gsomlo correct, this PR does not fix the inability to accept input. I have some updates on that issue too, I will post an update on #1049.

enjoy-digital:
Thanks @j-piecuch! This is merged.

PR #1061, opened 2021-10-08
Update icestorm.py with u4k device, since Yosys can target it

osterwood:


enjoy-digital:
Thanks @osterwood, this is merged.

PR #1065, opened 2021-10-12
microwatt: Fix family property

shenki:
In commit 061b89beffde ("cpu/picolibc: Add family property to CPUs and
directly use it for picolibc.") a family was added for meson cross
compilation, but this doesn't exist, leading to the following warning:

 WARNING: Unknown CPU family powerpc, please report this at https://github.com/mesonbuild/meson/issues/new

Instead use ppc64. While this seems wrong for a ppc64le machine, it
appears to be what meson expects.


enjoy-digital:
Thanks @shenki, this is merged.

PR #1069, opened 2021-10-16
efinix: don't hardcode timing model

trabucayre:
Current efinix support have an hardcoded 'C4' timing model. But this one is not available for all trion devices.
This PR add extraction of device and timing model from information provided by platform in litex-boards (*T120F324I4* instead of *T120F214* for instance)


enjoy-digital:

PR #1071, opened 2021-10-18
tools: Add moc server json generator

mateusz-holenko:
This moves the moc server generator by @CarlFK from litex-hub/litex-renode.

The moc server generator is related to the HDMI2USB project and is used to feed HW data to the web UI served by https://github.com/CarlFK/HDMI2USB-litex-firmware/blob/master/mocserver/moc-server.py without the need for the hdmi2usb side to exist.


enjoy-digital:
Thanks @mateusz-holenko, is this script only used by HDMI2USB project or is it also used by generic LiteX/Renode simulation?

enjoy-digital:
Sorry, the purpose seems to be specific to HDMI2USB project. I'm not sure this is useful to directly include the code here. This could probably be maintained externally. If we have a real need for a similar feature in the future we could reconsider it.

PR #1074, opened 2021-10-20
Use ibex main repo

### LINKED ISSUE #1073: Use Ibex from the main repo
Hi @enjoy-digital ,

Do you think it would be a good idea to use the Ibex CPU from [lowRISC/ibex](https://github.com/lowRISC/ibex), for which we already have the [pythondata-cpu-ibex](https://github.com/litex-hub/pythondata-cpu-ibex) package which tracks it, instead of the one from opentitan repo. I have already tried this in [my branch](https://github.com/navan93/litex/tree/use-ibex-main-repo). 
If you think this is useful I will open a PR.

Thanks
Navaneeth

navan93:
This PR includes 2 things:
1. Fix the need for a patch to add include to `ibex_top.sv` by changing the order in which the source files are added
2. Change Ibex to use files from pythondata-cpu-ibex package instead of taking from the pythondata-misc-opentitan

Should close #1073

### DIFF
--- a/litex/soc/cores/cpu/ibex/core.py
+++ b/litex/soc/cores/cpu/ibex/core.py
@@ -190,6 +190,19 @@ class Ibex(CPU):
     @staticmethod
     def add_sources(platform):
         ibexdir = get_data_mod("cpu", "ibex").data_location
+        platform.add_verilog_include_path(os.path.join(ibexdir, "rtl"))
+        platform.add_verilog_include_path(os.path.join(ibexdir,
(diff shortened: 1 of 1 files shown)


navan93:
@enjoy-digital With the changes in this PR only the [pythondata-cpu-ibex](https://github.com/litex-hub/pythondata-cpu-ibex) files are enough to run just Ibex.

navan93:
No problem :)

PR #1080, opened 2021-10-25
XilinxVivadoToolchain: support copying dependencies to the build dir

lschuermann:
This introduces a flag called `copy_deps` which, when specified, attempts to properly copy all required dependencies for the Xilinx Vivado build into the target build directory. This is particularly useful for shipping the generated Verilog and not have it reference system dependencies, or to integrate with scripts for building the bitstreams on a remote machine.

This mechanism is by no means perfect, there might still be absolute paths pointed to by included Tcl scripts, etc. However it's a good start and the proper way to handle dependencies such as the pythondata-cpu-* files. With these changes, developers can simply adjust their own scripts to also only have relative paths and it'll likely just work.

Signed-off-by: Leon Schuermann <leon@is.currently.online>

### DIFF
--- a/litex/build/xilinx/vivado.py
+++ b/litex/build/xilinx/vivado.py
@@ -8,7 +8,7 @@ import os
 import subprocess
 import sys
 import math
-from shutil import which
+from shutil import which, copyfile
 
 from migen.fhdl.structure import _Fragment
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Closing, this should be common feature applied on all build backends. Sorry @lschuermann and thanks for the PoC on Vivado, this will still be useful when we'll implement this for all build backends.

PR #1087, opened 2021-10-26
Add JTAG support on ECP5

gregdavill:
Following @gsomlo's initial efforts in #797. I've tidied it up and got JTAG support on ECP5 under the trellis toolchain.

In litex.build.openocd the `chain` variable determines which value is loaded into IR
```
"irscan $_CHIPNAME.tap {:d}".format(0x1 + chain),
```
So when calling litex_term/litex_server ensure chain is set to: (0x32 - 1) = 49. 
Curious if you have a better idea on how to handle this? Maybe an `--ecp5` option if '--jtag` flag is set, which sets chain internally to 0x32?

## Usage
```
# jtag uart
python litex_boards/targets/gsd_butterstick.py --uart-name=jtag_uart --build --load
litex_term --jtag-config ../prog/openocd_ecp5_ft232.cfg --jtag-chain 49 jtag

# jtagbone + crossover
# Add 'self.add_jtagbone()' to soc
python litex_boards/targets/gsd_butterstick.py --uart-name=crossover --csr-csv csr.csv --build --load
litex_server --jtag-config ../prog/openocd_ecp5_ft232.cfg --jtag-chain 49 --jtag
litex_term bridge
```

Here is `prog/openocd_ecp5_ft232.cfg` for reference
```
# ButterStick OpenOCD config

adapter driver ftdi
ftdi_vid_pid 0x0403 0x6014
ftdi_channel 0
# just TCK TDI TDO TMS, no reset
ftdi_layout_init 0xfff8 0xfffb
reset_config none

ftdi_tdo_sample_edge rising

# default speed
adapter speed 10000
transport select jtag

set _CHIPNAME ecp5

# ECP5 device - LFE5UM5G-85F
jtag newtap $_CHIPNAME tap -irlen 8 -expected-id 0x81113043
```

## Background
ECP5's `JTAGG` module doesn't provide RAW access to JTAG pins, rather it handles the `TMS` signal internally, passes through `TDI`, `TCK`, and gives us some control signals if IR register is loaded with either of these 2 private instructions `0x32` and `0x38`. Here it's just set-up using 0x32. 

It's important to note that TDI pin presented to user logic is registered with TCK in hard-logic. Trellis doesn't really understand the timing constraint here, so we simply pass TDI through a synchronizing FF based on sys-clk. This works so long as sys-clk is ~5x faster than JTAG_TCK. 

In my testing 10MHz TCK, with 60MHz sysclk works. But 15MHz with 60MHz did not (missing data).

### DIFF
--- a/litex/soc/cores/jtag.py
+++ b/litex/soc/cores/jtag.py
@@ -4,10 +4,11 @@
 # Copyright (c) 2019 Florent Kermarrec <florent@enjoy-digital.fr>
 # Copyright (c) 2019 Antti Lukats <antti.lukats@gmail.com>
 # Copyright (c) 2017 Robert Jordens <jordens@gmail.com>
+# Copyright (c) 2021 Gergory Davill <greg.davill@gmail.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
 from migen import *
-from migen.genlib.cdc import AsyncResetSynchronizer
+from migen.genlib.cdc import AsyncResetSynchronizer, MultiReg
 
 from litex.soc.interconnect import stream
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
@gregdavill: Just for info, with https://github.com/enjoy-digital/litex/commit/cb13f71b5611e5f4c66ad2e4cc9f4e743ffaa4b4 the `--jtag-chain=49` is no longer required (It was a bit difficult to remember and was also generating too much questions/support :))

josuah:
This still works flawlessly as of today!

One small OpenOCD config file adjustment to do, I'll open a PR for it.

And then I used this config file for the Lattice ECP5 EVN:

`./openocd.cfg` (OpenOCD master at commit `7023deb`)
```
adapter driver ftdi
adapter speed 25000
transport select jtag
ftdi vid_pid 0x0403 0x6010
ftdi channel 0
ftdi layout_init 0x00e8 0x60eb
reset_config none

jtag newtap ecp5 tap -irlen 8 -expected-id 0x81113043
```

And ran it like this with LiteX at commit `ff67781f` and LiteX boards at commit `2d3b81a`:

```
./lattice_ecp5_evn.py --uart-name=jtag_uart --cpu-type=serv --csr-csv=csr.csv --build --load
litex_term --jtag-config=openocd.cfg jtag
```

And tada!
2023-07-30-112950_914x581_scrot

For anyone trying, while this should not be required for JTAG I suppose, note that I did the modding of the board as advertised by the logs:
```
FT2232H will be used as serial, make sure that:
 -the hardware has been modified: R22 and R23 should be removed, two 0 Ω resistors shoud be populated on R34 and R35.
 -the chip is configured as UART with virtual COM on port B (With FTProg or https://github.com/trabucayre/fixFT2232_ecp5evn).
```

For anyone landing there from search engines and trying, it seems like it helped to first unplug the power and USB from the board, then plug it, and immediately running the build/load/litex_term without anything else in-between (poking at the ports).

A lot of thanks to everyone involved!

PR #1093, opened 2021-10-30
Add option in Makefile for (s)ccache support.

cr1901:
This also prevents meson from autodetecting `ccache` support automatically, which may be desirable if, for instance, `ccache` is actually slower (it happens!) or you use `sccache`. Otherwise, this is an opt-in feature for users.

On a Linux machine, I saw a clean build time go down by 50% after the first build. On Windows, there is no significant difference, possibly because the file I/O time dominates the compile times.

### DIFF
--- a/litex/soc/software/common.mak
+++ b/litex/soc/software/common.mak
@@ -6,13 +6,14 @@ endif
 
 RM ?= rm -f
 PYTHON ?= python3
+CCACHE ?=
 
 ifeq ($(CLANG),1)
-CC_normal      := clang -target $(TRIPLE) -integrated-as
-CX_normal      := clang++ -target $(TRIPLE) -integrated-as
+CC_normal      := $(CCACHE) clang -target $(TRIPLE) -integrated-as
+CX_normal      := $(CCACHE) clang++ -target $(TRIPLE) -integrated-as
 else
-CC_normal      := $(TARGET_PREFIX)gcc -std=gnu99
-CX_normal      := $(TARGET_PREFIX)g++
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @cr1901, this looks good and can indeed be useful. This is merged.

geertu:
@cr1901 Do you really need this? The standard way to use ccache (as used by popular Linux distros) is to have a symlink from your compiler name to ccache early in your path:
```
$ type -all gcc
gcc is /usr/lib/ccache/gcc
gcc is /usr/bin/gcc
$ ls -l /usr/lib/ccache/gcc
lrwxrwxrwx 1 root root 16 okt 23  2020 /usr/lib/ccache/gcc -> ../../bin/ccache*
```
I.e. `export PATH=/usr/lib/ccache:$PATH` and all compilers installed by your distro will be used with ccache.

cr1901:
I'm on Windows, but even if I wasn't, I don't like adding symlinks to `/usr/lib`... I'd rather specify on the command line whether I want to use `(s)ccache` or not per project (because if I'm not building the application repeatedly for debugging/hacking, then its just polluting the cache w/ objects I've never access again). For autotools/meson, that's `CC=sccache cc`, for cmake that `CMAKE_C_COMPILER_LAUNCHER`.

Additionally, `sccache` is not a `ccache` replacement. It nominally should be, but on Windows at least, symlinking `sccache` to `ccache` will hang.

geertu:
Fair enough (poor Windows user ;-)

Note that (at least) Debian/Ubuntu manage the symlinks in `/usr/lib/ccache/` automatically.

PR #1098, opened 2021-11-10
liblitedram: Write latency calibration failure fix

kaolpr:
For targets that do not feature `SDRAM_PHY_WRITE_LEVELING_CAPABLE`  write latency procedure does not work properly. Use of `best_bitslip` was conditioned with value of `_sdram_write_leveling_bitslips[module]`. However, when `SDRAM_PHY_WRITE_LEVELING_CAPABLE` is not defined, this array is not filled with meaningful data. It makes the procedure ignore found `best_bitslip` value and in turn DDR calibration passes only when 0 is a valid bitslip. This PR ignores `_sdram_write_leveling_bitslips` value for targets not capable of write leveling.

### DIFF
--- a/litex/soc/software/liblitedram/sdram.c
+++ b/litex/soc/software/liblitedram/sdram.c
@@ -1045,10 +1045,14 @@ static void sdram_write_latency_calibration(void) {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @kaolpr! This looks good and has been merged.

PR #1100, opened 2021-11-11
sim/xgmii_ethernet: fix XGMII start of frame char handling and RX IFG insertion (with optional DIC)

lschuermann:
This PR fixes various things about the initial xgmii_ethernet simulator module. There are corresponding changes to be made on the LiteEth side, for which I will also release a PR. Until then this is a draft PR.

- sim/xgmii_ethernet: switch internal bus representation to 32 bits
  
  The xgmii_ethernet module used to represent internal XGMII interface signals using 64-bit signals. While this is not incorrect per se, it makes a standards-compliant implementation of the XGMII significantly harder in the long run. This is because XGMII fundamentally is defined as a 32-bit bus, and thus has constraints in relation to that bus width. For example, it is specified that packets may only ever start on the XGMII lane 0, that is the first octet of a 32-bit XGMII bus word. Hence, to properly handle a XGMII start of frame control character using a 64-bit bus representation, the first and fifth octet would need to be respected, along with shifting data depending on the start octet.

  Hence this change causes the internal XGMII interface to be represented as a 32-bit bus. Because it is common for FPGAs to implement the XGMII as a 64-bit bus given the fact that a DDR bus cannot be represented in an FPGA and 32 bits without DDR would require a clock frequency of 312.5MHz, it still permits 64-bit operation. This is implemented by always transmitting two 32-bit bus words back to back.

- sim/xgmii_ethernet: add RX IFG insertion

  The previous state of the xgmii_ethernet module did not insert an inter-frame gap at all. The IFG is mandated as part of the IEEE802.3 specification for Ethernet and as such the module was incorrect.

  This fixes the inter-frame gap insertion by keeping track of the IFG using a counter.

- sim/xgmii_ethernet: support RX IFG deficit idle count mechanism

  Because XGMII only allows start of frame characters to be placed on lane 0 (first octet in a 32-bit XGMII bus word), when a packet's length % 4 != 0, we can't transmit exactly 12 XGMII idle characters inter-frame gap (the XGMII end of frame character counts towards the inter-frame gap, while start of frame does not). Given we are required to transmit a minimum of 12 bytes IFG, it's allowed to send packet length % 4 bytes additional IFG bytes. However this would waste precious bandwidth transmitting these characters.

  Thus, 10Gbit/s Ethernet and above allow using the deficit idle count mechanism. It allows to delete some idle characters, as long as an average count of >= 12 bytes IFG is maintained. This is to be implemented as a two bit counter as specified in IEEE802.3-2018, section four, 46.3.1.4 Start control character alignment.

  This commit optionally implements the deficit idle count algorithm as described by Eric Lynskey of the UNH InterOperability Lab[1]. It may not have much or any impact in the XGMII simulation when using a TAP interface, however when connecting this simulation to another the DIC mechanism is a valuable implementation for both saturating proper 10Gbit/s line rate and validating the LiteEth PHY (XGMII to stream converter) implementation.

[1]: https://www.iol.unh.edu/sites/default/files/knowledgebase/10gec/10GbE_DIC.pdf

### DIFF
--- a/litex/build/sim/core/modules/xgmii_ethernet/xgmii_ethernet.c
+++ b/litex/build/sim/core/modules/xgmii_ethernet/xgmii_ethernet.c
@@ -51,6 +51,9 @@ static const char macadr[6] = {0xaa, 0xb6, 0x24, 0x69, 0x77, 0x21};
 #define XGMII_CTLCHAR_END   0xFD
 #define XGMII_CTLCHAR_IDLE  0x07
 
+#define XGMII_RX_IFG_INC(s) \
+    if (s->rx_ifg_count < 12) s->rx_ifg_count++;
+
 // Type definitions for the 32-bit XGMII bus contents, irrespective of the XGMII
 // bus width used. The data here is then latched out over a bus with the
 // xgmii_*_signal_t types (either 32-bit or 64-bit, which is two 32-bit words
@@ -156,6 +159,10 @@ typedef struct xgmii_state {
     // ---------- RX (TAP -> Sim) STATE ---------
     xgmii_rx_state_t rx_state;
 
+    // How many bytes of inter-frame gap we've produced on the XGMII
+    // bus (bounded counter to 12).
+    size_t rx_ifg_count;
+
     // Packet currently being received over the XGMII bus (TAP ->
     // Sim). Packets copied here are already removed from the TAP
     // incoming queue. Fields are valid if current_rx_len != 0. This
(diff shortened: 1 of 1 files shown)


lschuermann:
Rebased on latest master.

enjoy-digital:
Thanks @lschuermann, the fixes make perfect sense and are related to a module you developped/maintains, so happy to merge it if it's no longer a draft.

lschuermann:
> [...] happy to merge it if it's no longer a draft.

Awesome. I'm just now working on preparing the corresponding LiteEth PR. Maybe we can wait until that is up so that we have the latest and greatest of everything. :smile: 

Also, I've been using this simulator and the improved LiteEth PHY<->Stream adapter in various scenarios for weeks now, also with proper Intel, Aquantia and SolarFlare NICs, and haven't noticed any weird behavior, so it seems pretty reliable to me. XGMII is tricky to get right. :smile:

enjoy-digital:
Great, sure let's wait the corresponding LiteEth PR then. Also great to know you have already been able to test things with various other network equipments!

enjoy-digital:
Thanks a lot @lschuermann for this PR and https://github.com/enjoy-digital/liteeth/pull/95. Both are now merged :)

PR #1102, opened 2021-11-13
Eos s3 fix wb adr

trabucayre:
With EOS-S3 the Wishbone address bus is 8bits aligned, but all transaction are 32bits (or  WBs_BYTE_STB is used for 8 or 16 bits requests).
Anyway, in all cases the 2LSB are always 0.

In other hand, when data_width == 32, litex consider an address bus aligned (ie 2LSB are already discarded).
This patch fix alignment to fit litex requirement.


enjoy-digital:
Thanks @trabucayre! That's merged.

PR #1111, opened 2021-11-23
Add initial support for Anlogic devices

mmicko:


enjoy-digital:
Thanks @mmicko! That's merged.

PR #1113, opened 2021-11-26
microwatt: Use Xilinx multiplier

shenki:
Use the technology specific multiplier implementation when building for a
Xilinx platform.

This isn't understood by yosys so we can't use it when yosys is used for
synthesis.


shenki:
cc @paulusmack

enjoy-digital:
Thanks @shenki, this is merged.

PR #1114, opened 2021-11-28
fix FemtoRV reset signal

andykitchen:


enjoy-digital:
Thanks @andykitchen, good catch, this is merged.

PR #1115, opened 2021-11-28
Improve Gowin clock primitive support

sergachev:
I started these changes because Tang Nano 4K example from litex-boards did not compile: the Gowin compiler complained about VCO frequency:

```
ERROR (EX0311) : Invalid VCO frequency "(FCLKIN*(FBDIV_SEL+1)*ODIV_SEL)/(IDIV_SEL+1)" to instance "PLLVR", suitable range is from 600MHz to 1200MHz
```

Then reading the datasheets I found that the frequency range should depend on the device model (GW1N-1 vs GW1N-1S) and speed grade (C7/I6 vs C5/I4 etc) - so the PLL class should be made aware of the full device name which includes the speed grade.

'device' argument was already passed to the PLL class - but that is actually 'devicename' in the platform definitions, and there is 'device' in the platform too. So I propose to rename the existing 'device' in the PLL class to 'devicename' and add 'device' to match platforms 1:1.

Lastly GW1NSRPLL is too similar to GW1NPLL - so I propose to merge them to avoid code duplication.
I'll submit a matching MR for litex-boards.


enjoy-digital:
Thanks @sergachev for the improvements/simplifications. This is is merged.

PR #1116, opened 2021-11-30
litex_sim/xgmii: remove IFG SUFFICIENT debug message

lschuermann:
Removes a leftover debug message for validating correct XGMII IFG DIC behavior. Useful for development, but gets annoying quickly while running the simulation.


enjoy-digital:
Thanks @lschuermann, this is merged.

PR #1121, opened 2021-12-05
soc.intergration.dram: Use nranks in capacity calc

gregdavill:
This ensures that dual ranked memory appears with the correct capacity.

When testing butterstick configurations I noticed that the SoC always showed 1/2 the memory.
It looks like this calculation of sdram_size ignores dram ranks. Since the butterstick is dual rank this halves the memory reported/mapped to the SoC.


enjoy-digital:
Thanks @gregdavill, this is merged.

PR #1123, opened 2021-12-06
build/altera: fix rbf generation on windows

LazyDodo:
Testing for a file existing is slightly different
between bash and .bat files which caused the batch
file to error out.


enjoy-digital:
Thanks @LazyDodo, that's merged.

PR #1124, opened 2021-12-07
Defer adding the CFU instance until cpu.do_finalize().

tcal-x:


enjoy-digital:
Thanks @tcal-x, that's merged.

PR #1126, opened 2021-12-08
Add initial Gowin EMCU support

sergachev:
This adds support of Gowin EMCU primitive for hard ARM core on GW1NS(E)(R)-4C connected to 
- 8x16kb SRAMs (SDPB)
- boot flash (FLASH256K) via AHB (read-only by design) 
- LiteX CSR bus via another AHB (read-only as a temporary solution avoiding some strange behavior of the EMCU)
- UART pins (optional)

I added general AHB interface and a trivial AHB2Wishbone bridge too - not tested much yet, but at least EMCU can read correctly scratch CSR through it.

Nominally GowinEMCU class has correct compiler settings to build ARM software, but LiteX BIOS does not compile yet because of minor problems (exceptions, uint64_t division, missing startup asm code).

The way I test this is:
 - program FLASH256K using Gowin programmer in MCU mode with some example firmware from Gowin (cm3_uart, cm3_led etc) - slow but persistent
 - program LiteX gateware with openFPGALoader - quick, can change them often

### DIFF
--- a/litex/soc/cores/cpu/__init__.py
+++ b/litex/soc/cores/cpu/__init__.py
@@ -92,6 +92,8 @@ from litex.soc.cores.cpu.zynq7000 import Zynq7000
 # EOS-S3
 from litex.soc.cores.cpu.eos_s3 import EOS_S3
 
+from litex.soc.cores.cpu.gowin_emcu import GowinEMCU
+
 CPUS = {
     # None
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Great, thanks @sergachev, this is merged.

PR #1128, opened 2021-12-10
software.fatfs: Align fatfs sector buffers

gregdavill:
When working with the litesdcard I noticed that if the FatFs sector buffers were not aligned FatFs would fail, I suspect this is due to the DMA in litesdcard that's copying data into the buffer only performing 32bit writes.

In the FatFs header we can tell GCC to ensure that these buffers always fall on a word boundary.


enjoy-digital:
Thanks @gregdavill, the DMA accesses are indeed aligned on 32-bit.

PR #1129, opened 2021-12-10
clock, mmc updates for json2dts_linux

gsomlo:
replace `soc/bus-frequency` with a dedicated `clocks` section containing a `litex_sys_clk` node.

update `mmc` node with `reg-names` property, and add a reference to the above-mentioned `sys_clk` -- the upstream driver can't rely on there being a cpu with the exact required frequency, and needs a local reference to the clock it should use as the source to be divided in order to obtain `sd_clk`.

@geertu -- PTAL as well -- thanks!


mateusz-holenko (approved):
LGTM

enjoy-digital:
Thanks @gsomlo, @mateusz-holenko, sorry for the delay, this is merged.

PR #1131, opened 2021-12-13
Efinix titanium support

fjullien:


enjoy-digital:
Thanks @fjullien! This is merged.

PR #1132, opened 2021-12-13
cores/prbs: Add a 'wrap' option to PRBSRX

smunaut:
If enabled, then the error count wraps around when it
reaches the limit of the 32b counter instead of saturating
to a max value.

Software can then detect the wrap and act accordingly.


enjoy-digital:
Thanks @smunaut!

PR #1134, opened 2021-12-17
Efinix titanium support

fjullien:
Several fixes to get HyperRAM Tristate bus to work


enjoy-digital:

PR #1135, opened 2021-12-17
bios: add write size option to cmd mem_write

fjullien:


enjoy-digital:

PR #1136, opened 2021-12-20
tools/litex_json2renode: Add more restrictions to memory regions filter

robertszczepanski:
This PR fixes a workaround applied earlier. Skipping all linker regions does break some configurations where e.g. regions have type `cached+linker`.
With these changes script will look at region base address and size, then if any region is included in another it will be skipped but if they overlap on each other script will be aborted since it's not supported by Renode.

FYI @mateusz-holenko


enjoy-digital:
Thanks @robertszczepanski. @mateusz-holenko if this is good for you, we can merge.

mateusz-holenko (approved):
LGTM

enjoy-digital:
Thanks @robertszczepanski, @mateusz-holenko, this is merged.

PR #1137, opened 2021-12-20
zynq7000: add TCL preset support

trabucayre:
For zynq, **processing_system** IP may be configured using an **XCI** file, a preset shipped with vivado or an external TCL file.

This PR add support to use TCL file for boards without official preset (shipped with **vivado**).
This file may be produce using **block design**  -> **Customize block** -> **Presets** -> **Save Configuration**


enjoy-digital:
Thanks @trabucayre, this looks good and is merged.

PR #1140, opened 2021-12-22
argparse: deduplicate defaults in help messages

sergachev:


enjoy-digital:
Thanks @sergachev this is merged. Using `argparse.ArgumentDefaultsHelpFormatter` formatter also seems interesting: https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help, I'll probably have a closer look at it.

PR #1143, opened 2021-12-26
tools/litex_json2dts_linux.py fix double {{

Dolu1990:
A string with a {{ was missing the .format call, generating some broken dts


enjoy-digital:
Thanks @Dolu1990!

PR #1144, opened 2021-12-27
WS2812 off by one error, last LED never updates.

derekmulcahy:
I tested WS2812 with 8 LEDs and the last LED was never updated.

The original code set led_count=Signal(3) and then tested against (nleds-1) for 8 LEDs. This caused a transition to the RST state before processing the last LED. I increased the size of led_count to ensure it was large enough and tested against nleds directly.

### DIFF
--- a/litex/soc/cores/led.py
+++ b/litex/soc/cores/led.py
@@ -138,7 +138,7 @@ class WS2812(Module):
         # Internal Signals.
         led_data  = Signal(24)
         bit_count = Signal(8)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @derekmulcahy for reporting and the proposed PR. This is fixed a bit differently (with other simplifications) by https://github.com/enjoy-digital/litex/commit/8ac3fbc0391b133775cbb6f12db89a27d7b18bd4.

PR #1145, opened 2021-12-27
litex_json2dts_linux: add config_earlycon constant

cbrune:
This patch checks for a configuration constant named "config_earlycon"
and if present uses it for the "earlycon=" value in the dts
"bootargs".

If the constant is not present, default to the uart defined in the
"csr_bases".

This is useful for the sim.py case, where that configuration wants to
use "earlycon=sbi", while true hardware platforms want the uart from
"csr_bases".

See litex-hub/linux-on-litex-vexriscv#258


cbrune:
Closing this as it is no longer needed.

PR #1151, opened 2022-01-06
soc: raise an error if adding a SoCRegion with incoherent cache configuration

fjullien:
We could also print a warning and discard the cache configuration of this Region


enjoy-digital:
Thanks @fjullien!

PR #1155, opened 2022-01-12
litex/soc/software/bios/cmds/cmd_mem.c: add mem_cmp

tongchen126:
Add mem_cmp function. Compare the memory content starting from two addresses.


enjoy-digital:
Thanks @tongchen126, this is merged.

PR #1157, opened 2022-01-12
tools/litex_json2dts_linux.py: add mmc voltage regulator

gsomlo:
Add MMC voltage regulator, and reference it from the mmc node.
For now, all boards we know of provide a 3.3V fixed voltage to
their SDCard receptacles, either directly or via PMOD.


mateusz-holenko (approved):
LGTM

gsomlo:
merged, thanks @mateusz-holenko, @mithro

PR #1159, opened 2022-01-14
bios: fix mem_write address increment

fjullien:


enjoy-digital:

PR #1160, opened 2022-01-14
platform: add a method to copy any file to the build directory

fjullien:


enjoy-digital:
Thanks, this is can indeed be useful to include `.mem/.init` files but also in the future to create self-contained gateware directory with all the sources; but I'm not sure we should introduce another method, maybe we could just add a copy parameter to `add_source`. I'll have a closer look at this.

fjullien:
I did something more generic here: https://github.com/fjullien/litex/commit/6a78bee8c8414a026930985c3279533863c7ff18

enjoy-digital:
Thanks @fjullien, this is better. Can you avoid `sources_tmp` and just use `sources` and  copy is specified for a source, copy it to gateware directory and replace path in `sources` with the copied version? With this, happy to apply it.

fjullien:
I've done a new version here: https://github.com/fjullien/litex/commit/598d678f59bf9f138a68b3f10eab5e34ac90b46c
This is not trivial because add_source can be called at anytime in the build process (i.e. before the gateware_dir exist or in the build method of the toolchain).

enjoy-digital:
@fjullien: Sorry maybe I was not clear. I was just suggesting to collect the `sources` with along with copy parameter and do the copy/replace in the Builder (so skip the direct copy that is done directly in `add_source`) but I understand it's not trivial/maybe possible since `add_source` is also called when adding the LiteX generated verilog. You can send a PR with your solution if you want, we'll merge it and I'll have a closer look later to see if we can avoid the direct copy in `add_source`. Thanks.

Issue #1154, opened 2022-01-11
Separate address maps for CPU and SoC?


sergachev:
Some hard CPU cores (Zynq-7000, EOS S3) do address translation on peripheral buses - say for the CPU addresses of a bus start at 0x40020000 and CSR_BASE C definition has to have that value, but the address output to that bus in the logic still have this base offset removed and start from 0x0 (so decoding 2nd time done by LiteX by default breaks the access). I'm not sure what is the best way to deal with this - to manage a separate address space with all boundary checks etc or to simply redefine CSR_BASE in the C code or else. I solved it temporarily this way: https://github.com/sergachev/litex/commit/197a6353d597a4e06bab7d2af4e3e9dde31fabdd + https://github.com/sergachev/litex/commit/e93d543891599eb4c4b2b34ace4e158cfe19f437

trabucayre:
for Zynq7000 I'm a bit asthonish, all my tests shows using same addr at CPU and logic side works perfectly. But It's true for EOS and ZynqMP FPGAs.
I have doing a quite similar hack to you (before reading your commits) [here](https://github.com/trabucayre/litex/commit/a53786090295c13f598677a6c50245cd9a1488c6)
The idea is to use in this specifics cases a dict to provides CPU addr (key) and logic addr (value). Result is working for both FPGAs.
For all others devices using an int works without any modifications
I don't know which, from both, is the best answer to this problem

sergachev:
> for Zynq7000 I'm a bit asthonish, all my tests shows using same addr at CPU and logic side works perfectly.

It depends on how do you assign CSR address in the SoC memory map: if you set it to the value valid for the CPU (smth like [0x43c0_0000](https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/digilent_zedboard.py#L51)) CSR_BASE in C will be correct but LiteX will generate a decoder which will prevent the bus from functioning normally.
However if you set mem_map["csr"] to 0 (I think I've seen that somewhere in your code) you will get a correct bus connection without a decoder and if you then override CSR_BASE with the right value everything will work fine.

So I'm thinking that an easy temporary solution could be to just have an option to remove the bus decoder - given all current problematic situations require a zero base value on the logic side.

trabucayre:
It's true, with master branch for both *litex* and *litex-boards* repositories, the communication is not working for redpitaya and arty_z7,
[...]

PR #1162, opened 2022-01-15
QuickLogic EOS S3 ARM software support

sergachev:
This enables LiteX BIOS compile for and run on EOS S3 ARM core.
Interrupt handling, peripheral configuration etc rely on [QORC SDK](https://github.com/QuickLogic-Corp/qorc-sdk) and a small shim [C library](https://github.com/litex-hub/litex-boards/pull/322) that I propose to put into litex_boards for now because it's partially specific to Quickfeather board.

Note that the gateware will be slightly broken because CSR bus address decoding needs to be disabled (memory map here is valid only for the CPU, not the gateware) - see https://github.com/enjoy-digital/litex/issues/1154 for details; I'm ready to submit a fix as soon as that issue is addressed.


enjoy-digital:
Thanks @sergachev, this looks good and has been merged. I'll have a look at #1154 very soon, I just want to spend time thinking at this, sorry for th delay.

PR #1166, opened 2022-01-17
RFC: software/fatfs: allow sata, [spi]sdcard disk ops to co-exist

gsomlo:
At the moment, sata and [spi]sdcard fatfs disk access methods
are mutually exclusive, as their names collide with each other.

Implement a DISKOPS structure with media-specific disk access
methods, and a way for the boot code to specifiy which media
type's methods to use when loading files into RAM at boot time.

This (together with https://github.com/litex-hub/litex-boards/pull/325) allows me to build both sdcard and sata support for rocket on nexys_video like so:

```
litex-boards/litex_boards/targets/digilent_nexys_video.py --build \
    --cpu-type rocket --cpu-variant full4d --sys-clk-freq 50e6 \
    --with-ethernet \
    --with-sdcard \
    --with-sata --sata-gen 1
```

NOTE 1: I haven't actually ***run*** it yet, but so far this allows the build process to complete without crashing :)
NOTE 2: There might be more elegant ways to do this, happy to do revisions if needed (hence the "RFC" in the title)!

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -812,9 +812,11 @@ void sdcardboot(void)
 {
 #ifdef CSR_SPISDCARD_BASE
 	printf("Booting from SDCard in SPI-Mode...\n");
+	fatfs_set_ops_spisdcard();	/* use spisdcard disk access ops */
 #endif
 #ifdef CSR_SDCORE_BASE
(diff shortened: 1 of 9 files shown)


enjoy-digital:
Thanks @gsomlo, this looks good enough for me and I've been able to test it on hardware with SPI-SDCard. So I'm going to merge and we could eventually revisit it later.

PR #1169, opened 2022-01-19
Fix Zynq software support

sergachev:
This makes the memory map valid from the CPU point of view, enables software compilation for Zynq and adds a boot helper (this one not tested yet, but something like this should work).


enjoy-digital:
Thanks @sergachev, this is merged.

PR #1176, opened 2022-01-22
gowin: add hack for copackaged PSRAM

Icenowy:
Some Gowin FPGAs has copackaged "PSRAM" instead of "HyperRAM". They're
in fact also HyperRAM, but named differently and because of this needs
slightly changed hack.

Add hack for PSRAM too. In case of two PSRAM chips are used, the bus is
wider so the needed definition is already in place, so the bus width
hack is not needed.

Signed-off-by: Icenowy Zheng <icenowy@aosc.io>


enjoy-digital:

PR #1177, opened 2022-01-23
bitbang.py: initialize SCL / SDA lines to high on reset

yetifrisstlama:
* otherwise might block the I2C bus on reset


enjoy-digital:
Thanks @yetifrisstlama!

PR #1183, opened 2022-01-25
litex/soc/cores/clock/xilinx_common: fix yosys synth: replace FD by FDCE

trabucayre:
for Xilinx devices, when *yosys* is used for synthesis, as mentionned in this [issue](https://github.com/YosysHQ/yosys/issues/1583), FD primitive is not supported and this step fails with
```
ERROR: Module `\FD' referenced in module `\digilent_arty_z7' in cell `\FD_7' is not part of the design
```
This PR replace `FD` by `FDSE` with same behaviour.


enjoy-digital:
Thanks @trabucayre!

suarezvictor:
Excellent, it's exciting to see how fast and good this community moves. Thanks  @trabucayre!

PR #1188, opened 2022-01-27
Add JTAGbone support for Altera Max 10 and Cyclone 10 LP

jevinskie:


enjoy-digital:
Great! Thanks @jevinskie. Can you provide more info on the `CorrectedResetFSM` addition?

jevinskie:
Added to docstring:

This wrapper is needed for FSMs where an ongoing signal from the FSM's reset state is used.
With the existing FSM, on SoC reset the FSM will be in the reset state but its ongoing signal will not be asserted because the existing FSM does not set the reset values of the ongoing signals.

enjoy-digital:
(BTW: I confirm this is working fine on a Deca on my side and also tested things while refactoring).

PR #1189, opened 2022-01-27
Get the default IP prefix from LITEX_DEFAULT_IP_PREFIX env-var if set.

jevinskie:
My home network is on 192.168.1.x and I like to keep my misbehaving FPGA designs off of it so I use a different IP range. This is designed to make that easy.


enjoy-digital:
Thanks, I was thinking about doing some changes on this would avoid duplication of this. I'm not sure I want to do things as suggested, please give me a bit of time to think about it.

PR #1190, opened 2022-01-28
Copy files

fjullien:
Feel free to not merge it.
At least it won't get lost here.


enjoy-digital:
Thanks @fjullien, this is merged and simplified with https://github.com/enjoy-digital/litex/commit/2e574baca7fe014e9f669bf064e8a176dcf9c1e8. Can you verify it's still working correctly for your use case?

fjullien:
Thanks, it still works.
And of course use relative path makes sense.

PR #1191, opened 2022-01-30
Fix generation of documentation of `atomic_write` CSRStorage

david-sawatzke:
Can't decrement a range by 1, this (probably) never worked.

Also improves the generated text.


enjoy-digital:
Thanks @david-sawatzke! That's merged.

PR #1195, opened 2022-02-03
i2c: handle multiple i2c controllers in the BIOS

fjullien:
A new command is introduced: "i2c_dev".
With this command you can list i2c devices and choose on to be active.

### DIFF
--- a/litex/soc/cores/bitbang.py
+++ b/litex/soc/cores/bitbang.py
@@ -23,7 +23,7 @@ class I2CMaster(Module, AutoCSR):
     Software get back SDA value with the read CSRStatus (_r).
     """
     pads_layout = [("scl", 1), ("sda", 1)]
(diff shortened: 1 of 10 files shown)


enjoy-digital:
Thanks @fjullien, this looks good. We could probably apply the same principle to other peripherals in the future but we'll probably polish things with I2C first.

enjoy-digital:
@fjullien: I just simplfified a bit the HAS_I2C collection with https://github.com/enjoy-digital/litex/commit/26c44af3f77ced29ec95783001a597215293146c. Just for info, I'm in fact trying to avoid adding new code to `soc_core.py` since want to slowly deprecate it :)

enjoy-digital:
Thanks @fjullien, we can close this. We'll probably polish this over time but for now it's enough for the targeted use case.

PR #1196, opened 2022-02-03
efinix: fix EfinixTristateImpl

fjullien:
This fixes cases where:
- i is None,
- o is only 1 bit wide


enjoy-digital:
Thanks @fjullien.

PR #1197, opened 2022-02-03
video: align hsync and vsync generated by the vtg

fjullien:
I tested this only on one target. It should be tested with at least another one.


enjoy-digital:
Thanks @fjullien, I'll test it very soon (probably tomorrow).

enjoy-digital:
Thanks @fjullien, I just tested it succesfully on another hardware. That's merged.

PR #1203, opened 2022-02-09
zynqmp/core: allows users to select data_size and interface id

trabucayre:
this PR add:
- data_width for interface. ZynqMP AXI master may be configured in 32/64/128 bits
- the id interface is provided as parameter and no more deducted by axi list size. Some board may use by default LPD (**maxi_gp2**).
- fix **maxi_gp2** clock name: *LPD* instead of *FPD*

For *ZCU102* to be coherent with default configuration these modifications are required since **maxi_gp0** and **maxi_gp2** are enabled with a width of 128 and *alinx AXU2CGA* has, by default, only LPD (ie **maxi_gp2**) enabled.

Of course, for LiteX bios, these modifications may be less important: the PS configuration is done directly by the firmware. But when Linux is used in PS side, these configurations are done by FSBL (aka uboot SPL) and requires to export **psu_gpl_init.c**, updates uboot configuration, build and flash again (and it may error prone).

### DIFF
--- a/litex/soc/cores/cpu/zynqmp/core.py
+++ b/litex/soc/cores/cpu/zynqmp/core.py
@@ -38,9 +38,9 @@ class ZynqMP(CPU):
         super().__init__(*args, **kwargs)
         self.platform = platform
         self.reset          = Signal()
-        self.periph_buses   = []  # Peripheral buses (Connected to main SoC's bus).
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #1208, opened 2022-02-15
Cpu fix csr decoding

trabucayre:
As mentioned in issue #1154 some hard CPUs may have address translated before fabric or a second time when AXI is converted to wishbone:
- **eos s3** use absolute adress in CPU but relative address is exported to eFPGA side;
- **zynq7000** and **zynqMP** have, currently, two adress decoding (one in [SoCRegion](https://github.com/enjoy-digital/litex/blob/master/litex/soc/integration/soc.py#L87), a second in [AXILite2Wishbone](https://github.com/enjoy-digital/litex/blob/0d2183062d14d8babf444fe196c84a48d099586d/litex/soc/interconnect/axi.py#L542)

In all case, these decoding are not required and communication CPU<->FPGA fails due to mismatch addresses.

This PR (inspired by @sergachev [commit](https://github.com/sergachev/litex/commit/197a6353d597a4e06bab7d2af4e3e9dde31fabdd):
- adds `SoCRegion` CTOR parameter to allow/disallow decoding
- adds `SoC` attribute to store if a CSR address translation must be done (default True to keep default behaviour)
- adds for **eos**, **zynq7000** and **zynqmp** previous attribute to False to bypass translation at `SoCRegion` level.

### DIFF
--- a/litex/soc/cores/cpu/eos_s3/core.py
+++ b/litex/soc/cores/cpu/eos_s3/core.py
@@ -48,6 +48,7 @@ class EOS_S3(CPU):
         self.pbus           = wishbone.Interface(data_width=32, adr_width=15)
         self.periph_buses   = [self.pbus]
         self.memory_buses   = []
+        self.csr_decode     = False # Wishbone address is decoded before fabric
 
         # # #
 
--- a/litex/soc/cores/cpu/zynq7000/core.py
+++ b/litex/soc/cores/cpu/zynq7000/core.py
@@ -43,12 +43,13 @@ class Zynq7000(CPU):
         super().__init__(*args, **kwargs)
         self.platform       = platform
         self.reset          = Signal()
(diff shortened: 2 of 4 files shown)


enjoy-digital:
Thanks @trabucayre, this looks good. We'll maybe revisit this to extend access to other peripherals of the SoC in the future.

enjoy-digital:
@trabucayre: I don't think the implementation was following your description, so I fixed it with https://github.com/enjoy-digital/litex/commit/f46d1c190b9dc87ecfea43def69862ce99d99559. Can you have a look to verify it matches what you wanted to do? Thanks.

trabucayre:
True!
After too many tries, copies and cleanup a typo was introduced: yep csr_decode must be in CPU class and using a class member make sense.
Tested with zynqMP: works
Thanks for the fix

PR #1214, opened 2022-02-18
Improve bios memspeed performances

Dolu1990:
Done by 
- manualy unrolling 4 times
- Using a do while instead of a for loop

On digilent video, at 100 Mhz got : 

nax :
  Write speed: 124.6MiB/s
  Read speed: 136.4MiB/s
vex smp : 
  Write speed: 166.5MiB/s
  Read speed: 85.9MiB/s
  
Which make sense, vex has the advantage of being write through, boosting its performances at the cost of more transactions for litedram to manage.

The read speed difference is likely mostly due to Nax having a better branch prediction (btb)


enjoy-digital:
Thanks @Dolu1990!

PR #1224, opened 2022-02-24
platform: fix add_sources

fjullien:
If the Builder class is not used to build the projet, platform sources
is still a list of tupples with 4 elements.

For now, we don't handle file copy when Builder is not used but at least
it won't crash.


enjoy-digital:
Thanks @fjullien!

PR #1225, opened 2022-02-24
efinix:dbparser: add get_pad_name_from_pin and use it

fjullien:
get_gpio_instance_from_pin returned the pad name.
It now returns the gpio block name.

To get the pad name, there is now get_pad_name_from_pin.


enjoy-digital:
Thanks!

PR #1226, opened 2022-02-25
cores/xadc: Improve support for Zynq Ultrascale+

smunaut:
Although for 'standard' UltraScale+ just letting Vivado upgrade
SYSMONE1 to SYSMONE4 works, it doesn't for the Zynq because the
default SIM_DEVICE not matching up creates a fatal DRC when
creating bitstream.

So instead we add separate classes for UltraScale, UltraScale+ and
Zynq UltraScale+, each instanciating the right blocks, with the
right params. Also we only create the VCCPSINTLP//VCCPSINTFP/VCCPSAUX
for the Zynq UltraScale+.

### DIFF
--- a/litex/soc/cores/xadc.py
+++ b/litex/soc/cores/xadc.py
@@ -156,27 +156,13 @@ class XADC(S7SystemMonitor): pass
 
 # Xilinx Ultrascale System Monitor -----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @smunaut, this looks perfect and is merged.

PR #1227, opened 2022-02-25
Fix video (vtg and colorbars)

fjullien:


enjoy-digital:
Thanks @fjullien, I've been able to test it sucesfully on 2 boards and like the fact the code is simpler. This is merged.

PR #1235, opened 2022-03-08
Add sigma-delta DAC LiteX core.

tcal-x:
Tested on Arty and Icebreaker.

Sample usage from the BaseSoC for Arty, to output the analog signal on the first pin of PMOD B:
```
    def add_dac(self):
        self.platform.add_extension(arty.raw_pmod_io("pmodb"))
        outsig = self.platform.request("pmodb")[0]
        self.submodules.dac = DAC(outsig, 12)
```

The value to convert is written from software using:
```
#include <generated/csr.h>

    uint32_t v = .....;
    dac_value_write(v);
```

### DIFF
--- a/litex/soc/cores/dac.py
+++ b/litex/soc/cores/dac.py
@@ -1,0 +1,24 @@
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2022 Tim Callahan <tcal@google.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+from migen import *
+
+from litex.soc.interconnect.csr import *
+
+# DAC  ---------------------------------------------------------------------------------------
+
+class DAC(Module, AutoCSR):
+
+    def __init__(self, out, data_width):
+        self.out    = out
+
+        self._value = CSRStorage(data_width, reset_less=True, description="Digital value to convert to analog.")
+        value       = Signal(data_width)
+        accum       = Signal(data_width+1)
+
+        self.comb   += value.eq(self._value.storage)
+        self.sync   += accum.eq(accum[0:data_width] + value)
+        self.comb   += out.eq(accum[data_width])
--- a/litex/soc/cores/dac.py
+++ b/litex/soc/cores/dac.py
@@ -8,17 +8,55 @@ from migen import *
 
 from litex.soc.interconnect.csr import *
 
-# DAC  ---------------------------------------------------------------------------------------
+# DAC -----------------------------------------------------------------------------------------------
 
 class DAC(Module, AutoCSR):
+    """First-order Sigma-Delta DAC.
 
-    def __init__(self, out, data_width):
-        self.out    = out
+    Provides a small one-bit Sigma-Delta DAC output. When
+    ``with_constant_transition`` is set, the raw Sigma-Delta bitstream is
+    encoded as a 3-cycle symbol:
 
-        self._value = CSRStorage(data_width, reset_less=True, description="Digital value to convert to analog.")
-        value       = Signal(data_width)
-        accum       = Signal(data_width+1)
+    - ``1`` -> ``1, 1, 0``
+    - ``0`` -> ``1, 0, 0``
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @tcal-x, I'll have a look and do some tests soon.

Dolu1990:
Hi ^^

I did a similar thing in SpinalHDL, and basicaly, outputing the MSB of the register endup creating quite some distortion.
The reason seems to be related to the switching caracteristics of the IO.

I had much better results (SNR / FFT) using something in that kind  : 
https://github.com/SpinalHDL/SpinalHDL/blob/e48dc0eb13b3d8139645a81ae824014909f2fb59/lib/src/main/scala/spinal/lib/misc/analog/DacSigmaDelta.scala#L23

Where basicaly, depending the value of the MSB, you put a given 3 cycle sequance on the pin :
- MSB = '1' => 1,1,0 on the pin
- MSB = '0' => 1,0,0 on the pin

This ensure that the frequancy of the signal on the pin stay constant, at the cost of a frequancy divided by 3.

enjoy-digital:
Thanks @Dolu1990! I'm not yet familiar with Delta-Signal DACs but @tcal-x implementation seemed too simple to be true :)

@tcal-x; Do you want to have a look at implementing it the way @Dolu1990 just described it or do you want me to have a look? (I'm interested to explore this, so could re-implement it if you don't have time).

Dolu1990:
Some ref : 
https://opencores.org/projects/sigma_delta_dac_dual_loop

```
dsm2
allows to obtain higher clock frequency, and therefore higher oversampling ratio, but number of rising and falling slopes in time unit depends on signal value. Therefore you may experience nonlinear distortions if those two slopes are not symmetrical.
```

tcal-x:
Thanks for the comments @enjoy-digital and @Dolu1990 !

Some background -- I needed a DAC just to check that I was getting good data from an ADC, specifically the Digilent MIC3 PMOD (https://digilent.com/shop/pmod-mic3-mems-microphone-with-adjustable-gain/).   I wasn't going to use the DAC in the final design.   But it actually sounded better than I expected (the overall connection from MIC3 through the DAC shown here)!   So I thought it might be generally useful if someone wanted a quick and easy monitor on a digital audio data stream.

I got my original design from https://www.fpga4fun.com/PWM_DAC_3.html.   But I think the description there is wrong --- they DON'T have a PWM DAC ---  PWM would lump all of the 1's together for each audio sample, while the sigma delta DAC intermixes the 1's and 0's as much as possible.

Dolu1990, is the better SNR with your approach because the number of transitions stays constant?   Or because the output voltage doesn't approach the bounds?   Or both :)   ?      Were you able to measure the SNR with equipment, or did it sound better to your ear?

Dolu1990:
> Dolu1990, is the better SNR with your approach because the number of transitions stays constant? Or because the output voltage doesn't approach the bounds? Or both :)

I think it is more related to the number of transition staying constant.

> Were you able to measure the SNR with equipment, or did it sound better to your ear?

I do not have any traces of the mesurements i did, but from memory, the sigma delta without the 3 cycle sequance had visible artefact on both the scope and FFT.
A sine wave test i did was very distorted, especialy in the middle range, and on the FFT the amount of armonics was more than 20 db worst.

But that was some fast test with a cheap scope ^^

tcal-x:
@enjoy-digital it's fine to close this without merging.   As we've seen there's many different way of building the DAC.   I mostly wanted to make a PR so that I could find the core easily down the road if I needed it.

enjoy-digital:
@tcal-x: I'm interested by the core but just want to play a bit with it, do some tests and itnegrate @Dolu1990's improvements.

tcal-x:
:+1:

enjoy-digital:
Thanks @tcal-x, this has been merged with follow up improvements (taking @Dolu1990 advices into account, improving codestyle, adding a test bench, etc...).

PR #1236, opened 2022-03-10
build/trelis: Compress bitstream by default

gregdavill:
Can we turn on bitstream compression as a default option?

Are there any reasons people can think of as to why this should default to disabled?


enjoy-digital:
@gregdavill: Thanks, it should be safe to enable it by default yes.

PR #1246, opened 2022-03-20
bios/memtest: fix cast warning

gsomlo:
I get a warning when building the bios (for rocket, 64-bit) -- so I believe that cast is missing a `*` to make it a pointer. @dolu, LMK what you think -- thanks!


Dolu1990:
@gsomlo Hooo, i think you are right, thanks ^^
Good to merge ^^

PR #1248, opened 2022-03-23
litex_setup: change to installing with pip

AndrewD:
fixes: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.


enjoy-digital:
Thanks @AndrewD!

PR #1250, opened 2022-03-24
Fix missing "last" pin connection when using connect_to_pads

pftbest:
When using  connect_to_pads on full AXI interface the "last" pin was not connected because of this typo.


enjoy-digital:
Thanks @pftbest, good catch!

PR #1251, opened 2022-03-24
export: fix mask for field size 31

AndrewD:
software/include/generated/csr.h:110:2: warning: integer overflow in expression '-2147483648 - 1' of type 'int' results in '2147483647' [-Woverflow]
  uint32_t mask = ((1 << 31)-1);
  ^~~~~~~~


enjoy-digital:
Thanks @AndrewD!

PR #1254, opened 2022-03-26
[enh] added efinix atmel programmer

chmousset:
The original programmer only supports a single hexlified bitstream format, this one also supports binaries programming


enjoy-digital (approved):
Thanks @chmousset, this looks good. I'll just do some minor changes to avoid `usb` package dependency.

PR #1255, opened 2022-03-28
Vivado IP packaging

sebo83910:
Hello,

This is my proposal for supporting Vivado IP packaging embedded in litex.
Due to the complexity of the process, I've added a separate file named vivadoippackaging.py.
There is also an impact on platform.py as I propose the method "package".
The packaging process has been tested on a single file using the following features:
- support for AXI Stream and AXI Lite interfaces
- Support custom interfaces (I used the wishbone to test)
- support interrupts
- Add generic parameters to the verilog code generated so they can be accessible in read only in the GUI
- Add GUI basic customization (for now).
- support XDC, Verilog and XCI files.
- support versionning.

The unittest is made available as well.

I hope you'll consider this add-on to be integrated in LiteX. It could be cool.

Take care.

Sébastien.


enjoy-digital:
Thanks @sebo83910! I'll do a closer review and see how we could integrate this. This could also be a tool available in `litex/tools` that would install iteself with LiteX as a `litex_vivado_packager` command. Please give me a bit of time to compare it with what I did with LitePCIe/LiteDRAM packaging and understand the additional features and I'll get back to you.

sebo83910:
That could also do the work indeed. However, keep in mind that I packaging is also possible for Intel FPGAs as well. 
I let you decide.

S/

ashuthosh-mr:
Is litex_vivado_packager feature available in the latest release? 

Thanks

PR #1258, opened 2022-03-30
Make memtest work with 4x DQ / DQS ratio

RRozak:
It depends on https://github.com/enjoy-digital/litedram/pull/300


enjoy-digital:
Thanks @RRozak, this is merged.

Issue #682, opened 2020-10-21
Enable LTO/remove unused symbols when compiling BIOS/firmware


DurandA:
The current GCC build (I did not try Clang) for BIOS/firmware does not strip away unused symbols, resulting in large binaries (especially when adding dependencies).

I tried adding various flags in [_common.mak_](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/common.mak) according to [answers on Stack Overflow](https://stackoverflow.com/questions/6687630/how-to-remove-unused-c-c-symbols-with-gcc-and-ld) but without much success. It either resulted in no effect or the build failed. With `-flto` I was not able to fix a `plugin needed to handle lto object` error.

mithro:
@DurandA -- There was some discussion about enabling link time option.

mithro:
See https://github.com/enjoy-digital/litex/pull/401

mithro:
https://github.com/enjoy-digital/litex/pull/253

DurandA:
@mithro Thanks for pointing it and sorry for the redundancy.

LTO support was originally merged but https://github.com/enjoy-digital/litex/commit/979f98ea3120dbba208c2b0b044ce37294653814 reverted it. At least we have an example on how to enable it.

mithro:
@DurandA - We most certainly want to get LTO enabled again. Maybe you could work with @mateusz-holenko to figure out what is needed to done to do that?

DurandA:
I am now into the realms of a GCC bug (https://github.com/riscv/riscv-gcc/issues/201) so that might not happen anytime soon. Note that precompiled GCC from SiFive (both version `gcc-8.3.0-2019.08.0-x86_64` and `gcc-8.3.0-2020.04.1`) triggered another _internal compiler error_ from the lto-wrapper.

mithro:
I'm going to reopen this issue so that @mateusz-holenko can give us an updated on the status around LTO and how to move forward here.

mateusz-holenko:
I guess that linked tasks/PRs provide all the status - we didn't work on the LTO issue lately. 
As described in #253, using it with optimisations caused problems back in the past. We were able to find a combination of switches that worked*, but it turned out not to be a generic solution (* as it broke some other platforms). Maybe a newer toolchain might help here? 

[...]

PR #1259, opened 2022-04-01
Remove -flto flag from demo.bin

developandplay:
Fixes the failing lift-off on the Rocket chip.

### DIFF
--- a/litex/soc/software/demo/Makefile
+++ b/litex/soc/software/demo/Makefile
@@ -8,6 +8,9 @@ ifdef WITH_CXX
 	OBJECTS += hellocpp.o
 endif
 
+CXXFLAGS:=$(filter-out -flto,$(CXXFLAGS))
+CFLAGS:=$(filter-out -flto,$(CFLAGS))
+
 all: demo.bin
 
 # pull in dependency info for *existing* .o files


mithro:
@developandplay - Can you upload the `demo.bin` (or even better the original `demo.elf` file) compiled with `-flto` and without `-flto`?

We should really figure out why `-flto` is breaking the demo on Rocket.

@kgugala - Do you have someone who could dive into this and figure out what is going on?

kgugala:
@mithro I'll find someone to check this.
@developandplay it'll be helpful if you can share the binaries

developandplay:
Sure here are the binaries and a default dump. I had a quick look at it but I couldn't make anything of the binary differences as there seem to be quite a lot of them.

[demo-binaries.zip](https://github.com/enjoy-digital/litex/files/8400004/demo-binaries.zip)

developandplay:
I just tried running the same thing on BlackParrot in simulation. (With `ram_boot_offset  = 0x80000000`)
```
lxsim --cpu-type blackparrot --cpu-variant standard --with-sdram   # for csr.csv generation
litex_bare_metal_demo --build-path=build/sim
lxsim --cpu-type blackparrot --cpu-variant standard --with-sdram --sdram-init demo.bin
```

The outcome is also that it works without `-flto` but fails with `-flto`.
So it seems like it's more a binary issue than a Rocket issue.

mithro:
@developandplay - It is most certainly an issue with the binary in some way. 

`-flto` enables "link time optimization" (https://gcc.gnu.org/wiki/LinkTimeOptimization) which is primarily suppose to remove unused code sections resulting is much smaller firmware sizes (as only used functionality is actually included).

enjoy-digital:
LTO has been disabled in the BIOS since still need to cause random issue and #682 has been re-opened (see https://github.com/enjoy-digital/litex/issues/682#issuecomment-1145927034). I'm closing this since no longer useful but it would still be interesting to properly investigate LTO issues to be able to enable it.

PR #1262, opened 2022-04-01
Fix demo.bin compilation for BlackParrot

developandplay:
Same issue as:
https://github.com/enjoy-digital/litex/pull/1245


enjoy-digital:

PR #1265, opened 2022-04-03
Improve WS2812 timings and add different hardware revision support

wnagele:
Started this when I had some newer model WS2812 LEDs for which the current timing didn't work.

So I added a test bench that validates everything according to the WS2812 datasheets for the old and new hardware revision.

As I was digging down into that rabbit hole I found a bunch of issues with the current FSM implementation causing timings to be off due to extra cycles spent in the low states. This doesn't matter much at higher clock speeds - but at lower clock speeds it can break the timing constraints.

Happy for feedback - this is my first PR towards LiteX so I did my best to try to make everything fit into the structure.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -124,6 +124,7 @@ Copyright (c) 2019      vytautasb <v.buitvydas@limemicro.com>
 Copyright (c) 2013      Werner Almesberger <werner@almesberger.net>
 Copyright (c) 2015-2016 whitequark <whitequark@whitequark.org>
 Copyright (c) 2015-2021 William D. Jones <thor0505@comcast.net>
+Copyright (c) 2022      Wolfgang Nagele <mail@wnagele.com>
 Copyright (c) 2013-2014 Yann Sionneau <yann.sionneau@gmail.com>
 Copyright (c) 2015 Yves Delley <hack@delley.net>
 Copyright (c) 2020      Xiretza <xiretza@xiretza.xyz>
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks a lot @wnagele, the initial implementation aims to be easy to follow and to be run in regular range of `sys_clk_freq ` (50-100MHz). Your testbench is very useful and indeed allow catching issues even at these frequencies. The way the code is designed however expect this _high_ `sys_clk_freq`. When lowering to 10MHz, we are really close to the error margin and don't have any spare cycles to loose in an FSM (as you saw while modyfying the design) and would require a redesign. Most of the modifications you did are working, but makes the code no longer easy to understand, which was the original intend.

So I'm going to do a compromise: Merge your PR but rework the changes a bit an avoid some of the modifications that make the code difficult to understand.

enjoy-digital:
@wnagele: The changes have been done in https://github.com/enjoy-digital/litex/commit/d39c3ed626842a4dce41ea3477b26e5ee2ce6af3. This is not as optimized as the different optimization you did, but should keep the code easy to follow and still allow `sys_clk_freq` down to 25MHz, which is less than the `sys_clk_freq` used on regular LiteX SoCs. Now that we have your unit-test in place, it will be more easy to further improve this. I'll try to have another look when I'll have more time. Happy to discuss optimization you can see that don't add too much complexity to the code.

Thanks a lot for your PR!

wnagele:
@enjoy-digital that all makes sense to me. I also didn't like the code becoming unwieldy because of those changes. Thanks for getting this merged quickly.

PR #1276, opened 2022-04-15
soc/cores/uart:Stream2Wishbone: fix address count

trabucayre:
With default behaviour, when address_width != data_width, address may be corrupted. This is due to the same bytes counter and done signal.
This PR add a dedicated counter + done for address with size accordingly.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -312,15 +312,17 @@ class Stream2Wishbone(Module):
 
         # # #
 
-        cmd         = Signal(8,                        reset_less=True)
-        incr        = Signal()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre. The change does not seem complete ("SEND-DATA" state is not adapted). For now I added an assert with https://github.com/enjoy-digital/litex/commit/8b2fc201cf0d6abd137963d481d22df9ba673524, can you explain your use case to see if we should try to support it or just keep this assert?

trabucayre:
True. There is a wrong diff before push :-/ sorry

trabucayre:
fixed in #1277

PR #1277, opened 2022-04-15
Uart fix stream2 wishbone v2

trabucayre:
With default behaviour, when address_width != data_width, address may be corrupted. This is due to the same bytes counter and done signal.
This PR add a dedicated counter + done for address with size accordingly.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -312,8 +312,6 @@ class Stream2Wishbone(Module):
 
         # # #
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre, as asked in the previous PR, can you explain your use case to see if we should try to support it or just keep this assert?

enjoy-digital:
Thanks for the info and update of the code. This is merged.

Issue #1270, opened 2022-04-11
Add the ability to insert boot methods to the BIOS's bootlist, from an external library


DaveBerkeley:
Background : I want to add a SPI Flash boot using my own bit-banged interface to a SPI Flash, not using the current Litex SPI Flash driver.

The only way seems to be to modify the source of litex/soc/software/bios/boot.c or main.c to manually add a call to an external library function. This can be added by creating a soft link to the external library at, eg. litex/soc/software/libother

In other projects I've used gcc's constructor attributes to create a function that gets run during cstart, after the memory is initialised and before main() is called.

eg:

```
__attribute__((constructor))
static void lib_init(void)
{
        // TODO : run code here
        // Gets called before main()
}
```

See https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html

I've used this to insert a static struct in a linked list. The function lib_init() gets called during cstart. It links a boot method into a list. The boot code in main.c boot_sequence() can then simply iterate over this list, calling each boot method in turn. This mechanism allows new methods to be dynamically inserted into the list without having to modify the BIOS source.

However, __attribute__((constructor)) requires support in the linker and cstart code. The linker definition file, linker.ld, needs .ctors and .dtors sections adding to it. This is also required for any static initialisation of C++ classes, so eases the way to providing full C++ support.

See https://blog.thea.codes/the-most-thoroughly-commented-linker-script/ for an example linker script.
 
The init code needs to add a function to a linked list. I would like to add a simple structure and some support code, which allows adding and iterating of the list.

Existing boot methods could be adapted to use the same technique, so the main boot sequence function could be reduced to eg

```
static void boot_sequence(void)
{
#ifdef CSR_UART_BASE
	if (serialboot() == 0)
		return;
#endif
    fnlist_run(bootlist);
}
```

The order of the boot methods could be determined by using the 'priority' parameter of the constructor attribute.

The same technique could be used by the BIOS to call init methods for peripherals etc.

I have a working prototype of this proposed system. I'd welcome comments on the idea.

A secondary problem is adding external libraries to the BIOS build. This is partially solved by calling eg : 

    for lib in [ "libother" ]:
        builder.add_software_package(lib)
        builder.add_software_library(lib)

But there is no way to tell the BIOS builder where the code is. It currently has to be located in the litex code tree.

DaveBerkeley:
[...]

PR #1278, opened 2022-04-17
Added init_array support for vexriscv to allow static C++ ctors and C…

DaveBerkeley:
init_array is a technique used by gcc to construct a list of pointers to functions that should be run just prior to calling main(). This is used in C++ to support static initialisation of objects and is supported in C via __ attribute __((constructor)).

See discussion in : https://github.com/enjoy-digital/litex/issues/1270

To provide support for other processors, "call __libc_init_array" needs to be added to their crt0.S files. I don't have the facilities to test these other processors at present. I've also only tested with gcc.

The technique can be used to run functions, in a defined order, prior to calling main. These functions can be in other libraries. This allows initialisation code to be added to the BIOS without the need to edit the bios C source for a new library.

The technique is similar to that currently used to add BIOS commands (see bios_cmd_start bios_cmd_end in the linker definition file)

It does not add a requirement for, or dependency on C++.


DaveBerkeley:
Zephyr seems to use a similar method of adding initialisation functions. 

https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/init.h#L85-L90

enjoy-digital:
Thanks @DaveBerkeley and sorry for the delay, this has been integrated with https://github.com/enjoy-digital/litex/commit/97ed83b6d309ec23d4883595b9e9362c90ce6f58: Updated and applied to all CPUs.

PR #1282, opened 2022-04-25
software/bios: Fix build error for targets with sdcard but without main_ram_base

cklarhorst:
sdcardboot_from_bin assumes that the MAIN_RAM_BASE is set which is not always
the case. This commit deactivates the sdcardboot_from_bin support for targets
without main-ram. In those cases, the boot_from_json method can still be used.


enjoy-digital:
Thanks @cklarhorst, this makes sense and is merged.

PR #1283, opened 2022-04-26
Etherbone testing

michael-betz:
Changes to stream_sim.py and stream.py to make liteeth testbenches and etherbone work with a 64 bit datapath.
Corresponding [liteeth pull request](https://github.com/enjoy-digital/liteeth/pull/106).

  * PacketStreamer / PacketLogger: make them take care of `last_be` and the conversion from bytes to words and vice versa. The interface towards the user is always in bytes
  * Also, make PacketStreamer / PacketLogger recognize stalls, which some PHYs don't like
  * StrideConverter: add last_be support when down-converting

### DIFF
--- a/litex/compat/__init__.py
+++ b/litex/compat/__init__.py
@@ -23,7 +23,7 @@ def compat_notice(name, date, info=""):
         date       = colorer(date),
         update     = colorer("update", color="red"),
         info       = info,
-    ), end="")
+    ))
     # Annoy user to force update :)
     for i in range(10):
         time.sleep(0.2)
--- a/litex/compat/stream_sim.py
+++ b/litex/compat/stream_sim.py
@@ -114,16 +114,20 @@ class PacketStreamer(Module):
         # # #
 
         self.packets = []
+        self.packet_bytes = []
+
+        self.n_bytes = 0
         self.packet = packet_cls()
         self.packet.done = True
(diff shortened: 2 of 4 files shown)


enjoy-digital:
Thanks for this work.

I reworked the still-relevant LiteX-side part in a focused commit:

  enjoy-digital/litex@a2a62dcc662465015dcd55d0de701d41de40a660

The useful remaining issue here was `StrideConverter` handling of `last_be` when down-converting a wide packet stream to a narrower one. A final wide word can end in an earlier narrow slice; previously `StrideConverter` only asserted `last` on the final ratio slice, so invalid trailing slices could be exposed.

LiteX now handles this generically when both endpoints expose compatible `last_be` fields: the slice carrying `last_be` is emitted as the packet `last` beat, and the remaining internal slices are drained. The old behavior is kept when `last_be` is absent or zero.

I also added regression coverage for:
  - partial final words;
  - full final words;
  - the legacy `last_be=0` path.

The other parts of this PR are no longer good merge candidates for current LiteX: `litex/compat/*` has been removed, the stream test helper work now belongs on the LiteEth side, and the Etherbone/debug changes should be reviewed separately if still useful.

So I think we can close this PR now, with the focused `StrideConverter` fix integrated separately.

PR #1284, opened 2022-04-26
fix overflow in BIST

smunaut:
With decent burst length, fast RAM and fast clock rate,
the intermediate result can easily overflow 32 bits.


enjoy-digital:
Thanks @smunaut!

PR #1287, opened 2022-04-28
tools: litex_json2dts_zephyr: Comply with changes in Zephyr

michalsieron:
**Should be merged after https://github.com/zephyrproject-rtos/zephyr/pull/45198 is merged first**

Changes in Zephyr require updated script to generate correct overlay.

This commit splits CSR regions that were being added to overlay as single register into separate registers with names.

It also prints their size in bytes instead of used subregisters.

I also added LITEX_CSR_DATA_WIDTH to generated config parameters.


litex/tools/litex_json2dts_zephyr.py
```diff
@@ -137,10 +166,28 @@ def generate_dts_config(csr):
...
+        csr['constants']['config_csr_data_width'],
+    )
+
     return dts, cnf
 
 
 # helpers
+def get_registers_of(name, csr):
+    registers = csr['csr_registers']
+    data_width = csr['constants']['config_csr_data_width']
```
mateusz-holenko, line 179: Is this used anywhere?

litex/tools/litex_json2dts_zephyr.py
```diff
@@ -42,11 +43,31 @@ def ram_handler(name, parm, csr):
...
+        reg['name'] = 'tx_' + reg['name']
+
+    eth_buffers = {
+        'name': 'buffers',
+        'addr': csr['memories'][name]['base'],
+        'size': csr['memories'][name]['size'],
+        'type': csr['memories'][name]['type'],
+    }
+    registers = rx_registers + tx_registers + [eth_buffers]
+    if len(registers) == 0:
```
mateusz-holenko, line 61: we create `eth_buffers` and always include them in `registers`, so they can't be empty at this point, right?

litex/tools/litex_json2dts_zephyr.py
```diff
@@ -137,10 +166,28 @@ def generate_dts_config(csr):
...
 
 # helpers
+def get_registers_of(name, csr):
+    registers = csr['csr_registers']
+    data_width = csr['constants']['config_csr_data_width']
+
+    return [
+        {
+            **params,
+            'size': params['size'] * 4,
```
mateusz-holenko, line 184: a comment why we are multiplying `size` would be great

litex/tools/litex_json2dts_zephyr.py
```diff
@@ -42,11 +43,31 @@ def ram_handler(name, parm, csr):
...
+        'name': 'buffers',
+        'addr': csr['memories'][name]['base'],
+        'size': csr['memories'][name]['size'],
+        'type': csr['memories'][name]['type'],
+    }
+    registers = rx_registers + tx_registers + [eth_buffers]
+    if len(registers) == 0:
+        raise KeyError
+
+    sep = ' ' * 8
```
mateusz-holenko, line 64: this seems to be duplicated - perhaps we could extract a common helper?

litex/tools/litex_json2dts_zephyr.py
michalsieron: Yes, in updated handlers above. If you want, I can move it up, so it is interpreted, before functions that use it.

michalsieron: Moved function just below imports and removed unused `data_width` variable

litex/tools/litex_json2dts_zephyr.py
michalsieron: That is correct, removed this check, as `KeyError` would be raised during creation of `eth_buffers` anyway if there is no ethmac

litex/tools/litex_json2dts_zephyr.py
michalsieron: Done

litex/tools/litex_json2dts_zephyr.py
michalsieron: Added functions to indent text and refactored formatting functions to use them

mateusz-holenko (approved):

lachlansmith:
Could this please be merged

michalsieron:
I added i2s and mmcm handlers as well

mithro:
@mateusz-holenko - Can you take another look at this?

mateusz-holenko (approved):
I'd suggest splitting this into more commits, but otherwise LGTM.

michalsieron:
Ok, I extracted last addition of i2s and mmcm handlers into separate commit

PR #1290, opened 2022-05-03
cores/video: VideoHDMIPHY: enable driving both + and - diff outs

swetland:
Add a new parameter (drive_both), defaulting False, that will
drive data#_n as well as data#_p (and with the inverse of data#_p)

VideoHDMI10to1Serializer is similarly extended.

Existing users of these should be unaffected (tested that with
the Radiona ULX3S, which was not impacted).

### DIFF
--- a/litex/soc/cores/video.py
+++ b/litex/soc/cores/video.py
@@ -740,7 +740,7 @@ class VideoDVIPHY(VideoGenericPHY): pass
 # HDMI (Generic).
 
 class VideoHDMI10to1Serializer(Module):
-    def __init__(self, data_i, data_o, clock_domain):
+    def __init__(self, data_i, data_o, clock_domain, data_o_n=Signal(), drive_both=False):
         # Clock Domain Crossing.
         self.submodules.cdc = stream.ClockDomainCrossing([("data", 10)], cd_from=clock_domain, cd_to=clock_domain + "5x")
         self.comb += self.cdc.sink.valid.eq(1)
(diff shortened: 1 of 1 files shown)


swetland:
It was suggested (thanks, tnt) that a simpler fix would be to adjust the IOStandard of the pins, but it turns out that in the case of the IceSugar Pro, the pins were not chosen well:
ERROR: cannot place differential IO at location PIOB

swetland:
I'm very new to Migen so hopefully there's a cleaner solution.  The underlying issue is the IceSugarPro board fails to locate the HDMI signals on suitable adjacent IOs that can be used for differential signaling, so you have to actually route the + and - sides manually.

enjoy-digital:
This has been reworked a bit with https://github.com/enjoy-digital/litex/commit/7e1d2bdf9b7e50cf723fce02e495754288b2b10b. I'd like to avoid propagating this feature to VideoHDMI10to1Serializer, so this is self-contained to VideoHDMIPHY which is less optilmal in term of resources (2 x serializers) but we are fixing an hardware issue here so I also want to avoid propagating too much mess in the code for this. If resource usage is really an issue for you on this board, we could rediscuss it.

swetland:
I don't think it'll be a problem.  The more compact, contained solution seems like an improvement to me.  Thanks!

swetland:
I need to shuffle around some local work in progress to sync up and test, but should be able to do so later today.

swetland:
Confirmed: Everything continues to work with these changes on IceSugarPro.  Thanks!

PR #1292, opened 2022-05-08
cpu/vexriscv: Fix compilation of crt0.S with new binutils.

zeldin:
The csrw opcode is no longer part of the "I" instruction set but has
been moved to a separate extension.  Enable that extension in crt0.S.


Dolu1990:
Ok thanks ^^

PR #1294, opened 2022-05-09
add cva6 cpu

suppamax:
The CVA6 code is https://github.com/suppamax/pythondata-cpu-cva6.

I added a SystemVerilog wrapper around CVA6, to glue the CPU, PLIC, CLINT and debug module together.

### DIFF
--- a/litex/soc/cores/cpu/cva6/__init__.py
+++ b/litex/soc/cores/cpu/cva6/__init__.py
@@ -1,0 +1,1 @@
+from litex.soc.cores.cpu.cva6.core import CVA6
--- a/litex/soc/cores/cpu/cva6/boot-helper.S
+++ b/litex/soc/cores/cpu/cva6/boot-helper.S
@@ -1,0 +1,18 @@
+.section    .text, "ax", @progbits
+.global boot_helper
+.global smp_ap_args
+.global smp_ap_target
+.global smp_ap_ready
+
+boot_helper:
+  // boot core saves args and jump target for ap cores:
+  sd a0, smp_ap_args, t1
+  sd a1, smp_ap_args+8, t1
+  sd a2, smp_ap_args+16, t1
+  sd a3, smp_ap_target, t1
+  fence w, w
+  // notify application cores to proceed with boot:
+  li t0, 1
+  sd t0, smp_ap_ready, t1
+  // boot core now also ready to boot:
+  jr a3
--- a/litex/soc/cores/cpu/cva6/core.py
+++ b/litex/soc/cores/cpu/cva6/core.py
@@ -71,6 +71,7 @@ class CVA6(CPU):
     def gcc_flags(self):
         flags = GCC_FLAGS[self.variant]
         flags += "-D__cva6__ "
+        flags += f" -DUART_POLLING"
         return flags
 
     # Memory Mapping.
--- a/litex/soc/cores/cpu/cva6/core.py
+++ b/litex/soc/cores/cpu/cva6/core.py
@@ -1,0 +1,195 @@
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2021 Hensoldt Cyber GmbH <www.hensoldt-cyber.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+import os
+import re
+
+from migen import *
+from migen.fhdl.specials import Tristate
+
+from litex import get_data_mod
+from litex.soc.interconnect import axi
+from litex.soc.interconnect import wishbone, stream
+from litex.soc.interconnect.csr import *
+from litex.soc.cores.cpu import CPU, CPU_GCC_TRIPLE_RISCV64
+
+class Open(Signal): pass
+
+# Variants -----------------------------------------------------------------------------------------
+
+CPU_VARIANTS = ["standard", "full"]
+
+# GCC Flags ----------------------------------------------------------------------------------------
+
+GCC_FLAGS = {
+    #                       /-------- Base ISA
+    #                       |/------- Hardware Multiply + Divide
+    #                       ||/----- Atomics
+    #                       |||/---- Compressed ISA
+    #                       ||||/--- Single-Precision Floating-Point
+    #                       |||||/-- Double-Precision Floating-Point
+    #                       imacfd
+    "standard": "-march=rv64imac    -mabi=lp64 ",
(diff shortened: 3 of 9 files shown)


enjoy-digital:
Hi @suppamax,

thanks for the initial CVA6 support! I started looking at this, but had to do some changes to the flist for things to compile with Vivado:
```
diff --git a/pythondata_cpu_cva6/system_verilog/Flist.cv64a6_imafdc_sv39 b/pythondata_cpu_cva6/system_verilog/Flist.cv64a6_imafdc_sv39
index 8e7d140..a8baecb 100644
--- a/pythondata_cpu_cva6/system_verilog/Flist.cv64a6_imafdc_sv39
+++ b/pythondata_cpu_cva6/system_verilog/Flist.cv64a6_imafdc_sv39
@@ -68,6 +68,11 @@ ${CVA6_REPO_DIR}/common/submodules/common_cells/src/shift_reg.sv
 ${CVA6_REPO_DIR}/common/submodules/common_cells/src/unread.sv
 ${CVA6_REPO_DIR}/common/submodules/common_cells/src/popcount.sv
 ${CVA6_REPO_DIR}/common/submodules/common_cells/src/exp_backoff.sv
+${CVA6_REPO_DIR}/common/submodules/common_cells/src/addr_decode.sv
+${CVA6_REPO_DIR}/common/submodules/common_cells/src/spill_register.sv
+${CVA6_REPO_DIR}/common/submodules/common_cells/src/spill_register_flushable.sv
+${CVA6_REPO_DIR}/common/submodules/common_cells/src/stream_register.sv
+${CVA6_REPO_DIR}/common/submodules/common_cells/src/cdc_2phase.sv
 
 // Common Cells for example coprocessor
 ${CVA6_REPO_DIR}/common/submodules/common_cells/src/counter.sv
```
and we should also probably integrate the custom wrapper/flist directly in LiteX for consistency with other CPUs, but we could see this later.

With the changes I did, I haven't been able to run it on hardware (tested compilation on an SQRL Acorn). Could you provide more information about the test you did to allow us to reproduce your result and use this as a basis? Thanks.

suppamax:
I didn't notice it before as I was testing it only with Verilator - I've just bought a Nexys board and you're right: it starts but it gets stuck.
I see a bunch of LEDs turning on and off, and in the terminal I see the first 2 underscores of the _LITEX_ "splashscreen" get printed.
Is this the same behavior you see?
Let me debug a bit...

suppamax:
The workaround is to add the flag `UART_POLLING` but I didn't find the root cause yet.

One more problem I've noticed is that there is a memory initialization failure when I use a Nexys A7 board. I used the Nexys4DDR without changing any parameter, it might be that there are minor differences around the DDR.

I didn't notice any issue when building with the `--integrated-main-ram-size` option.

enjoy-digital:
Hi @suppamax,

thanks for looking at this and for the updates. At which frequency are you running the design? For low frequencies (<50Mhz), it's possible the clocking would need some adaptation in the target. If it's working correctly with `--integrated-main-ram-size`, the issue is probably in LiteX now and not in the wrapper. Using `UART_POLLING` is fine for this initial support, there is probably just a small issue in the interrupt support that we could investigate later.

I'll also do some tests on my side with your updated version.

zarubaf:
@suppamax how do you set cachability/peripheral regions?

suppamax:
@zarubaf is this what you're looking for?
https://github.com/litex-hub/pythondata-cpu-cva6/blob/8d27ff8e635fe8208f12129ba45edd2dd64a7a1a/pythondata_cpu_cva6/system_verilog/cva6_wrapper/cva6_wrapper_pkg.sv#L51

enjoy-digital:
@suppamax: I've been able to reproduce your results on hardware (with the same memtest issue but on a different board). The issue could be related to handling of the wstrb on the AXI and data-width convertion. I'm going to merge and will do more test in simulation.

zarubaf:
> @zarubaf is this what you're looking for? https://github.com/litex-hub/pythondata-cpu-cva6/blob/8d27ff8e635fe8208f12129ba45edd2dd64a7a1a/pythondata_cpu_cva6/system_verilog/cva6_wrapper/cva6_wrapper_pkg.sv#L51

Ah indeed, that looks good. Thanks. Happy to see that.

enjoy-digital:
@suppamax: I just did some cleanup on top of your last commits: https://github.com/enjoy-digital/litex/commit/808f29ed2e92430ac0ab4e6bbe13b912cb59c164 and https://github.com/enjoy-digital/litex/commit/e1c132809ee4e0dd15ee220720cea75d3b25f234.

In simulation, the some line returns are missing and the echo of the console seems to be delayed by one character. I did a quick observation of the AXI signals but haven't seen anything particular. The issue could be related to cached/uncatched regions configuraiton mismatch between CVA6 and LiteX. I'm pretty sure this issue is also responsible for the memtest failure on hardware.

suppamax:
@enjoy-digital the simulation doesn't work any longer for me: the CPU polls 0x80001804 (at AXI level I see only 0x0000000100000000, so the CPU should read 0x1) and that's it.
Do you see the same behaviour?

enjoy-digital:
@suppamax: I think I saw this yes, the remaining issue seems to manifest itself differently when doing AXI <-> AXI-Lite vs AXI <-> Wishbone. Pressing a key on the console seems to unlock things.

mithro:
I'm still trying to debug an issue with the autoupdate mechnaism @ https://github.com/litex-hub/pythondata-cpu-cva6/runs/6599567940?check_suite_focus=true

suppamax:
@enjoy-digital please have a look at #1313 - any chance that this is also related to the problem with the DDR?

suppamax:
@mithro I'll have a look

mithro:
The `pythondata-cpu-cva6` repository is now being automatically created from https://github.com/openhwgroup/cva6 and being auto-published to GitHub @ https://github.com/litex-hub/pythondata-cpu-cva6 and PyPi @ https://pypi.org/project/pythondata-cpu-cva6/4.2.0.post387/

suppamax:
@mithro Thank you!

@enjoy-digital I've seen that this version has some incompatibility with Verilator. I see that CVA6 uses `#ifndef VERILATOR` to exclude Verilator-unfriendly code: is there a way to define variables when compiling the verilog code?

suppamax:
@mithro Thank you!

PR #1301, opened 2022-05-19
Refactor ghdl code in microwatt and neorv32 core files

umarcor:
This PR is a refactorisation of microwatt and neorv32 `core.py` files:

- Use `pathlib` instead of `os.path`.
- Generate `ys` commands at once.
- Reduce code duplication.

Functionality should be exactly the same.

### DIFF
--- a/litex/soc/cores/cpu/microwatt/core.py
+++ b/litex/soc/cores/cpu/microwatt/core.py
@@ -6,7 +6,7 @@
 # Copyright (c) 2020 Raptor Engineering <sales@raptorengineering.com>
 # SPDX-License-Identifier: BSD-2-Clause
 
-import os
+from pathlib import Path
 
 from migen import *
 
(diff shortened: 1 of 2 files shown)


umarcor:
@enjoy-digital are microwatt and/or neorv32 used/tested in CI?

enjoy-digital:
Thanks @umarcor. I'm sorry but I have a preference for the previous coding style that matches the other coding style practices in LiteX. If we do such coding style changes at some point, we'll make them globally. The sequential `ys` generation is also wanted and matches the .scripts/constraints generation in the build backends. So for now we should probably only integrate the ICP/ICS changes to avoid the duplication. (Sorry for being very open minded about this, I'm just trying to simplify maintenance of the project).

Microwatt and NeoRV32 are not yet tested in CI no, we are planning to do this but haven't had the time yet.

enjoy-digital:
BTW, we've been discussing very recently with @pepijndevos about moving this VHDL -> Verilog convertion steps directly to the build backend. For toolchains not supporting VHDL, we could try to automatically convert VHDL to Verilog and then re-integrate the Verilog in the build flow. This would also avoid duplication of these blocks of code in each VHDL core.

enjoy-digital:
Sorry @umarcor, merging the PR would increase the maintenance burden since I'm not familiar with the coding style introduced. As explained in the previous message, I'm more interesting in reducing duplication in the VHDL -> Verilog conversion and allow it to be automatically done during the build (and available on all build backends).

PR #1302, opened 2022-05-19
litex_setup: fix path reference for python3 binary on non-Linux hosts

tonymcdowell-rs:
the python binary is stored in the "Program Files" directory.  without
delimiting the path the calls to the binary will fail on Windows hosts.


enjoy-digital:
Thanks @tonymcdowell-rs!

PR #1303, opened 2022-05-21
Add request_remaining("name") that returns unallocated pins.

jevinskie:
Improve error reporting on `request_all()`.

This allows me to grab a few LEDs for debugging purposes with `platform.request()` and pass the rest to `LedChaser` with `platform.request_remaining()` instead of `request_all` (which fails) so I still get the handy "proof of life" indicator for the overall design.


jevinskie:
Whoops, closing to split out in a feature branch.

PR #1304, opened 2022-05-21
Quartus build: bail out on error

jevinskie:


enjoy-digital:
Thanks @jevinskie, that's merged.

PR #1308, opened 2022-05-22
Add get_signals function (w/ optional recursion) for litescope usage

jevinskie:
Useful for collecting signals for litescope: e.g. `analyzer_signals = [*get_signals(self.usb, recurse=True)]`

### DIFF
--- a/litex/gen/fhdl/module.py
+++ b/litex/gen/fhdl/module.py
@@ -105,3 +105,13 @@ class LiteXModule(Module, AutoCSR, AutoDoc):
         if module is not None:
             assert isinstance(module, Module)
         return module
+
+    def get_signals(self, recurse=False):
+        """
+        Return public Signals/Record fields from the current module.
+
+        Args:
+            recurse (bool): Also collect Signals/Record fields from submodules.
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @jevinskie, I'll have a closer look at it.

enjoy-digital:
Thans @jevinskie and sorry for the delay.

I rebased/reworked this on current master and force-pushed the branch.

The original idea is now integrated as:

  enjoy-digital/litex@6e7535eb51dc84956c6baffc0692449fe0c91ac7

The implementation now adds:

  - `litex.gen.fhdl.utils.get_signals(obj, recurse=False)`
  - `LiteXModule.get_signals(recurse=False)`

Compared to the initial version, it now preserves signal order, removes duplicates, avoids walking arbitrary `dir()`/properties, supports Records and simple containers, can recurse through registered named/unnamed submodules, and has cycle protection.

I also added regression tests for the helper and checked it with:

  ```sh
  pytest -q test/test_fhdl_utils.py test/test_hierarchical_verilog.py
 ```

PR #1310, opened 2022-05-23
ClockFrequency() to get/set the frequency of a domain or signal

jevinskie:
`ClockFrequency(cd_or_signal="sys", set_freq=None)`

Can be used to verify correctness parameters like with the Intel chip ID functionality that specifies the clock should be <= 100 MHz.

```
if ClockFrequency() > 100e6:
    logging.warn(f"Altera Chip ID block should be clocked {colorer('<= 100 MHz')}, not " +
        colorer(f"{ClockFrequency()/1e6:.0f} MHz")
    )
```

Right now it is only integrated into Intel clocking, but if you like the idea I'll put it in all the clockers.

### DIFF
--- a/litex/soc/cores/clock/common.py
+++ b/litex/soc/cores/clock/common.py
@@ -8,6 +8,7 @@ import logging
 import math
 
 from migen import Record
+from migen.fhdl.structure import ClockDomain
 
 from litex.soc.integration.soc import colorer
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @jevinskie. Having the frequency attached to signals/clock domains would indeed be useful in several cases (in your for example but also in some cases to decide which logic to insert, if a clock is faster than another, in which order do a CDC + Conversion etc...) I've been thinking about adding this for quite a bit of time but haven't spend the time to properly think of it. This has to be done very carefully (since I'm also trying to decouple LiteX from Migen as much as possible), so I'll wait to have more time to look at it.

PR #1311, opened 2022-05-23
AlteraQuartusToolchain: Add ip_dir attribute pointing to IP directory

jevinskie:


enjoy-digital:
Thanks for the contribution, this still looks useful.

I updated the PR against current `master` and resolved the conflict with:
https://github.com/enjoy-digital/litex/pull/1311/commits/35bdb06557b0784fd2629b648a73ea2937c51003

The main adjustment was to keep the new `AlteraQuartusToolchain.ip_dir` attribute, but avoid making Quartus mandatory just to instantiate the toolchain. So `ip_dir` is now populated when Quartus is found in `PATH`, and left as `None` otherwise. This keeps dry-run generation and tests working on machines without Quartus installed.

I also added unit tests for both cases. The PR is now updated and should be mergeable once CI is green.

PR #1315, opened 2022-05-26
liteeth: Add DMA interface

sergpolkin:


enjoy-digital:
Thanks a lot @sergpolkin. This is an interesting feature, but if possible in the future I'd like to be able to discuss a such feature first in issue or by mail: This makes things smoother and can avoid architecture/implementation issues.

The code seems clean but seems to have a design issue: When using a DMA vs SRAM, the bandwidth is not guaranteed on the DMA MMAP bus. So each ethernet packet will probably need to be stored in a Packet FIFO before being transmitted to Ethernet PHY (TX) or written to memory (RX). Without it, it will probably work in most case but will start failing when the main bus of the SoC will not be able to provide enough bandwidth for the DMA. Using LiteX's PacketFIFO could be enough. I'll have a closer look in the next weeks.

sergpolkin:
Thanks you too, for this project.

I'm test this with RMII phy on two boards, with SDRAM (16 bit data width) and DDR3 (128 bit) memory.
The `LiteEthMACCore` module has CDC fifo, so this stuff some how works.

PR #1318, opened 2022-06-02
Add litex.gen.fhdl.verilog.VerilogTime to emit $time

jevinskie:


enjoy-digital:
Thanks @jevinskie (and sorry for the delay...), I just updated/reworked this PR on top of current `master` with:

  https://github.com/jevinskie/litex/commit/841dbf7adae288c00335584b1efea2aaef08d467

 The feature is now kept minimal and limited to Verilog `$display` generation: `VerilogTime()` emits the raw Verilog `$time` expression when used as a `Display()` argument.

 Example:

  ```python
  Display("time=%t value=%d", VerilogTime(), value)
  ```
  now generates:

  $display("time=%t value=%d", $time, value);

 I also added a regression test to make sure $time is emitted unquoted.

PR #1321, opened 2022-06-04
cpu/vexriscv: Redo csr instruction fix

zeldin:
Instead of enabling the zicsr extension in crt0.S, change -march to
specify that VexRISCV implements version 2.0 of I, rather than the
latest (2.1).  In 2.0 the csr instructions were still part of I.
This approach has two advantages:

* It is compatible with older versions of binutils, since they do not
  need to know about the new zicsr extension

* By modifying the -march in CFLAGS, csr instructions can be used in any
  code (for example by the inline functions in irq.h), not just in crt0.S.

### DIFF
--- a/litex/soc/cores/cpu/vexriscv/core.py
+++ b/litex/soc/cores/cpu/vexriscv/core.py
@@ -46,30 +46,30 @@ CPU_VARIANTS = {
 # GCC Flags ----------------------------------------------------------------------------------------
 
 GCC_FLAGS = {
-    #                               /-------- Base ISA
-    #                               |/------- Hardware Multiply + Divide
-    #                               ||/----- Atomics
-    #                               |||/---- Compressed ISA
-    #                               ||||/--- Single-Precision Floating-Point
-    #                               |||||/-- Double-Precision Floating-Point
-    #                               imacfd
-    "minimal":          "-march=rv32i      -mabi=ilp32",
-    "minimal+debug":    "-march=rv32i      -mabi=ilp32",
-    "lite":             "-march=rv32im     -mabi=ilp32",
-    "lite+debug":       "-march=rv32im     -mabi=ilp32",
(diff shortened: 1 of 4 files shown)


Dolu1990:
Ahhhh nice, i wasn't aware of rv32i2p0 syntax XD
Thanks ^^

tcal-x:
@zeldin , @Dolu1990  --- did anyone report this breaking their compile?   I'm trying to bump LiteX in my build, including this change, and it's breaking my link step related to floating point emulation routines:

```
/media/tim/GIT/google/CFU-Playground/env/conda/envs/cfu-common/bin/../lib/gcc/riscv32-elf/10.1.0/../../../../riscv32-elf/bin/ld:
src/tensorflow/lite/kernels/internal/quantization_util.o: in function `tflite::CheckedLog2(float, int*)':
/media/tim/GIT/google/CFU-Playground/proj/proj_template_v/build/src/tensorflow/lite/kernels/internal/quantization_util.cc:399: 
undefined reference to `__extendsfdf2'
```
(and many more)

If I just change `-march=rv32i2p0_m` back to `-march=rv32im`, then it works.

* Is it likely an issue with the toolchain I'm using (a Conda package), or
* Am I missing something else I need to change ?

Thanks in advance for any hints!

zeldin:
@tcal-x Interresting.  I wonder if multilib selection gets screwed up because of the unusual combination.
Could you please test if `-march=rv32i2p0_m2p0` works better?

zeldin:
Anyway, I am unable to reproduce here using a toolchain built with Gentoo `crossdev`, so whatever the issue is it needs to be investigated with your specific toolchain (or build environment in general), I think.
```
hakua:/tmp% riscv32-unknown-elf-gcc -march=rv32i2p0_m -nostartfiles -o riscy riscy.c
hakua:/tmp% nm riscy | fgrep extend
000100a0 T __extendsfdf2
hakua:/tmp% 
```

PR #1322, opened 2022-06-04
software/common.mak: Disable _FORTIFY_SOURCE

zeldin:
If the toolchain is a hosted one, it may have fortification turned on
by default.  Make sure to turn it off so that we don't end up with
undefined references to e.g. write().


enjoy-digital:
Thanks @zeldin and sorry for the delay...

I updated this on top of current `master` with:

  https://github.com/zeldin/litex/commit/98e7046b0d8d94cdb2f04bcdee11939fc3ccb42d

The change is still useful for hosted/hardened toolchains where `_FORTIFY_SOURCE` can be enabled by default and introduce fortified libc references into LiteX bare-metal firmware builds.

Since `common.mak` has changed since the original PR, the override is now applied in `BASEFLAGS`, so it covers both C and C++ firmware code:

  ```make
  -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0
  ```

I also added a small makefile regression check.

PR #1324, opened 2022-06-06
Quartus: Remove bashism from win32 bat script and add shebang on Linux

jevinskie:
Sorry for (probably? unsure about `set` command on windows) breaking win32 in #1304.


enjoy-digital:
Thanks @jevinskie.

PR #1326, opened 2022-06-08
Updated OpenOCD stream for expr use 

kaolpr:
In latest OpenOCD version (0.11.0, 2022-03-25) when using `litex_server` in JTAG mode a noisy warning is issued by OpenOCD saying that form of `expr` use is deprecated. This PR fixes use of `expr` in generated `stream.cfg` file.


enjoy-digital:
Thanks @kaolpr!  I just did a test to verify that it still works with older version of OpenOCD, which seems to be the case. So let's merge, thanks.

PR #1329, opened 2022-06-13
Misc fixes to Altera PLL support

Icenowy:
This PR contains two small fixes to Altera/Intel PLL support, one is fixing N multiplier is ignored when generating ALTPLL instance, another is adding `derive_pll_clocks` to SDC to allow PLLs to be analyzed by TimeQuest.

### DIFF
--- a/litex/build/altera/quartus.py
+++ b/litex/build/altera/quartus.py
@@ -88,6 +88,9 @@ def _build_sdc(clocks, false_paths, vns, named_sc, build_name, additional_sdc_co
             tpl = "create_clock -name {clk} -period {period} [get_nets {{{clk}}}]"
             sdc.append(tpl.format(clk=vns.get_name(clk), period=str(period)))
 
+    # Enable automatical constraint generation for PLLs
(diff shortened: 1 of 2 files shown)


jevinskie:
Do we also want the -derive_base_clocks arg and derive_clock_uncertainties command?

https://github.com/litex-hub/litex-boards/blob/e02bee4265832eba259b6a68cdad7a58b908f4ac/litex_boards/platforms/terasic_deca.py#L310

Icenowy:
> Do we also want the -derive_base_clocks arg and derive_clock_uncertainties command?
> 
> https://github.com/litex-hub/litex-boards/blob/e02bee4265832eba259b6a68cdad7a58b908f4ac/litex_boards/platforms/terasic_deca.py#L310

I think both are not needed. For `-derive_base_clocks`, it tries to create a clock for PLL in, but we have already manually created it; for `derive_clock_uncertainties`, do we have clock uncertainties for input clock?

enjoy-digital:

PR #1331, opened 2022-06-16
Gowin: Fix `copyfile` paths

cr1901:
When copying from `impl/pnr/project.fs`, we are _already_ in the `build` directory.

So including those path components in the `copyfile` function will point to nonexistent files. This PR fixes this.

I'm probably gonna keep this branch around locally and add a few quality of life fixes in addition to this PR.


enjoy-digital:
Thanks @cr1901, this is merged.

PR #1333, opened 2022-06-16
bios/main: Wrap CONFIG_NO_BOOT around boot_sequence()

gsomlo:
This eliminates a "defined but unused" warning on the
`boot_sequence()` function if `main.c` is compiled with
`#define CONFIG_BIOS_NO_BOOT`


enjoy-digital:
Thanks @gsomlo, that's merged.

PR #1334, opened 2022-06-17
Move from deprecated Symbiflow to F4PGA

robertszczepanski:
Symbiflow has been renamed to F4PGA and there was new tool created. This PR introduces usage of this tool by importing it directly as a Python library. LiteX doesn't support using something like that as a submodule so I decided to just throw an exception with a helper message about what's wrong if `f4pga` is missing. Let me know what you think about this aproach @enjoy-digital.

For now usage of this tool is introduced for Xilinx only since QuickLogic is not well updated yet but it's in progress. **PR might either wait for QL to get full support or it might get merged in current form where Xilinx uses new approach and QL uses the old one.**

There is also related issue in LiteX Boards that changes symbiflow name to f4pga - [#401](https://github.com/litex-hub/litex-boards/pull/401).


enjoy-digital:
Thanks @robertszczepanski, this looks good. The approach with the exception seems fine for this particular use case (toolchain that is renamed), we can keep this and we'll be able to remove it in the future when we'll consider that everyone has switch (ex 6-months/1-year).

PR #1335, opened 2022-06-19
Move valentyusb install to litex_setup.py

benstobbs:


enjoy-digital:
Thanks @benstobbs, use of this branch of ValentyUSB was initially just temporary and that's the reason it was not installed directly with LiteX, but since it's there for quite some time now, let's do this yes. We'll change this once/if we have a better alternative.

PR #1336, opened 2022-06-20
Extend FSM class

fjullien:


enjoy-digital:
Hi @fjullien and sorry for the delay....

Thanks for the PR. Since then, LiteX has added generic GTKWave savefile support, including FSM state translation filters for regular Migen `FSM` objects. So the functionality is now covered, but differently from the original PR: we use `GTKWSave.fsm_states()` instead of introducing a new FSM subclass.

While reviewing this PR, I found and fixed one missing case in the existing support: when there was only one FSM, the filter file was generated but the `.gtkw` savefile did not reference it. This is fixed with:

  - enjoy-digital/litex@2245d34a198950d4d35d4148d884af7141f27645

The related LiteScope-side request has also been handled separately with an explicit GTKWave export API:

  - enjoy-digital/litescope@4c49853612fddb12089a0cfd82b0d1bd6cecb1ee

So I think we can close this PR as addressed by the current generic GTKWave/FSM support.

PR #1346, opened 2022-06-28
build/xilinx/ise: Fix yosys flow

cklarhorst:
The top name changed in 2016 but only XST was changed.


enjoy-digital:
Thanks @cklarhorst!

PR #1352, opened 2022-07-02
build/lattice/trellis: Add MCLK frequency option

gregdavill:
Add options for increasing the SPI config MCLK frequency for nextpnr_ecp5/trellis.

Because the frequencies are somewhat odd I've also added a basic assert that lists the valid options.

### DIFF
--- a/litex/build/lattice/trellis.py
+++ b/litex/build/lattice/trellis.py
@@ -257,7 +257,7 @@ def trellis_argdict(args):
         "ignoreloops":  args.nextpnr_ignoreloops,
         "bootaddr":     args.ecppack_bootaddr,
         "spimode":      args.ecppack_spimode,
-        "freq":         float(args.ecppack_freq),
+        "freq":         float(args.ecppack_freq) if args.ecppack_freq is not None else None,
         "compress":     args.ecppack_compress,
         "seed":         args.nextpnr_seed,
     }


enjoy-digital:
Thanks @gregdavill. This has been applied manually since the build backends have been reworked by @trabucayre. Can you verify this has been applied correctly and close the PR if so? Thanks.

gregdavill:
That looks to be working nicely! Thanks.

gregdavill:
Whoops, spoke too soon. I'm seeing an error parsing `None` into a `Float`. Specifically on projects when I've not set a frequency.

```
  File "./orbtrace_builder.py", line 112, in <module>
    main()
  File "./orbtrace_builder.py", line 101, in main
    builder.build(**trellis_argdict(args), run=args.build)
  File "/home/runner/work/orbtrace/orbtrace/deps/litex/litex/build/lattice/trellis.py", line 260, in trellis_argdict
    "freq":         float(args.ecppack_freq),
TypeError: float() argument must be a string or a number, not 'NoneType'
```

The default frequency is 2.4MHz, So maybe instead of setting the default here to `None` we set it to 2.4?
But that might break code that's manually setting it. 

Screenshot from 2022-07-05 09-01-21

gregdavill:
Adding a check like this when you are parsing as float allows the `None` value to pass through. Which is useful to remain backwards compatible.

```diff
diff --git a/litex/build/lattice/trellis.py b/litex/build/lattice/trellis.py
index 4cea92be..2a03fa4c 100644
--- a/litex/build/lattice/trellis.py
+++ b/litex/build/lattice/trellis.py
@@ -257,7 +257,7 @@ def trellis_argdict(args):
         "ignoreloops":  args.nextpnr_ignoreloops,
         "bootaddr":     args.ecppack_bootaddr,
         "spimode":      args.ecppack_spimode,
-        "freq":         float(args.ecppack_freq),
+        "freq":         float(args.ecppack_freq) if args.ecppack_freq is not None else None,
         "compress":     args.ecppack_compress,
         "seed":         args.nextpnr_seed,
     }
```

enjoy-digital:
Thanks!

PR #1355, opened 2022-07-04
integration/soc Add accessible_region to add_memory_buses

cklarhorst:
Commit Message:
Enables CPUs to know which memory addresses are accessible via
the connected memory_bus.

Explanation:
Current situation around the NaxRiscv:
- `litex_sim --cpu-type naxriscv` working
- `litex_sim --cpu-type naxriscv --integrated-main-ram-size 8192` Hangs, litex does not connect anything to the CPU's memory_bus, but NaxRiscv uses the memory_bus for every main_ram request.
- Any `litex_sim --cpu-type naxriscv --with-sdram --sdram-module=MT41K128M16 --sdram-data-width=16` but with other main_ram origins would also not work because the ioRange of naxriscv is hardcoded.

Idea of this PR: Configure the nax IORange with the help of SoCRegion data.

The following NaxRiscv pythondata changes are necessary for testing:
config/gen.scala:14 replace ioRange with:
```
def ioRange (address : UInt) : Bool = {
  if(args.contains("mem-region-origin"))
    !SizeMapping(BigInt(args("mem-region-origin").toString), BigInt(args("mem-region-length").toString)).hit(address)
  else 
    True
}
```

@enjoy-digital How do you think about this PR?

Additional notes:
- Unfortunately, I was not able to write the Scala code in a cleaner way.
- This PR only changes the ioRange, the FetchRange setting could be improved later as well.


enjoy-digital:
Thanks @cklarhorst, I'll try to look at it soon.

Dolu1990:
Hi ^^

I just see it now, do not hesitate to put a @Dolu1990 in issues related to Naxriscv :)

@cklarhorst
We would also need a PR in https://github.com/litex-hub/pythondata-cpu-naxriscv for your https://github.com/litex-hub/pythondata-cpu-naxriscv/blob/master/pythondata_cpu_naxriscv/verilog/configs/gen.scala#L14

You can open a pr there too, and i can edit it :)

@cklarhorst unrelated, but got debian to work with NaxRiscv. apt install gcc and stuff like it :D

Dolu1990:
Ahh i think one thing which is missing in this PR is the handeling of the netlist hash generation, which need to depend on those added options.

See https://github.com/enjoy-digital/litex/blob/91a115584f25771733373549480f036a5cdbe1bc/litex/soc/cores/cpu/naxriscv/core.py#L193

Idea is to have a unique ID for each netlist configuration, in order to cache them using that unique ID, to avoid regenerating them each time. So you would need to store the accessible_region from add_memory_buses into the NaxRiscv class for later use when it generate the unique ID by hashing the parameters.

cklarhorst:
I put the new parameter in `scala_args` and therefore thought that it will already be hashed.
Looks like the `generate_netlist_name` will be called before the `add_memory_buses` func is called.

I will investigate that.

> I just see it now, do not hesitate to put a @Dolu1990 in issues related to Naxriscv :)

Thanks, I will do so in the future.

cklarhorst:
> I put the new parameter in scala_args and therefore thought that it will already be hashed.
> Looks like the generate_netlist_name will be called before the add_memory_buses func is called.

Ok I was wrong, I think the hashing is already working.
For targets that don't have a memory bus the `add_memory_buses` is not called, therefore scale_args is empty and nothing is hashed in the `generate_netlist_name` func.
For targets with a memory bus, the origin and length is stored in `scale_args` and therefore the `generate_netlist_name `will also change.
(only tested with litex_sim)

You could add a print here (https://github.com/cklarhorst/litex/blob/91a115584f25771733373549480f036a5cdbe1bc/litex/soc/cores/cpu/naxriscv/core.py#L238-L239) to verify it.

Dolu1990:
> I put the new parameter in scala_args and therefore thought that it will already be hashed.

Hooo yes you are right, my bad :)

> Looks like the generate_netlist_name will be called before the add_memory_buses func is called.

Ahhh damned, that would be problematic

> Ok I was wrong, I think the hashing is already working.

Cool ^^

Dolu1990:
So, for the fetchRange todo fix. Here we would need to have the mem-region + the rom+ram which are mapped as peripheral.

The reason why NaxRiscv need to know the fetchRange is because the BTB branch prediction may predict bad stuff which would generate instruction cache line refill with an address to a peripheral (ex UART, SPI), which would break stuff.

@enjoy-digital Would there be way to obtain those information in the core.py ? (rom + ram memory mapping)

cklarhorst:
> The reason why NaxRiscv need to know the fetchRange is because the BTB branch prediction may predict bad stuff which would generate instruction cache line refill with an address to a peripheral (ex UART, SPI), which would break stuff.

I did not check if the SoCRegion class is used for all mem regions (https://github.com/enjoy-digital/litex/blob/50a5e137ff4372cd0fb341d055f486a9839ca35b/litex/soc/integration/soc.py#L63).
But if that is the case, we could extend the mode field to also contain 'x' for regions with executable code.
Example:
ROMRegions would have the mode 'rx', flash would have 'rwx'.

enjoy-digital:
@Dolu1990: We could eventually add the soc as an attribute of the CPU and get the information in the do_finalize with something like:
```
soc.bus.regions["rom"].origin/size
soc.bus.regions["main_ram"].origin/size
```
Would it be useful? Would it be early enough to have access to it in the `do_finalize`?

Dolu1990:
@cklarhorst
> But if that is the case, we could extend the mode field to also contain 'x' for regions with executable code.

Ahh yes, right.

@enjoy-digital 
> Would it be useful? Would it be early enough to have access to it in the do_finalize?

Yes i think that would be good ^^
would it be possible to also have the mode of access ?  (having origin/size/mode)
Then i guess we can just iterate over soc.bus.regions and generate the proper stream of parameters to the scala code generation.

enjoy-digital:
@Dolu1990: We would get directly access to the region, so to all parameters.

cklarhorst:
Todos for FetchRange:
- [ ] Make the SoCRegions accessible from the CPU
- [ ] Add a new mode "x" to the SoCRegions
- [ ] Fix all current mode checks f.e. readonly is checked with mode=='r'. That would not work with the new mode. Could be changed to 'w' not in mode
- [ ] Put the FetchRange into the scale_args
- [ ] Add FetchRange Code to gen.scala

Generally good to have regarding FetchRange
- [ ] Bios interrupt code to print an error on access violations 

Shouldn't we create a new PR for the FetchRange changes?
The changes will touch more code and this ioRange PR affect all users with dram while the FetchRange is only a problem when someone has mem mapped to unusual addresses and tries to run code from that :)

@enjoy-digital 
> We could eventually add the soc as an attribute of the CPU

Where should I hand over the SoC instance to the CPU? Right in the CPU constructor? I think that would be there:
https://github.com/enjoy-digital/litex/blob/ece86a7673983123d4735334aa39d3f993e6b879/litex/soc/integration/soc.py#L981

Dolu1990:
> Shouldn't we create a new PR for the FetchRange changes?
The changes will touch more code and this ioRange PR affect all users with dram while the FetchRange is only a problem when someone has mem mapped to unusual addresses and tries to run code from that :)

Hmm, to me all of this (fetch + io ranges) feel like a single thing, it may just be easier to keep things in one PR (for the NaxRiscv fixes), to keep a simpler tracking.

@cklarhorst If you want i can work on the scala side of things, to provide flexible parameters, allowing to setup multiple region as io / fetch range ?

cklarhorst:
> @cklarhorst If you want i can work on the scala side of things, to provide flexible parameters, allowing to setup multiple region as io / fetch range ?

That would be very kind of you.
The current version now generates multiple `--scala-args=executable-region=(origin, size)` arguments. But I can happily change that to any format you propose.

Dolu1990:
@cklarhorst So i updated NaxRiscv repo with : 
https://github.com/SpinalHDL/NaxRiscv/commit/7c61b646a82e883aee103ed4a7b62d9a484d3362
And your python-data PR with : 
https://github.com/cklarhorst/pythondata-cpu-naxriscv/commit/8b808becb50383a4d213508016f9e8bfe28f5bc7

Basicaly now you can give parameters as : 
--memory-region=base,size,mode

Where base and size should be decimal values, and mode can contains : 
- r : regular memory load
- w : regular memory store
- i : peripheral memory load
- o : peripheral memory store
- x : instruction fetchable (execute)

For instance : 
--memory-region=1,2,rw --memory-region=3,4,rwx --memory-region=5,6,rwxio

no order is required in the mode leters, multiple --memory-region can be specified.

is it good for you ?

Dolu1990:
Note it isn't a scala-arg anymore, so it need to be hashed

cklarhorst:
So I think this PR is now ready for feedback.

Please be aware that you need https://github.com/cklarhorst/pythondata-cpu-naxriscv/commit/8b808becb50383a4d213508016f9e8bfe28f5bc7 for testing.

I tested bios boots for:
-  `litex_sim --cpu-type naxriscv` 
-  `litex_sim --cpu-type naxriscv --integrated-main-ram-size 8192` -> main_ram becomes "IOX"
-  `litex_sim --cpu-type naxriscv --with-sdram --sdram-module=MT41K128M16 --sdram-data-width=16` -> main_ram becomes "RWX"

Things where I'm not sure:
- Mode of the opensbi region -> I made it "IO"
- Mode of the clint region -> I made it "IO"
- Mode of the plic region -> I made it "IO"

Downsides of this approach:
- Each ROM/RAM size change now requires a new naxriscv sbt build

Dolu1990:
@cklarhorst 

clint and plic should be io, that's right, but having a opensbi region seems weird. I mean, from a software perspective yes, but from a hardware design, opensbi is mapped into the main memory (DDR), so it is "encapsuled" in it.
DDR being rwx

> Each ROM/RAM size change now requires a new naxriscv sbt build

Right.

> Currently, I don't understand why the bios doesn't receive interrupts on access violations, which would make the bios boot up process easier to debug.

hmm, it maybe possible that the error pin comming from the wishbone isn't connected to the SpinalHDL soc part.

Ahhh yeah, it isn't connected => 
https://github.com/SpinalHDL/SpinalHDL/blob/b8bb31bec8321cf20da2f0cc2db0da9baad1249d/lib/src/main/scala/spinal/lib/bus/bmb/BmbToWishbone.scala#L54

I will fix it.

cklarhorst:
```
IO Regions: (1)
io0                 : Origin: 0x80000000, Size: 0x80000000, Mode: RW, Cached: False Linker: False
Bus Regions: (6)
rom                 : Origin: 0x00000000, Size: 0x00020000, Mode: RX, Cached: True Linker: False
sram                : Origin: 0x10000000, Size: 0x00002000, Mode: RWX, Cached: True Linker: False
opensbi             : Origin: 0x40f00000, Size: 0x00080000, Mode: RW, Cached: True Linker: True
csr                 : Origin: 0xf0000000, Size: 0x00010000, Mode: RW, Cached: False Linker: False
clint               : Origin: 0xf0010000, Size: 0x00010000, Mode: RW, Cached: False Linker: False
plic                : Origin: 0xf0c00000, Size: 0x00400000, Mode: RW, Cached: False Linker: False
```
I wonder what the concrete definition of that linker column is. As the opensbi region is the only one that has Linker=True, does that column maybe indicate that a specific region is only defined for software reasons?

Then I would filter out all regions with Linker=True for naxriscv + I think LiteX should also make sure that there is some real region encapsulating it. With that knowledge, the above example should be invalid, I think.

I also forgot those IO Regions -.-. Do you know the purpose of them?

> Ahhh yeah, it isn't connected =>
https://github.com/SpinalHDL/SpinalHDL/blob/b8bb31bec8321cf20da2f0cc2db0da9baad1249d/lib/src/main/scala/spinal/lib/bus/bmb/BmbToWishbone.scala#L54
>
>I will fix it.

That is so great. It will make boot debugging a fun experience.

Dolu1990:
> I also forgot those IO Regions -.-. Do you know the purpose of them?

So memory load/store can be of two sorts : 
- Cached : To access regular memory => rw
- Uncached : To access peripheral as UART / SPI / ... => io

So SocRegion with Cached == True => rw, while if Cached == False => io

> Linker

I have no idea what that does. It maybe a way to specify that's some software related stuff, not hardware.

cklarhorst:
> So memory load/store can be of two sorts :
> 
> Cached : To access regular memory => rw
> Uncached : To access peripheral as UART / SPI / ... => io
>
> So SocRegion with Cached == True => rw, while if Cached == False => io

My current understanding:
```
LiteX main? Bus <-> NaxRiscV Peripheral
       |
       | - ROM
       | - RAM 
       | - CSR
       | - L2 - LiteDRAM

LiteX direct memory bus <-> NaxRiscV "regular" AXI -> !not connected to LiteX main bus!
       | - LiteDRAM
```
So at least in the present situation the cached column can't be used to decide between RW/IO as everything is IO except sometimes the main_ram can also be accessed with RW (from the standpoint of which bus to use).

So naxriscv's RW/IO currently decides about which AXI to use and whether the access is cached in I/D or not?

Dolu1990:
> My current understanding:

Yes i think you are right.

> So at least in the present situation the cached column can't be used to decide between RW/IO as everything is IO except

I think that's right too

> So setting naxriscv's RW/IO mode currently decides about which AXI to use and whether the access is cached in I/D or not?

So mostly yes, with one exception is that the instruction are always cached, IO setting on instruction cache will only specify which memory bus to use.

Load/Store => 
- RW : Cached + ram_dbus
- IO   : Uncached + peripheral_dbus + memory ordering between IO fully preserved

Fetch => 
- RW : Cached + ram_ibus
- IO   : Cached + peripheral_ibus

Dolu1990:
Seems like i have issues running opensbi with the changes. I will investigate ^^

Dolu1990:
Ahh i think the issues is that the opensbi region is emited in the Region list, and also, it is defined as a IO, confusing things i guess : 
[info] LitexMemoryRegion(40f00000 80000,io)
[info] LitexMemoryRegion(f0c00000 400000,io)
[info] LitexMemoryRegion(f0010000 10000,io)
[info] LitexMemoryRegion(0 20000,ix)
[info] LitexMemoryRegion(10000000 2000,iox)
[info] LitexMemoryRegion(40000000 10000000,rwx)
[info] LitexMemoryRegion(80000000 2000,io)
[info] LitexMemoryRegion(f0000000 10000,io)


Investigating more.

cklarhorst:
I think the "IO" setting from the overlapping opensbi region overrides the RW setting for the main_ram region, because currently IO takes priority before RW.

cklarhorst:
Just an Idea:

Couldn't we just add a new character to the mode field 'c' for cached regions? 
That way, we could also mark regions on the peripheral_bus as cached.
I think that would improve the performance for accessing the ROM/RAM/SRAM.

Dolu1990:
> Could you please recheck with the new commit.

I did an equivalent change localy on my side, and then linux boot was good ^^

So your change is good too.

> Couldn't we just add a new character to the mode field 'c' for cached regions?
I think that would improve the performance for accessing the ROM/RAM/SRAM.

Yes right, this would boost the perf on it.
Now i have a doubt. @enjoy-digital Does memory buses added as following have access to the wishbone memory mapped ram/rom ? Or only the DDR ?

```python
    def add_memory_buses(self, address_width, data_width):
        dbus = axi.AXIInterface(
            data_width    = nax_data_width,
            address_width = 32,
            id_width      = 4,
        )
        self.memory_buses.append(dbus)
```

enjoy-digital:
Hi @Dolu1990,

the buses added in add_memory_buses only have direct access to the DRAM, if doing specific connection between the CPU and some components (ROM/SRAM) could boost performance, I could have a look to try to add this.

cklarhorst:
@Dolu1990 
https://github.com/enjoy-digital/litex/blob/153182a01442ffd5f1548db76e79bc49a2280103/litex/soc/cores/cpu/naxriscv/core.py#L40

I still don't understand what this is about.
Are those regions listed in io_regions (in this case only 0x8000_0000) considered CPU internal? With the meaning that those regions aren't allowed to be used by litex, and that it is granted that load/stores in that region never leave the CPU?

Dolu1990:
> if doing specific connection between the CPU and some components (ROM/SRAM) could boost performance, I could have a look to try to add this.

@enjoy-digital  Ahhh, don't bother ^^
Thanks :D

> Are those regions listed in io_regions (in this case only 0x8000_0000) considered CPU internal? With the meaning that those regions aren't allowed to be used by litex, and that it is granted that load/stores in that region never leave the CPU?

@cklarhorst I'm not sure if there is any usage for it to be honnest. But for sure it isn't something CPU internal. Maybe it is just to say, "if you want to place IO peripherals, they have to be in that range, as the CPU will not cache memory for that range".

cklarhorst:
Uff sorry for this lengthy PR. I should have read https://github.com/enjoy-digital/litex/blob/153182a01442ffd5f1548db76e79bc49a2280103/litex/soc/integration/soc.py#L182 more detailed.

So the regions that naxriscv specifies in https://github.com/enjoy-digital/litex/blob/153182a01442ffd5f1548db76e79bc49a2280103/litex/soc/cores/cpu/naxriscv/core.py#L40 are the **only** regions that are allowed to be uncached. The check is [here](https://github.com/enjoy-digital/litex/blob/153182a01442ffd5f1548db76e79bc49a2280103/litex/soc/integration/soc.py#L224-L230).

In my option, that means naxriscv should cache (I&D cache) everything except its own specified `io_region`. There is no more information needed for the caching part in the sbt build process.
The fetchable/executable range is solved with this PR I think.
Lastly, naxriscv have to decide which bus (main/peripheral) to use for accesses (I&D). Currently, this PR uses the mode field in the `memory-region` parameter, where RW means main and IO means peripheral.

With all that, I would propose to change the `memory-region` format into `(base,size,mode,bus)` with the allowed values:
`mode` subset of [rwx]
`bus` = m or p

Dolu1990:
@cklarhorst Thanks ^^, I think it is right to add the information of "on which bus it is".

Just before implementing it we have to be sure such addition would be ok for @enjoy-digital, as the PR start to have quite a few changes ^^
Also, one thing to consider is to not break legacy stuff.
So maybe having this new field with an "unknown" value by default, which would trigger an assert in NaxRiscv/core.py

enjoy-digital:
@cklarhorst, @Dolu1990: Thanks, the changes related to LiteX looks fine for now, so feel free to continue.

Dolu1990:
@cklarhorst So, i think specifying on whch bus which address range should be accessed is a good idea (instead of implying it from the IO attribute)

Then, adding the support in NaxRiscv for the data cache to be able to cache values from the peripheral bus is kind of costly in area, for something only used in the bootloader / when not running on DDR, so, not so usefull ?
It cost : 
+ 1 AXI4 decoder (1->2)
+ 1 AXI4 to AXI4lite converter
+ 1 AXI4lite memory master to the peripheral bus

So my proposal would be that we could add that specification in litex, use it in NaxRiscv to properly specify where to access things, but let the load/store not being able to cache values from the peripheral bus in order to save area.

Does that sound good ?

cklarhorst:
> @cklarhorst So, i think specifying on whch bus which address range should be accessed is a good idea (instead of implying it from the IO attribute)
> 
> Then, adding the support in NaxRiscv for the data cache to be able to cache values from the peripheral bus is kind of costly in area, for something only used in the bootloader / when not running on DDR, so, not so usefull ? It cost :
> 
> * 1 AXI4 decoder (1->2)
> * 1 AXI4 to AXI4lite converter
> * 1 AXI4lite memory master to the peripheral bus
> 
> So my proposal would be that we could add that specification in litex, use it in NaxRiscv to properly specify where to access things, but let the load/store not being able to cache values from the peripheral bus in order to save area.
> 
> Does that sound good ?

Yes sound good, I will work on this PR tomorrow.
So as NaxRiscV doesn't support caching over the peripheral bus, is it possible to specify IORegions on the main bus?

Dolu1990:
> I think the caching field of the SoCRegion is more like "is allowed to be cached" 
> while if something is marked IO than that is mandatory

Yes right

> is it possible to specify IORegions on the main bus?

Main bus => DDR ? =>  no it isn't possible, the load/store unit will use the peripheral bus for io region.
To explaine a bit more : 
Fondamentaly, in NaxRiscv there is 4 data path : 
- Fetch line refill (splited in two bus in the Scala toplevel, one going to the memory, one to the peripheral bus)
- Data cache refill (memory bus)
- Data cache writeback (memory bus)
- Load/Store Unit io access (peripheral bus by nature)

> So if it is not possible to specify IORegions on the main bus, I might add an assertion for that.

Yes


So we would get stuff like : 

```
io-region: --memory-region=0x80000000,0x80000000,io,p
csr:       --memory-region=0xf0000000,0x00010000,rw,p
rom:       --memory-region=0x00000000,0x00020000,rxc,p
ram:       --memory-region=0x40000000,0x00200000,rwxc,m
``` 

(note i added the "c" for cacheble)

let's me know if things look wrong ^^
else i will also implement support for this arguement syntax

Dolu1990:
So, i pushed the NaxRiscv and python-data changes in your repo.
On my side all seems good, but i guess this CI will not be happy until the python data is merged, but that will break the litex upstream ci XD

cklarhorst:
So I tested all initial use cases, and they are now working.
Thanks for all the assistance I learned a lot. 

Maybe it is also interesting for you. With the current changes, the memspeed shows the following results:

With memory bus (`litex_sim --cpu-type naxriscv --with-sdram --sdram-module=MT41K128M16 --sdram-data-width=16 --l2-size 0`):
NaxRiscV:
  Write speed: 2.2MiB/s
   Read speed: 1.2MiB/s
VexRiscV:
  Write speed: 725.6KiB/s
   Read speed: 261.6KiB/s

With p bus (`litex_sim --cpu-type naxriscv --integran-ram-size 8192`):
NaxRiscV:
Memspeed at 0x40000000 (Sequential, 8.0KiB)...
  Write speed: 322.4KiB/s
   Read speed: 7.7KiB/s
VexRiscV:
  Write speed: 1.6MiB/s
   Read speed: 915.0KiB/s

I have not investigated that further.

Dolu1990:
> Maybe it is also interesting for you. With the current changes, the memspeed shows the following results:

Thanks :D

> Read speed: 7.7KiB/s

Amagad XD That one is weirdly slow, but hooo well. As long that the system work well with the DRAM :)

Also, got some mp4 file to run in debian / ffmpeg : 
https://drive.google.com/file/d/1vkZyLLNxZY0leKUpz68gmWzAOyOXUsxO/view?usp=sharing

Thanks for your help too :D

Dolu1990:
@enjoy-digital Ready to merge, you can do it after your released, to not release a so fresh feature ^^

Dolu1990:
hmm i just tested on my side the read speed on io memory, and i get 
  Write speed: 318.6KiB/s
   Read speed: 269.5KiB/s

So not sure what happend on your case with the Read speed: 7.7KiB/s

cklarhorst:
@Dolu1990 
I just found the problem. It's in this line:
https://github.com/enjoy-digital/litex/blob/15f72174cef1bfa24d43f8654b697e740300d540/litex/soc/cores/cpu/naxriscv/core.py#L126-L127
So on the litex side the p bus interface is always 64bit wide, but I think on the nax side it has xlen size. This (among other problems) leads to bus timeouts on my mem_speed test => So it reads one word too much out of the memory area.

With a data_width of 32, the new speed for accesses on the p bus becomes:
```
Write speed: 430.3KiB/s
Read speed: 467.9KiB/s
```

I'm not quite sure how I should solve that in general.
Should we force the p bus to always have a data_width equal to xlen? Or is there a better solution?

Dolu1990:
> Should we force the p bus (the AXILiteInterface) to always have a data_width equal to xlen? 

Ahhh yes you are right, that's currious that this went through XD

> And I think the wishbone data width also needs to be forced to xlen size. 

Yes right, if possible that would be good. 

Typicaly, one current issue is that with xlen=64, the CLINT is still 32 bits, which make the CPU sometime reading wrong values for the timer when it overflow 32 bits

Let's fix that in another PR :D

>  Or does a 64bit nax somehow support 32bit atomic accesses?

Atomic in the meaning of AMOADD access for instance ? they are done inside the CPU itself so far (single memory master), and proper atomic support for multiple master agents would require memory coherency infrastructure. But atomic access for 64 bits load on 32 bits timer (CLINT) is problematic)

enjoy-digital:
Thanks @cklarhorst, @Dolu1990, I just did minor cosmetic changes on top while reviewing: https://github.com/enjoy-digital/litex/commit/ded3bad1781b37af1eada47276b313f7f47ec0bd.

Dolu1990:
@cklarhorst Found something which slow down the peripheral bus, with xlen=64 each read is translated into  two wishbone transaction, wasting something around 4 cycles.

I'm now looking around to improve this.

cklarhorst:
@Dolu1990 
I started to prepare a new branch for the p-bus data_width fixes and got assertions for a wishbone size of 64bit.

So with pbus width == xlen == wishbone == 64 I got assertion errors in:
```
 File "/litex/soc/interconnect/axi/axi_lite_to_wishbone.py", line 20, in __init__
```

I will open a draft PR, let's move the discussion to that one.

PR #1360, opened 2022-07-06
build/microsemi/libero_soc: fix missing self for _format_io_constraint and tcl_name

trabucayre:
With the move to `GenericToolchain` some reference to `_format_io_constraint` and `tcl_name` have not been updated to reference to the class instance.
This PR fix these errors by adding `self` each time it's required.


enjoy-digital:
Thanks @trabucayre!

PR #1362, opened 2022-07-07
remote: usb: fix multi-word reads and writes

zeldin:
One of the debug traces correctly added the address offset, but not the other one and none of the actual code.  :smile:


enjoy-digital:
Thanks @zeldin!

PR #1369, opened 2022-07-13
cores/clocks/lattice_ecp5: Allow manual placement

zeldin:
`nextpnr-ecp5` always picks the PLL BEL that is closest to the clock input pin, which in my design is on the left side.  However, all my high-speed I/Os are on the right side, and I need to use the right side PLL BEL in order to satisfy timing constraints on ECLK.  (The delay from the input clock on the other hand is completely irrelevant, it's just a frequency reference.)  The easiest way I have found to accomplish this is to hardcode the BEL name `X70/Y49/EHXPLL_LR`.  So I need a way for Litex to let me do that.


enjoy-digital:
Thanks @zeldin!

PR #1375, opened 2022-07-18
[cores/neorv32] add CPU cache and interrupt management functions

stnolting:
This PR add cache flush functions to `system.h` and interrupt management functions to `irq.h` (using same function names as [`litex/soc/cores/cpu/ibex`](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/ibex/irq.h)).


enjoy-digital:
Thanks @stnolting!

PR #1383, opened 2022-07-21
Litex soc gen develop

jwfaye:
Add the possibility to have more than one interface for core communication.

### DIFF
--- a/litex/tools/litex_soc_gen.py
+++ b/litex/tools/litex_soc_gen.py
@@ -4,6 +4,7 @@
 # This file is part of LiteX.
 #
 # Copyright (c) 2022 Florent Kermarrec <florent@enjoy-digital.fr>
+# Copyright (c) 2022 Joseph Wagane Faye <joseph-wagane.faye@insa-rennes.fr>
 # SPDX-License-Identifier: BSD-2-Clause
 
 """
--- a/litex/tools/litex_soc_gen.py
+++ b/litex/tools/litex_soc_gen.py
@@ -62,56 +62,66 @@ class Platform(GenericPlatform):
 # LiteX SoC Generator ------------------------------------------------------------------------------
 
 class LiteXSoCGenerator(SoCMini):
-    def __init__(self, name="litex_soc", sys_clk_freq=int(50e6), **kwargs):
+    def __init__(self, name="litex_soc", sys_clk_freq=int(50e6), n_mi=3, n_si=0, **kwargs):
+        assert (n_mi + n_si) < 5
+        self.kwargs = kwargs
+
         # Platform ---------------------------------------------------------------------------------
-        platform = Platform(device="", io=get_common_ios())
-        platform.name = name
-        platform.add_extension(get_uart_ios())
+        self.platform = Platform(device="", io=get_common_ios())
+        self.platform.name = name
+        self.platform.add_extension(get_uart_ios())
 
         # CRG --------------------------------------------------------------------------------------
         self.submodules.crg = CRG(
-            clk = platform.request("clk"),
-            rst = platform.request("rst"),
+            clk = self.platform.request("clk"),
+            rst = self.platform.request("rst"),
         )
 
         # SoC --------------------------------------------------------------------------------------
-        if kwargs["uart_name"] == "serial":
-            kwargs["uart_name"] = "uart"
-        SoCCore.__init__(self, platform, clk_freq=sys_clk_freq, ident=f"LiteX standalone SoC - {name}", **kwargs)
+        if self.kwargs["uart_name"] == "serial":
+            self.kwargs["uart_name"] = "uart"
+        SoCCore.__init__(self, self.platform, clk_freq=sys_clk_freq, ident=f"LiteX standalone SoC - {name}", **kwargs)
 
-        # MMAP Slave Interface ---------------------------------------------------------------------
-        s_bus = {
-            "wishbone" : wishbone.Interface(),
-            "axi-lite" : axi.AXILiteInterface(),
+        for n in range(n_mi):
+            origins=[0xa000_0000, 0xb000_0000, 0xc000_0000, 0xd000_0000]
+            self.add_master_interface(mmap_id=n, origin=origins[n])
 
-        }[kwargs["bus_standard"]]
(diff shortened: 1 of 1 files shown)


litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
 # LiteX SoC Generator ------------------------------------------------------------------------------
 
 class LiteXSoCGenerator(SoCMini):
-    def __init__(self, name="litex_soc", sys_clk_freq=int(50e6), **kwargs):
+    def __init__(self, name="litex_soc", sys_clk_freq=int(50e6), n_mi=3, n_si=0, **kwargs):
+        assert (n_mi + n_si) < 5
```
enjoy-digital, line 67: Is this assert arbitrary? If yes, it's probably better to remove it, if not it's better to put a comment.

litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
...
+        self.kwargs = kwargs
```
enjoy-digital, line 68: Not sure this is required, it seems to be used to get the bus standard, but you can get it with `self.bus.standard`.

litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
...
+
         # Platform ---------------------------------------------------------------------------------
-        platform = Platform(device="", io=get_common_ios())
-        platform.name = name
-        platform.add_extension(get_uart_ios())
+        self.platform = Platform(device="", io=get_common_ios())
```
enjoy-digital, line 71: Please don't change code that don't require changes.

litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
...
+        self.platform.name = name
+        self.platform.add_extension(get_uart_ios())
 
         # CRG --------------------------------------------------------------------------------------
         self.submodules.crg = CRG(
-            clk = platform.request("clk"),
-            rst = platform.request("rst"),
+            clk = self.platform.request("clk"),
```
enjoy-digital, line 77: Same as previous.

litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
...
+            rst = self.platform.request("rst"),
         )
 
         # SoC --------------------------------------------------------------------------------------
-        if kwargs["uart_name"] == "serial":
-            kwargs["uart_name"] = "uart"
-        SoCCore.__init__(self, platform, clk_freq=sys_clk_freq, ident=f"LiteX standalone SoC - {name}", **kwargs)
+        if self.kwargs["uart_name"] == "serial":
```
enjoy-digital, line 82: Same as previous.

litex/tools/litex_soc_gen.py
```diff
@@ -62,56 +63,66 @@ def build(self, fragment, build_dir, build_name, **kwargs):
...
+            self.kwargs["uart_name"] = "uart"
+        SoCCore.__init__(self, self.platform, clk_freq=sys_clk_freq, ident=f"LiteX standalone SoC - {name}", **kwargs)
 
-        # MMAP Slave Interface ---------------------------------------------------------------------
-        s_bus = {
-            "wishbone" : wishbone.Interface(),
-            "axi-lite" : axi.AXILiteInterface(),
+        for n in range(n_mi):
+            origins=[0xa000_0000, 0xb000_0000, 0xc000_0000, 0xd000_0000]
```
enjoy-digital, line 87: If we expose the number of interface, we should also probably allow configuration of base addresses. We can wait to address the other points before looking at it.

litex/tools/litex_soc_gen.py
```diff
@@ -122,17 +133,21 @@ def main():
...
     target_group.add_argument("--build",         action="store_true", help="Build SoC.")
     target_group.add_argument("--sys-clk-freq",  default=int(50e6),   help="System clock frequency.")
+    target_group.add_argument("--n-master-inter",  default=int(2),   help="Number of master interfaces")
+    target_group.add_argument("--n-slave-inter",  default=int(0),   help="Number of slave interfaces.")
+
     builder_args(parser)
     soc_core_args(parser)
     args = parser.parse_args()
-
+    print("Argument dict {}".format(args))
```
enjoy-digital, line 142: Please remove debug prints.

litex/tools/litex_soc_gen.py
```diff
@@ -122,17 +133,21 @@ def main():
...
     # SoC.
     soc = LiteXSoCGenerator(
         name         = args.name,
         sys_clk_freq = int(float(args.sys_clk_freq)),
+        n_mi = int(args.n_master_inter),
+        n_si = int(args.n_slave_inter),
         **soc_core_argdict(args)
     )
-
```
enjoy-digital, line 135: Please avoid touching formatting when not required.

enjoy-digital (changes requested):
Thanks @jwfaye, this will be useful. We'll maybe rename the parameters and also allow configuration of the exposed regions, but before doing this, can you fix the issues related to changes that are not required.

litex/tools/litex_soc_gen.py
jwfaye: Hi, I add the "self" to, set it as an attribute as "platform" is needed in the "add_slave_interface" and "slave_master_interface" methods. Sorry for the delay by the way. I was on vacation.

litex/tools/litex_soc_gen.py
jwfaye: I set the "assert" as I was thinking about a limitation in the number of interfaces (4 interfaces per standalone SoC.).

litex/tools/litex_soc_gen.py
jwfaye: Okay, I will revise the code and submit another version to you.

PR #1385, opened 2022-07-24
sim: enable relative include paths for verilator

sergachev:
This is required for example to simulate openc906 which uses verilog includes at relative paths with litex_sim.


enjoy-digital:
Thanks @sergachev!

Issue #1313, opened 2022-05-25
CVA6 and AXILiteDownConverter


suppamax:
I think the problem mentioned in the PR #1294 and regarding the need to press a key to let the simulation run relates to this

1. the core executes `uart_sync()`, which means polling the address `(CSR_BASE + 0x1804L)`
2. the CPU interface is 8bytes aligned while the CSR is 4 bytes aligned
3. the read request is converted into 2 reads, using offset 0x1804 and 0x1808 
4. the requested data in the lower half of the databus
5. the requested address is 4 bytes aligned, the CPU has an 8bytes databus, so it expects the data on the higher half of the bus

Masking out the last 3 bits of A*ADDR in `cva6_wrapper.sv` works
```
    assign AWADDR_o   = {master[cva6_wrapper_pkg::External].aw_addr[AxiAddrWidth-1:3], 3'b000};
...
    assign ARADDR_o   = {master[cva6_wrapper_pkg::External].ar_addr[AxiAddrWidth-1:3], 3'b000};
```

Is this acceptable?

enjoy-digital:
Thanks for looking at this @suppamax, I'll have a closer look. This could be better to fix the AXILiteDownConverter's behavior if incorrect.

sergachev:
AXILiteDownConverter is not used here, it's axi64 -> axi32 -> axilite32 -> wb32 with the first conversion being the problem.

One possible solution is here: https://github.com/enjoy-digital/litex/pull/1387

enjoy-digital:
The AXIDownConverter has been improved/fixed and now handle things correctly and allow CVA9 simulation.

PR #1386, opened 2022-07-24
test: add axi 64b to 32b conversion test

sergachev:
After looking at https://github.com/enjoy-digital/litex/issues/1313 and other issues caused by AXI width conversion I wrote these minimal tests that demonstrate problems with the existing AXIDownConverter and AXIUpConverter. The validity of the tests can be checked by using DUT_ref instead of DUT which replaces axi64->axi32->wb32 with axi64->wb64->wb32 configuration - that works fine.

One of the problems with the width converters is here:
https://github.com/enjoy-digital/litex/blob/b9a1fec30fd1f0691f9c6463bac24ea1da75355a/litex/soc/interconnect/axi/axi_full.py#L210

Because AXI awlen/arlen=0 means 1 transfer, 1 means 2 transfers, 2 means 3 transfers and so on (https://developer.arm.com/documentation/ihi0022/c/Addressing-Options/Burst-length) it is not correct to bit-shift the signal to multiply the length by the conversion ratio. Instead one has to do something like `axi_to.aw.len.eq(((axi_from.aw.len + 1) << log2_int(ratio)) - 1)`.
Forcing burst via omit + `axi_to.aw.burst.eq(1)` in addition and doing the same at the read channel makes the downconverter pass these simple tests, but does not make the converter fully correct yet - it did not work with any of the CPUs I tested it with. Rocket CPU in addition fails many assertions with these edits.

### DIFF
--- a/test/test_axi.py
+++ b/test/test_axi.py
@@ -335,3 +335,86 @@ class TestAXI(unittest.TestCase):
             r_valid_random  = 90,
             r_ready_random  = 90
         )
+
+    def test_axi_width_converter(self):
+        class DUT(Module):
+            def __init__(self, dw_from=64, dw_to=32):
+                self.axi_master = axi_master = AXIInterface(data_width=dw_from)
+                self.axi_slave = axi_slave = AXIInterface(data_width=dw_to)
+                converter = AXIConverter(axi_master, axi_slave)
+                self.submodules += converter
+                wb = wishbone.Interface(data_width=dw_to,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @sergachev for the additional test and investigation, let's merge this and switch to DUT_ref for now. We'll be able to switch to DUT locally when investigating and we'll switch back to DUT on CI when things will be fixed.

PR #1387, opened 2022-07-24
cpu/cva6: add optional peripheral bus conversion +

sergachev:
Because of the problem with native AXI bus width conversion (https://github.com/enjoy-digital/litex/pull/1386) I propose to have this optional conversion to wishbone and enable it by default for now.

In addition I fixed add_jtag() and removed a bunch of 'self's.


enjoy-digital:
Thanks @sergachev, let's allow both yes and defaults to Wisbone until we fix AXI Converter.

PR #1389, opened 2022-07-25
RFC: yosys nextpnr wrapper

trabucayre:
For devices toolchains there is three posibilities:
1. a full Open Source toolchain using **Yosys** for *synthesis*, **nextpnr** for *PNR* and a *packer*. With potentially a converter between *PRN* and *packer*
2. a full proprietary toolchain
3. a mix where synthesis is done by *Yosys*, rest by proprietary tools.

This PR add two wrapper:
1. `YosysWrapper`: it is used to fill a **.ys**, with default/specifics commands and options and to provides a string use to execute this application. Template may be provided to replace default.
2. `nextPNRWrapper`: used to forge the command line and to return it

To provides a solution to case 1 (full Open Source toolchain) an additional class is added: it inherits from `GenericToolchain` and must be inhereted by specifics toolchain (**icestorm**, **trellis**, **oxide** and **XilinxYosysNextPRN**)

To provides a solution to simplify case 3 (**radiant**, **vivado**, **ise**) `YosysWrapper` is directly used and command string is added at script begin (before proprietary tools call).

*build/xilinx/yosys_nextpnr* has been rewriten to produce a script instead of a Makefile.

Two point requires some discussions:
1. **yosys** options are more or less always provided by dedicated toolchains -> maybe kwargs must be partly consumed by this class to extract at least user options. Same question for target used to "forge" `synth_xxx` command: knowledge about **yosys** specifics stuff may (or not) left to this class instead of requiring a bit of knowledge to specific subclass.
2. `NextPNRWrapper` is only used by `YosysNextPNRWrapper` and has only one goal: producing the command to call through script. A dedicated class is maybe not the right way.


mithro:
@trabucayre There is also `Yosys` --> `VPR` flow which is part of F4PGA to support Xilinx Series 7 and potentially ultrascale tooling.

trabucayre:
@mithro true! I have missed to mention this sequence handled by **F4PGA** which is similar to the first point (same global sequence, with a different PNR tool) but where the wrapper is an external tool instead of a custom set of classes.

enjoy-digital:
Thanks a lot @trabucayre! I'll do a close review of this and will get back to you.

enjoy-digital:
@trabucayre: I've been able to have a closer look and do some tests. I really like the direction it's going and de-duplication so think we can merge this now, and discuss the remaining points in different issues/PRs. Thanks!

PR #1390, opened 2022-08-01
cores/cpu/vexriscv_smp: add default cores used by linux with l2 cache

tpwrules:
This prevents needing to install `sbt` etc. for the demos in linux-on-litex-vexriscv which use the Wishbone interface and the L2 cache.


Dolu1990:
Look good to me ^^

PR #1391, opened 2022-08-04
build/sim/verilator: fixed missing placeholder

dlobato:
Fst tracing stopped working since: https://github.com/enjoy-digital/litex/pull/1381 
It added a new variable but missed to add an extra placeholder on the formatted string.


enjoy-digital:
Thanks @dlobato!

PR #1392, opened 2022-08-05
cores/cpu/vexriscv_smp: define SYNTHESIS in Quartus

tpwrules:
This fixes issues along the lines of `Verilog HDL Unsupported Feature error at ***.v(14192): system function "$urandom" is not supported for synthesis`.

I tested that this issue is fixed on with Quartus 20.1.1 building the `de10nano` example of `linux_on_litex_vexriscv`. I also checked that Vivado 2020.1 building `arty_a7` and Yosys 0.16 building `butterstick` don't need a similar fix. I also tested that the regular `vexriscv` CPU doesn't seem to need a similar fix.

This partially addresses https://github.com/litex-hub/linux-on-litex-vexriscv/issues/295


enjoy-digital:
Thanks @tpwrules!

PR #1393, opened 2022-08-05
build/xilinx/vivado: Insert the yosys call into script_content only when synth_mode == yosys

trabucayre:


enjoy-digital:
Thanks @trabucayre!

PR #1395, opened 2022-08-09
Add missing soc/cores/cpu directories to MANIFEST.in

lschuermann:
Adds some missing directories under `soc/cores/cpu` to the `MANIFEST.in` file. This prevents them from being removed when installing LiteX as a Python package.

For what its worth, this list was curated through `find litex/soc/cores/cpu/ -type d -maxdepth  1 | sort | sed -e 's/^/graft /'`.


mithro (approved):
LGTM!

PR #1403, opened 2022-08-18
integration/soc: fix sata irq initialization

gsomlo:
The first argument to `self.irq.add()` should match the name of the
`EventManager()` object being added, i.e., "sata_irq" rather than
just plain "sata". This is necessary for interrupt signals to be
asserted as intended.


gsomlo:
tested successfully (on nexys-video with rocket/linux), so I applied & closed this PR

PR #1405, opened 2022-08-23
soc/integration/builder: more precise error message when meson is too old

jrudolph:


enjoy-digital:
Thanks a lot @jrudolph!

PR #1406, opened 2022-08-26
quartus warnings pll instantiation.

svhb1000:
fixed quartus warnings using cyclone4 pll, 
and using reset signal pll_reset for clarity, the reset seemed not to be connected anyway in the original code. 

Rather using a dedicated name for the reset looks to create less confusion according to me.

### DIFF
--- a/litex/soc/cores/clock/intel_common.py
+++ b/litex/soc/cores/clock/intel_common.py
@@ -28,7 +28,7 @@ def geometric_mean(vals):
 class IntelClocking(Module, AutoCSR):
     def __init__(self, vco_margin=0):
         self.vco_margin = vco_margin
-        self.pll_reset  = Signal()
+        self.reset  = Signal()
         self.locked     = Signal()
         self.clkin_freq = None
         self.vcxo_freq  = None
@@ -108,7 +108,7 @@ class IntelClocking(Module, AutoCSR):
             p_OPERATION_MODE         = "NORMAL",
             i_INCLK                  = Cat(self.clkin, 0),
             o_CLK                    = clks,
-            i_ARESET                 = self.pll_reset,
+            i_ARESET                 = self.reset,
             i_CLKENA                 = 2**self.nclkouts_max - 1,
             i_EXTCLKENA              = 0xf,
             i_FBIN                   = 1,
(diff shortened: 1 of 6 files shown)


enjoy-digital:
Hi @svhb1000,

thanks. What was the warning and what it the fix for it in the code change? I would recommend doing only one specific change per PR. The reset signal are also consistent between the different PLLs, so we should not change it (unless deciding to do so on all the PLLs.

svhb1000:
Hi, 

I aggree that the "pll_reset" should be reverted back to "reset". Despite I like it a bit more specific, since reset sounds more like a general reset for the FPGA. Keep in mind that the reset signal was anyhow not connected to the ARESET input of the PLL. 

The warnings fro the Quartus tool is about sizes of signals conneted to the PLL : 
`Warning (12030): Port "CLK" on the entity instantiation of "ALTPLL" is connected to a signal of width 1. The formal width of the signal in the module is 6.  The extra bits will be left dangling without any fan-out logic.
Warning (12010): Port "CLKENA" on the entity instantiation of "ALTPLL" is connected to a signal of width 5. The formal width of the signal in the module is 6.  The extra bits will be driven by GND.
Warning (12010): Port "INCLK" on the entity instantiation of "ALTPLL" is connected to a signal of width 1. The formal width of the signal in the module is 2.  The extra bits will be driven by GND.
`

Should I make you a new pull-request with the 'pll_reset' signal renamed to 'reset' again? I'm new to GIT, it's a bit a struggle to get things right (more a mercurial fan for > 15 years).

enjoy-digital:
Hi @svhb1000,

thanks for the warning message. I'll study thing along with the code change. I could handle the changes myself but for future PR, I indeed have a preference for 1 PR per specific change, this is easier to review and maximize the chance to get things reviewed/integrated quickly :)

svhb1000:
The warnings only have to do with some signal-witdh mismatches, not critical and there will be nothing wrong with the FPGA itself. I'm just starting to look at Litex, Migen and VexRiscV, it's really a beatifull approach to shape some open source hardware, and very versatile for all kind of different boards. 
It's however a quite heavy learning curve to get some grip on the details (Python, C, verilog, SpinalHDL, Migen, ... all mixing in my head)

PR #1410, opened 2022-08-29
Improve connectors

trabucayre:
This PR try to provides a solution to comments in [PR](https://github.com/litex-hub/litex-boards/pull/418).
- the first commit add a serie of method, similar to `add_extension`, to allows extending connectors dictionary dinamicaly
- second commit add, when a connector reference another connector, a recursive way to retrieve pin name.

### DIFF
--- a/litex/build/generic_platform.py
+++ b/litex/build/generic_platform.py
@@ -132,6 +132,12 @@ def _resource_type(resource):
 class ConnectorManager:
     def __init__(self, connectors):
         self.connector_table = dict()
+        self.add_connector(connectors)
+
+    def add_connector(self, connectors):
+        if isinstance(connectors, tuple):
+            connectors = [connectors]
+
         for connector in connectors:
             cit       = iter(connector)
             conn_name = next(cit)
@@ -192,6 +198,9 @@ class ConstraintManager:
 
     def add_extension(self, io):
         self.available.extend(io)
+
+    def add_connector(self, connectors):
+        self.connector_manager.add_connector(connectors)
 
     def request(self, name, number=None, loose=False):
         resource = _lookup(self.available, name, number, loose)
@@ -355,6 +364,9 @@ class GenericPlatform:
 
     def add_extension(self, *args, **kwargs):
         return self.constraint_manager.add_extension(*args, **kwargs)
+
+    def add_connector(self, *args, **kwargs):
+        self.constraint_manager.add_connector(*args, **kwargs)
 
     def finalize(self, fragment, *args, **kwargs):
         if self.finalized:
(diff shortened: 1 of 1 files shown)


litex/build/generic_platform.py
```diff
@@ -132,6 +132,9 @@ def _resource_type(resource):
 class ConnectorManager:
     def __init__(self, connectors):
         self.connector_table = dict()
+        self.add_connectors(connectors)
+
+    def add_connectors(self, connectors):
```
enjoy-digital, line 137: Maybe we should have a `add_connector` and move the `for connector in connectors` loop in __init__?

litex/build/generic_platform.py
```diff
@@ -193,6 +199,9 @@ def __init__(self, io, connectors):
     def add_extension(self, io):
         self.available.extend(io)
 
+    def add_connectors(self, connectors):
```
enjoy-digital, line 202: Same: `add_connector`?

litex/build/generic_platform.py
```diff
@@ -356,6 +365,9 @@ def add_platform_command(self, *args, **kwargs):
     def add_extension(self, *args, **kwargs):
         return self.constraint_manager.add_extension(*args, **kwargs)
 
+    def add_connectors(self, *args, **kwargs):
```
enjoy-digital, line 368: Same: add_connector?

enjoy-digital (changes requested):
Thanks @trabucayre, this will be very useful. The exposed method should however probably only add one connector at once, so call it`add_connector` and do the appropriate change (mostly move for loop).

litex/build/generic_platform.py
trabucayre: I think testing `connector` type is more easy, avoid the overhead to call this method multiple time and allows to pass a serie of connectors: this method may be used with a scalar (tuple) or with a list (as done for **lite** variant for the **sipeed Tang Primer 20k** dock):

```
if isinstance(connector, tuple):
    connector = [connector]
```

Thanks to that this method will be more generic and avoid `[]` boilerplate when only one connector must be added.

trabucayre:
> Thanks @trabucayre, this will be very useful. The exposed method should however probably only add one connector at once, so call it`add_connector` and do the appropriate change (mostly move for loop).

In fact I have used `connectors` with **s** because it's called like this everywhere. But if you prefer without I will change that.
The first motivation to this method is to answer case like tang primer where 4 pmods and 2 connectors must be added.

enjoy-digital:
Thanks @trabucayre, we could keep `connectors` in `__init__` (for retro compatibility), but I would have a preference to have a `add_connector` method.

trabucayre:
Done

PR #1412, opened 2022-09-02
litex/build/xilinx/f4pga: update imports

umarcor:
Some modules/functions were reorganised in f4pga (see https://github.com/chipsalliance/f4pga/pull/620). This PR updates the imports.


enjoy-digital:
Thanks @umarcor.

PR #1414, opened 2022-09-02
ci: run on ubuntu-22.04 and cleanup

umarcor:


enjoy-digital:
Thanks @umarcor, we'll probably be able to apply some of the changes. I'll have a closer look at this after doing the 2022.08 release and will also need to update the scripts of the others repos.

enjoy-digital:
Just for info @umarcor, I merged https://github.com/enjoy-digital/litex/pull/1515 from @shenki since was introducing less modification and easier to merge, but I'm still planning to merge the interesting addition/cleanup you made.

enjoy-digital:
Thanks @umarcor for this cleanup, and sorry this stayed open for so long.

I had another look at it against current master. Most of the useful parts have since been integrated or superseded by later CI updates: the workflow now runs on Ubuntu 22.04, uses newer GitHub actions, has the package cleanup, and has evolved quite a bit with sharding/caching/toolchain setup.

The one remaining idea that could still be useful is adding `workflow_dispatch` to allow manually triggering the CI. I’ll handle that separately if we decide to add it.

So I’m going to close this PR as obsolete, but thanks again for the work here.

PR #1418, opened 2022-09-03
build/lattice/platform: lattice_args, lattice_argdict: refactorize toolchains args

trabucayre:
Currently toolchains args are managed at targets level. This imply code dupplication and sometime missed args for a toolchain (for example with lattice nexus). FPGAs may have one toolchain (**ice40**, **efinix**), two (**ecp5**, **nexus**) or more (*xilinx* serie7 with **vivado**, **nextpnr** and **f4pga**).
This approach mean, when one wish to add a new toolchain for an existing device, an update must be done at **litex-boards** level to reflect alternate toolchain, with potential forget.

The idea of this PR is to add a two functions, at `build/xxx/platform.py` level, taking family model:
- first: fills argparse by calling all supported toolchain for family FPGAs and, when necessary, adds `toolchain` argument
- second: according to `family` and `toolchain` arguments fills the dict with corresponding values.

Thanks to that **targets** level is simplified because argparse is correctly filled transparently and all build/xxx updates will be automatically available at **litex-boards** level. This reduce redundancy because these task are written once and use everywhere and, finally, one wishing adding new toolchain (**apicula** or **mistral**) has to update these two functions instead of many **targets** files.

Unfortunately, since **toolchain** is only known after `args.parse` all toolchains args are shown instead of dedicated ones.

This PR isn't to be merge as it: once everyone are agrees with that I have to update all toolchains accordingly.

### DIFF
--- a/litex/build/altera/platform.py
+++ b/litex/build/altera/platform.py
@@ -16,6 +16,8 @@ class AlteraPlatform(GenericPlatform):
     bitstream_ext = ".sof"
     create_rbf    = True
 
+    _supported_toolchains = ["quartus"]
+
     def __init__(self, *args, toolchain="quartus", **kwargs):
         GenericPlatform.__init__(self, *args, **kwargs)
         self.ips = set()
--- a/litex/build/anlogic/platform.py
+++ b/litex/build/anlogic/platform.py
@@ -15,6 +15,8 @@ from litex.build.anlogic import common, anlogic
 class AnlogicPlatform(GenericPlatform):
     bitstream_ext = ".fs"
 
+    _supported_toolchains = ["td"]
+
     def __init__(self, device, *args, toolchain="td", **kwargs):
         GenericPlatform.__init__(self, device, *args, **kwargs)
         if toolchain == "td":
--- a/litex/build/argument_parser.py
+++ b/litex/build/argument_parser.py
@@ -1,0 +1,194 @@
+#
+# This file is part of LiteX.
+#
+# This file is Copyright (c) 2022 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
+# SPDX-License-Identifier: BSD-2-Clause
+
+import argparse
+import importlib
+import sys
+
+from litex.soc.integration.soc_core import *
+from litex.soc.integration.builder import *
+
+# LitexArgumentParser ------------------------------------------------------------------------------
+class LiteXArgumentParser(argparse.ArgumentParser):
+    """
+    ArgumentParser subclass used to intercept parse_args call and to simplify
+    common arguments addition
+    Attributes
+    ==========
+    _platform: GenericPlatform subclass
+        target platform
+    _device: str
+        target device family
+    _args: argparse.Namespace
+        list of args after parse_args call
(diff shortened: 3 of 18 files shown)


enjoy-digital:
Hi @trabucayre,

sorry for the delay.  Thanks for the detailed description, I perfectly agree with the direction. To avoid showing the arguments of all the toolchains, we could also use a similar mechanism than the one we are using for the CPUs (basically 2 stage s parser but that would also needs probably to be improved). We could first do the implementation you are suggesting and in a second time try to only expose the arguments of the selected toolchain.

trabucayre:
Don't worry :)
I'm agree: displaying options for all toolchains instead of ones for the selected toolchain doesn't make sense. I have to explore solution used for CPUs because with only one stage parser the toolchain is only known after parse() ;)

trabucayre:
I have reworked this PR according to your comments/requests.
Now:

First commit:
- each *build/XXX/platform.py* contains a class attribute `_supported_toolchains` with a list of possible *toolchains* to use. This attribute may be a `dict` (**device** <-> **list**) when more than one device is supported by this platform (**lattice** and **xilinx**) or a simple list when no distinction must be done (rest of vendors)
-  three new methods are added: 
   -  `fill_args` to access to the targeted toolchain's `xxx_build_args`
   -  `get_ardict` to access to the targeted toolchain's `xxx_build_argdict`
   -  `toolchain` to get toolchain(s) list according to the (optional) `device` parameter
-   to simplify `device` information: dummies classes has be added for Xilinx (`XilinxSpartan6Platform` and `Xilinx7SeriesPlaform`) and for lattice (`LatticeiCE40Platform`, `LatticeECP5Platform` and `LatticeNexusPlatform`). Each of these dummies class are used to provides *device* name. Since,for all others platforms, a distinction is not (currently) required nothing was added.

Second commit:
It introduce a new class `LiteXArgumentParser` which it picks up idea/base from `LiteXSoCArgumentParser` for argparse `parse_args` intercept:

- At constructor level it create the *Target option* group with the most common options added at *targets* level (**build**, **load** and **toolchain**). For **toolchain** it uses, when the platform is not None, the list of toolchains supported by the device. It also calls `builder_args` and `soc_core_args`. the toolchain may be passed at construction parameter of after this step by using `set_toolchain` or simply `LitexArgumentParserInstance.platform = myPlatform`
- when `parse_args` is called, this class try to search for a `--platform` option if platform is None, then search for a `--toolchain` and inject toolchain args before parsing (if no `--toolchain` is provided by the user default one is injected). lastly, before from `LiteXSoCArgumentParser` is kept.
- since **Target Options** is added by this class a method `target_group` allows to have access to this group to add more option
- `builder_args` and `soc_core_args` are also called internally: two methods `builder_argdict` and `soc_core_argdict` are available to retrieve list of options
- in the same principle a method `toolchain_argdict` allows to have access to the toolchain options according to the selected device and toolchain.

enjoy-digital (approved):
Thanks @trabucayre, awesome work! Let's merge and go further :)

trabucayre:
I'm crossing my fingers that it doesn't break anything ;-)

PR #1423, opened 2022-09-10
Improve DMA

zyp:
I'd like to use the `WishboneDMAWriter` to receive packets from a stream, and this PR addresses two issues I were having:

The first is that originally `sink.ready` were set while idle, effectively blackholing data until the DMA is enabled. I think in general it's more reasonable to apply backpressure until DMA is enabled, so I changed this.

The second is that I've got a packetized stream and would like to preserve the packet boundaries, so I added the ability to stop early when the `last` flag is set. For applications where this behavior is not desired, the `last` flag can be stripped from the stream before passing it to `WishboneDMAWriter`.

### DIFF
--- a/litex/soc/cores/dma.py
+++ b/litex/soc/cores/dma.py
@@ -187,7 +187,7 @@ class WishboneDMAWriter(Module, AutoCSR):
         )
         fsm.act("RUN",
             self._sink.valid.eq(self.sink.valid),
-            self._sink.last.eq(offset == (length - 1)),
+            self._sink.last.eq(self.sink.last | (offset + 1 == length)),
             self._sink.address.eq(base + offset),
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @zyp! The changes make sense and this is merged.

For `sink.ready`, we are in fact probably relying on this behavior in cores using the `WishboneDMAWriter `since this can be useful to avoid stalling a pipeline or flushing it, so I added https://github.com/enjoy-digital/litex/commit/3cd4a3830c852cd34c2181e3b9ff1c6dbd69f258 in top of your changes to keep the previous behavior and allow you to set `ready_on_idle` to 0 to have the expected behavior. 

For `last` signal, this should not cause issues with the actual cores and I'm also fine disconnecting it externally if required.

PR #1425, opened 2022-09-12
NaxRiscv: Force main bus data_width == xlen

cklarhorst:
This fixes the size of the AXI-Interface on the LiteX site (makes it equal to xlen).
So bus accesses for xlen==32 are now correct.
This also increases the throughput to:
`Write speed: 430.3KiB/s
Read speed: 467.9KiB/s`

Xlen==64 is broken:
With `litex_sim --cpu-type naxriscv --xlen 64 --bus-data-width 64` I get:
```
  File "/litex/litex/soc/interconnect/axi/axi_lite_to_wishbone.py", line 20, in __init__
```

### DIFF
--- a/litex/soc/cores/cpu/naxriscv/core.py
+++ b/litex/soc/cores/cpu/naxriscv/core.py
@@ -112,7 +112,7 @@ class NaxRiscv(CPU):
         if args.xlen:
             xlen = int(args.xlen)
             NaxRiscv.xlen                 = xlen
-            NaxRiscv.data_width           = xlen
+            NaxRiscv.p_bus_data_width     = xlen
             NaxRiscv.gcc_triple           = CPU_GCC_TRIPLE_RISCV64
             NaxRiscv.linker_output_format = f"elf{xlen}-littleriscv"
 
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi @cklarhorst 

So, i also used the day to track this, and apparently we are just better staying with 32 bits io peripheral bus. 64 bits peripheral bus seems to break a few things.

I pushed all the required change via :
https://github.com/enjoy-digital/litex/commit/0a380b9c3ba278830dc840314b9930c7eef589a2

That i will merge into master when it pass the regression.
For the CLINT, i added some buffering that only sample the 32 bits MSB of the timer when the 32 bits LSB are read. that way it is fine.

Also, with those change, got the SPI sdcard speed to bump from 405 KB/s to 490 KB/s in linux

Dolu1990:
Another issue with the 64 bits io bus, is that the AXI-lite4 doesn't capture precise read access, i mean :
- Imagine you want to read address 0x1000 to 0x1003 (4 bytes)  using a 64 bits bus.
- AXI-lite do not provide any way to specify that you only need toe lower 4 bytes of the bus
- So, a down sizer 64 bits -> 32 bits will always generate two 4 bytes reads, waisting time (4 cycles round trip in litex)

So many peripherals are designed with a 32 bits bus, i think we are just better of staying on it :)
All good for you if i merge https://github.com/enjoy-digital/litex/commit/0a380b9c3ba278830dc840314b9930c7eef589a2 ?

cklarhorst:
@Dolu1990 Sure merge it. You are right, there are too many stones in the way.

Dolu1990:
Merged :)

PR #1430, opened 2022-09-16
Axi_lite_to_wishbone: Reduce latency, make latency configurable

cklarhorst:
- First commit reduces the latency for the request handover by 1
- Second commit reduces the latency for the rw data handover by 1

Benchmark with NaxRiscv (this uses the Bridge two times (for I and D accesses):
I used `litex_sim --cpu-type naxriscv` for benchmarking, I always run `mem_speed` 5 times and used the last run for comparison.

early_wb_req=False, early_data_handover=False => Default case, old behavior:
```
litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 428.9KiB/s
   Read speed: 466.9KiB/s
```

early_wb_req=True, early_data_handover=False:
```
litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 482.2KiB/s
   Read speed: 530.7KiB/s
```

early_wb_req=True, early_data_handover=True:
```
litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 550.6KiB/s
   Read speed: 614.8KiB/s
```

Current problems:
- I would like to make it configurable per target because I think some boards would benefit from more or less FF stages on the main bus than others, but the bridge creation in liteX is very dynamic. I don't know how I should set the configuration from the target file for this bridge.

Are there further improvements?
- NaxRiscv also increases the latency by two (edit, see below). I made that configurable (https://github.com/cklarhorst/NaxRiscv/commit/b04fdfc4f19d50ff166d67ab882e6d69590a555a) that improved the throughput to:
```
litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 768.7KiB/s
   Read speed: 899.9KiB/s

```
But I was not able to reach timing closure on the arty board.
- This bridge could further buffer the addrs of the AXI side. So that an AXI module could send the next addr earlier. I have a patch for this, but in my tests the NaxRiscv doesn't provide addrs earlier. Having the next addr early would also allow to use bursts on the wb side.

@Dolu1990 Maybe this also increases your sdcard throughput.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -196,6 +196,7 @@ class SoCBusHandler(LiteXModule):
         bursting         = False,
         interconnect     = "shared", interconnect_register=True,
         arbiter          = "default",
+        low_latency      = False,
         reserved_regions = None,
     ):
         self.logger = logging.getLogger(name)
@@ -276,6 +277,7 @@ class SoCBusHandler(LiteXModule):
         self.interconnect          = interconnect
         self.interconnect_register = interconnect_register
         self.arbiter               = arbiter
+        self.low_latency           = low_latency
         self.masters               = {}
         self.slaves                = {}
         self.regions               = {}
@@ -661,7 +663,10 @@ class SoCBusHandler(LiteXModule):
                         colorer(type(master).__name__),
                         colorer(type(slave).__name__)))
                     raise SoCError()
-                bridge = bridge_cls(master, slave)
+                bridge_kwargs = {}
+                if getattr(bridge_cls, "supports_low_latency", False):
+                    bridge_kwargs["low_latency"] = self.low_latency
+                bridge = bridge_cls(master, slave, **bridge_kwargs)
                 self.submodules += bridge
                 return adapted_interface
 
@@ -1281,6 +1286,7 @@ class SoC(LiteXModule):
         bus_bursting         = False,
         bus_interconnect     = "shared",
         bus_arbiter          = "default",
+        bus_low_latency      = False,
         bus_reserved_regions = None,
 
         csr_data_width       = 32,
@@ -1332,6 +1338,7 @@ class SoC(LiteXModule):
             bursting         = bus_bursting,
             interconnect     = bus_interconnect,
             arbiter          = bus_arbiter,
+            low_latency      = bus_low_latency,
             reserved_regions = bus_reserved_regions,
            )
 
@@ -3464,6 +3471,7 @@ class SoCCore(LiteXSoC):
         bus_bursting               = False,
         bus_interconnect           = "shared",
         bus_arbiter                = "default",
+        bus_low_latency            = False,
 
         # CPU parameters.
         cpu_type                   = "vexriscv",
@@ -3545,6 +3553,7 @@ class SoCCore(LiteXSoC):
             bus_bursting         = bus_bursting,
             bus_interconnect     = bus_interconnect,
             bus_arbiter          = bus_arbiter,
+            bus_low_latency      = bus_low_latency,
             bus_reserved_regions = {},
 
             csr_data_width       = csr_data_width,
@@ -3740,6 +3749,7 @@ def soc_core_args(parser, cpu_type="vexriscv", cpu_variant=None):
     soc_group.add_argument("--bus-bursting",      action="store_true",                                                               help="Enable burst cycles on the bus if supported.")
     soc_group.add_argument("--bus-interconnect",  default="shared",                   choices=SoCBusHandler.supported_interconnect,  help="Select bus interconnect.")
     soc_group.add_argument("--bus-arbiter",       default="default",                  choices=SoCBusHandler.supported_arbiter,        help="Select bus arbiter.")
+    soc_group.add_argument("--bus-low-latency",  action="store_true",                                                               help="Enable low-latency bus bridges when available.")
 
     # CPU parameters.
     soc_group.add_argument("--cpu-type",                 default="vexriscv",                 help="Select CPU: {}.".format(", ".join(map(str, cpu.CPUS.keys()))))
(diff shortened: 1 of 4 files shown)


Dolu1990:
> early_wb_req=True, early_data_handover=True:

Ahhh nice :D
That's would be great

> NaxRiscv also increases the latency by one.

Isn't increased by two ? as it add pipelining on both command and responses ?

> But I was not able to reach timing closure on the arty board.

Typicaly, the pipelining on the naxriscv peripheral bus response is kind of very necessary, as it directly goes into some reformating + writes to the register file combinatorialy. Maybe we don't realy need the pipelining on the command channels, not sure.

> I have a patch for this, but in my tests the NaxRiscv doesn't provide addrs earlier. Having the next addr early would also allow to use bursts on the wb side.

Could work on that :)

cklarhorst:
> NaxRiscv also increases the latency by one.
> 
> Isn't increased by two ? as it add pipelining on both command and responses ?

Yes by two :)

> Typicaly, the pipelining on the naxriscv peripheral bus response is kind of very necessary, as it directly goes into some reformating + writes to the register file combinatorialy. Maybe we don't realy need the pipelining on the command channels, not sure.

Nice to know. I will try that out. Thanks

Dolu1990:
@cklarhorst So i did some tests, adding the additional stream with address ahead of data availability.

So instead of 
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 428.9KiB/s
   Read speed: 466.9KiB/s
   
It improve speed to : 
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 468.6KiB/s
   Read speed: 468.6KiB/s

Which kind of make sense, it close the gap with the read speed, but for sure can't go faster than read speed.

Funny thing is that the val special = new Area  { ... } also use the writeback cmd logic to provide the address and the size of the transfer XD, so it cost quite some logic.

So overall, i don't know if it is a good thing to bring the address channel in NaxRiscv if we consider the following : 
- The improvement only apply to io store, and if the store data is late or another store isn't written back yet (corner cases)
- It cost some hardware
- Add some complexity

cklarhorst:
FYI:
With early_wb_req=True, early_data_handover=True and the latex NaxRiscv master we are now at:

```
litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 641.6KiB/s
   Read speed: 730.5KiB/s
```
Thanks for: https://github.com/SpinalHDL/NaxRiscv/commit/7abf9e75fda6c6d05561e32425516ae9d0971870 I will play with it

Dolu1990:
By the way, did you know about  --bus-standard axi-lite ? 

litex> mem_speed 0x10001000 0x500
Memspeed at 0x10001000 (Sequential, 1.2KiB)...
  Write speed: 643.10KiB/s
   Read speed: 734.8KiB/s
   
When the latex NaxRiscv master ^^

Still the PR would be great, as --bus-standard axi-lite would not work so well for peripherals which are only wishbone :)

Dolu1990:
I tried debian with the related performance improvement, gived me around 20 % more speed on both ethernet and sdcard ^^

cklarhorst:
Did you use the latest nax master and `early_wb_req=True, early_data_handover=True` ? With 100mhz?
It seems to not reach timing on all boards I have, but I'm sadly still testing it a bit

Dolu1990:
I use Artix 7 speed grade -1,  latest nax master, with --bus-standard axi-lite which provide similar latency than your PR, and 100 Mhz.
Timings where ok with that configuration.

Does it also break on some of your board with  --bus-standard axi-lite  ?

Dolu1990:
I only have that Artix 7 speed grade -1 board :)

cklarhorst:
>  --bus-standard axi-lite
Nice, this one seems to always reach timing. Maybe because there are fewer converters used? Idk

All in all, the axilite2wishbone converter is now more configurable. I left the default as it was.

With the latest commit, I added an option to pass options to converters. Sadly, you can't configure each converter individually, but maybe there aren't so many users for this feature anyway.
I also made the log slightly more verbose. I think it is important to know about the specific features of a used converter. For example, that some converter doesn't forward/convert error signals. Currently, only the changed converter has a feature list.

@enjoy-digital Are you ok with or have any better ideas about the argparser & logging changes for the bus standard converter? (So basically the last commit in this PR)

Dolu1990:
> Nice, this one seems to always reach timing. Maybe because there are fewer converters used? Idk

likely :)

> Sadly, you can't configure each converter individually

That's always something tricky to do in generic systems. i don't know enough of litex to know if there is already something to deal with it.

enjoy-digital:
I updated the PR to make the option simpler and safer to merge.

There is now a single opt-in option:

  ```bash
  --bus-axi-lite-to-wishbone-low-latency
  ```
The default behavior is unchanged and keeps the existing registered path. The low-latency mode only shortens the AXI response path on Wishbone ACK; the Wishbone command path stays registered. This keeps the useful latency improvement while avoiding the larger timing risk I saw with the early command path.

I also checked this on Arty A7-35 with Vivado 2025.2 and --bus-standard=axi-lite:
  <table>
    <thead>
      <tr>
        <th>Config</th>
        <th>LUTs</th>
        <th>FFs</th>
        <th>BRAM</th>
        <th>DSP</th>
        <th>WNS / TNS</th>
      </tr>
    </thead>
    <tbody>
      <tr>
        <td>default</td>
        <td>4770</td>
        <td>3745</td>
        <td>15.5</td>
        <td>4</td>
        <td>-0.463 / -15.462 ns</td>
      </tr>
      <tr>
        <td>low-latency</td>
        <td>4796</td>
        <td>3713</td>
        <td>15.5</td>
        <td>4</td>
        <td>-0.049 / -0.424 ns</td>
      </tr>
    </tbody>
  </table>

So the resource impact is small and the timing looks acceptable on this target, while still keeping the safer registered behavior as default.

enjoy-digital:
@cklarhorst: This is now merged, thanks!

PR #1435, opened 2022-09-18
yosys_nextpnr_toolchain: add flow3 option to abc9 mode

gsomlo:
Add "flow3" option to abc9 mode. This runs FPGA mapping several times, producing a generally better mapping at the cost of increased runtime (see https://github.com/Ravenslofty/yosys-cookbook/blob/master/ecp5.md).

Also, add a "--yosys-flow3" build option to both "trellis" and "oxide".


enjoy-digital:
Thanks @gsomlo, this is merged.

PR #1436, opened 2022-09-19
Etherbone retransmission

mkurc-ant:
This PR adds retransmissions to the Etherbone host code. Instead of immediately throwing `socket.timeout` exception the modified code retries the request up to 10 times within the specified timeout period. The timeout can be provided in the `CommUDP` class constructor and defaults to 1.0 seconds.

A potential danger I see with this approach is that when a reply packet gets lost on the device to host route and the request was a write the write would be executed multiple times for each retransmission. This however could be mitigated by adding some kind of "sequence counter" field to the Etherbone packets. Then the device would simply ignore duplicated requests with the same sequence number.

### DIFF
--- a/litex/tools/remote/comm_udp.py
+++ b/litex/tools/remote/comm_udp.py
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import socket
+import time
 
 from litex.tools.remote.etherbone import EtherbonePacket, EtherboneRecord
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @mkurc-ant, sorry for the delay: I wanted to have a closer look at the code. It seems https://github.com/enjoy-digital/litex/pull/1458 was a duplicate and has now been merged.

I agree with the multiple writes analysis, the current implementation should enough for now and for the use-cases covered, but we can keep this in mind in case of issues in the future.

PR #1437, opened 2022-09-19
build/yosys,nextpnr, lattice: refactor args

trabucayre:
Toolchains based on *yosys+nextpnr+packer* have more or less same options: in this [PR](https://github.com/enjoy-digital/litex/pull/1435)  3 toolchains have to be updated to add the same option.
*packer* part is target dependant and it make sense to keep tihis at toolchain level, but only a subset of *yosys* and *nextpnr* are hardware dependants: consequently having those at the wrapper level is more logical.

This PR move *yosys* args to `YosysWrapper`, *nextpnr* args to `NextPNRWrapper` and keep only specifics options (primarely packer) at toolchain level. Since *icestorm*, *trellis* and *oxyde* toolchains uses `YosysNextPNRToolchain`, this one is updated too to add an abstraction layer.

### DIFF
--- a/litex/build/lattice/icestorm.py
+++ b/litex/build/lattice/icestorm.py
@@ -17,7 +17,7 @@ from migen.fhdl.structure import _Fragment
 from litex.build.generic_platform import *
 from litex.build import tools
 from litex.build.lattice import common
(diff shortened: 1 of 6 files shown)


enjoy-digital:
Thanks @trabucayre, this looks good!

PR #1438, opened 2022-09-22
Prefix argument added to litex_setup.py

bilalahmed-RS:


bilalahmed-RS:
Yes we don't need it any more.

enjoy-digital:
Thanks @bilalahmed-RS.

PR #1440, opened 2022-09-25
Naxriscv: Add two more argparser options for devs

cklarhorst:
I found both options useful.

@Dolu1990


Dolu1990:
Ahhh --update-repo --no-netlist-cache nice :D

Yes that is usefull, i'm used to modify litex code when doing dev to implement this behaviour, but with this all would go in command line, thanks <3

Issue #1439, opened 2022-09-23
Digital video: blue and red signals seem swapped


suarezvictor:
Digital video output have swapped R and B signals, according to https://en.wikipedia.org/wiki/Digital_Visual_Interface, and to my tests using analog VGA (which has clearer channel assignment and produces the expected colors).
Note also that sync signals should be sent on DATA0 (blue clannel) and not "r" channel as in the current code (change to be tested).

See: https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/video.py#L802
`c2d  = {"r": 0, "g": 1, "b": 2}` should be changed to
`c2d  = {"r": 2, "g": 1, "b": 0}` (same for all other lines where that code appears)

and `self.comb += encoder.c.eq(Cat(sink.hsync, sink.vsync) if color == "r" else 0)`
to `self.comb += encoder.c.eq(Cat(sink.hsync, sink.vsync) if color == "b" else 0)`

trabucayre:
I had noticed this swap too (I don't know why I haven't sent a PR :-/).
Most all the phy must be updated accordingly (and all targets using pn_swap too).

enjoy-digital:
Thanks @suarezvictor, I'll try to integrate your fix in the next days and do the swap in the platforms.

suarezvictor:
Please consider that chaging the sync signal channel may break things, so test are needed. I'll report an issue with that when I can dig more and report a easier reproducible bug

trabucayre:
PR for both litex and litex-boards sent!

enjoy-digital:
Thanks @suarezvictor, @trabucayre. I remember having to swap "r" and "b" on some design in the past and now understand why :)

PR #1442, opened 2022-09-27
soc/cores/video: swap red and blue channel

trabucayre:
As mentionned in issue #1439 *blue* and *red* channels, for **DVI/HDMI**, are swapped.
This PR:
- replaces classes's `c2d` by a global dict to reduce code dupplicate.
- reorders colors
- the color iteration is done using the global dict and `items()` to directly have color and channel
- test  for *TMDS* with/without sync channels, is now based on channel id instead of color name.

Another PR will be opened ASAP for litex_boards if pn_swap must be updated accordingly.

### DIFF
--- a/litex/soc/cores/video.py
+++ b/litex/soc/cores/video.py
@@ -163,6 +163,9 @@ video_data_layout = [
     ("b",     8),
 ]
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #1444, opened 2022-09-29
Vivado: Make directives configurable via argparser

cklarhorst:
+ add option to limit vivado threads.

I'm a big fan of the argparser choice field, but sadly the directives change depending on the vivado version, so the options now accept everything. This change should also not change any defaults!


enjoy-digital (approved):
Thanks @cklarhorst, this looks good.

Issue #1319, opened 2022-06-03
Simplify VHDL -> Verilog conversion flow.


enjoy-digital:
Since VHDL is not well supported by open-source tools, we generally have to convert VHDL design to Verilog before being able to simulate with Verilator or do a synthesis with Yosys.

The VHDL -> Verilog flow has been used to integrate some VHDL design with LiteX but also to simulate MicroWatt and NeoRV32 CPU with verilator.

The conversion code for Microwatt is the following:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/microwatt/core.py#L217-L232

And the one for NeoRV32 the following:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/neorv32/core.py#L178-L192

The flow is similar in both case: GHDL Synth -> Yosys -> Verilog. The generated Verilog file is then added to the LiteX flow as any other Verilog file.

We should move this to `litex.build` to avoid duplications and should also explore GHDL -> Verilog directly (which now seems to be possible).

Once simplified, we should also install GHDL (and eventually Yosys) to CI to be able to simulate Microwatt/NeoRV32 CPUs.

enjoy-digital:
This should now be covered.

The duplicated VHDL -> Verilog flow has been moved to the common `litex.build.vhd2v_converter.VHD2VConverter` helper, which uses GHDL synthesis and can optionally flatten through Yosys when available.

Both Microwatt and NeoRV32 now use this common converter instead of local duplicated conversion code, and CI now provides OSS CAD Suite/GHDL/Yosys and exercises both CPUs in the integration boot tests. Microwatt is also tested with the IRQ variant.

So the original cleanup/CI coverage requested here is now in place, and I think we can close this.

PR #1453, opened 2022-10-05
Vhdl wrapper

trabucayre:
VHDL code is not always supported by synthesis tools. In this situation a conversion step, using **glhdl_yosys_plugin** and **yosys**, is required.
This imply code duplication when an IP is written in VHDL (currently two CPU in LiteX) because:
-  **ysosys** must be called with the list of the IP's files with, optionnaly, a list of `-g` parameters to set generics;
- an `Instance` must be added to instanciate resulting verilog code.

This PR try to provides a solution to issue #1319.
- alls toolchains are updated to provides status about VHDL support
- a new class is added to handle files conversion when required by the toolchain or when user force this step, the Instance may be optionnaly added automatically is requested by the user.
- neorv32 CPU is updated to use this new approach (microwatt must be done but currently conversion fails with new and old approach).

### DIFF
--- a/litex/build/VHDLWrapper.py
+++ b/litex/build/VHDLWrapper.py
@@ -1,0 +1,144 @@
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2022 Gwenhael Goavec-Merou <gwenhael.goavec-merou@trabucayre.com>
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks a lot @trabucayre! This is a very good base to start simplifying VHDL to Verilog conversions. Let's merge and continue the work together.

PR #1458, opened 2022-10-11
Added retransmission logic for EtherBone UDP reads.

tmichalak:
Occasionally we see problems with litex_server failing when it misses a single EtherBone UDP packet.
This PR adds some retransmission logic, so we retry a couple of times before giving up.


enjoy-digital:
Thanks @tmichalak, @mkurc-ant, I also saw it a couple of times. Sorry for the delay to merge it.

PR #1472, opened 2022-10-27
Zynq7000: UART fix

mkuhn99:
The compilation of the bios.elf for the Zynq7000 CPU on the Zybo-Z7-20 Board (using litex-boards from this pull request:  https://github.com/litex-hub/litex-boards/pull/423) fails, because there are two conflicting definitions of the uart-functions (e.g. uart_rxtx_write) in the generated csr.h and in the system.h of the zynq7000 cpu. 

Therefore I made the uart-functions in the genereated csr.h optional and don't use them for the zynq.

But when running the bios on the zybo-board it prints the message "BIOS CRC failed (expected 00000000, got 731aced4)" and I cannot send to the board and the last line is "sL5DdSMmkekro" instead of "Litex>".


trabucayre:
I'm not sure this is the correct way to deal with the **zynq7000** if you check **zedboard**,  [digilent_arty_z7](https://github.com/trabucayre/litex-boards/blob/arty_z7_bios/litex_boards/targets/digilent_arty_z7.py) (WIP) or **zynqmp** based boards: all informations are provided through `xparameters.h` filled into `finalize` method.
If you hardcode UART controller it's imply all zynq7000 based boards MUST use this peripheral (it's not always true).

I have to finalize digilent_arty_z7 before opening a PR

trabucayre:
I have created a [PR](https://github.com/litex-hub/litex-boards/pull/443) for the digilent arty Z7 board: it's works.
See PR message for a way to load bios and bitstream

One possible reason why your bios stall during boot is an access to a blank PL (PS wait forever for an answer).
I have too a CRC error but doesn't seems to be an issue.

trabucayre:
Great!

Issue #1459, opened 2022-10-11
Generate high-level CSR block information for automated dumping


staticfloat:
It would be useful for debugging if the generated CSR headers would define some high-level CSR block information.  I'm imagining something like the following:

```
struct csr_block {
    const char * name;
    uint32_t addr;
    uint32_t length;
};
```

And then that can be filled into the generated headers, allowing easy dumping of the full CSR space from a C program rather easily.  We have a somewhat-manual approach being developed on [this PR](https://github.com/JuliaComputing/xtrx_julia/pull/121/files) but it would be nice if we didn't have to hardcode these values.

enjoy-digital:
Hi @staticfloat,

the CSR block information can be exported to .json with `--csr-json=csr.json`:
```

litex_sim --csr-json=csr.json
cat csr.json
{
    "csr_bases": {
        "ctrl": 4026531840,
        "identifier_mem": 4026533888,
        "timer0": 4026535936,
        "uart": 4026537984
    },
    "csr_registers": {
        "ctrl_reset": {
            "addr": 4026531840,
            "size": 1,
            "type": "rw"
        },
        "ctrl_scratch": {
            "addr": 4026531844,
            "size": 1,
            "type": "rw"
        },
        "ctrl_bus_errors": {
            "addr": 4026531848,
            "size": 1,
            "type": "ro"
        },
        "timer0_load": {
            "addr": 4026535936,
            "size": 1,
            "type": "rw"
        },
        "timer0_reload": {
            "addr": 4026535940,
            "size": 1,
            "type": "rw"
        },
        "timer0_en": {
            "addr": 4026535944,
            "size": 1,
            "type": "rw"
        },
        "timer0_update_value": {
            "addr": 4026535948,
            "size": 1,
            "type": "rw"
        },
        "timer0_value": {
            "addr": 4026535952,
            "size": 1,
            "type": "ro"
        },
        "timer0_ev_status": {
            "addr": 4026535956,
            "size": 1,
            "type": "ro"
        },
        "timer0_ev_pending": {
            "addr": 4026535960,
            "size": 1,
            "type": "rw"
        },
        "timer0_ev_enable": {
            "addr": 4026535964,
            "size": 1,
            "type": "rw"
        },
        "uart_rxtx": {
            "addr": 4026537984,
            "size": 1,
            "type": "rw"
        },
        "uart_txfull": {
            "addr": 4026537988,
            "size": 1,
            "type": "ro"
        },
[...]

PR #1473, opened 2022-10-28
[WIP] Add CSR metadata generation

staticfloat:
This creates a new section in `csr.h` that contains metadata about all CSRs contained wtihin the SOC build. This is useful for projects that wish to do some amount of introspection, such as dumping all known CSRs to a file.

This implements one possible solution to https://github.com/enjoy-digital/litex/issues/1459.  I am marking it as a draft until I have internally verified it working in our designs.

It is intended to be used by CSR-dump-like routines, such as the following:

```C
void dump_litex_regs(const char * filename) {
    FILE *f = fopen(filename, "w");
    if (!f) {
        printf("Failed to open file %s)", filename);
        return;
    }
    uint32_t last_region = (uint32_t)-1;
    for (uint32_t csr_idx=0; csr_idx < CSR_METADATA_LEN; csr_idx++) {
        // Print out region headers
        CSRRegion region = CSR_REGIONS[CSR_METADATA[csr_idx].region];
        if (last_region != CSR_METADATA[csr_idx].region) {
            last_region = CSR_METADATA[csr_idx].region;
            fprintf(f, "[%s]\n", region.name);
        }

        // Dump all CSR values
        printf("%s: 0x%02x\n", CSR_METADATA[csr_idx].name, 
                               CSR_EXTRACT(litepcie_readl(fd, CSR_METADATA[csr_idx].addr), CSR_METADATA[csr_idx]));
    }
    fclose(f);
}
```

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -198,6 +198,19 @@ def _get_csr_addr(csr_base, addr, with_csr_base_define=True):
     else:
         return f"{hex(csr_base + addr)}L"
 
+def get_busword_type(busword_len):
+    num_bytes = busword_len//8
+    if num_bytes > 8:
+        return None
+    elif num_bytes > 4:
+        return "uint64_t"
+    elif num_bytes > 2:
+        return "uint32_t"
+    elif num_bytes > 1:
+        return "uint16_t"
+    else:
+        return "uint8_t"
+
 def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, read_only, csr_base, with_csr_base_define, with_access_functions):
     r = ""
 
@@ -208,17 +221,10 @@ def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, read_onl
     r += f"#define {size_str} {nwords}\n"
 
     size = nwords*busword//8
-    if size > 8:
+    ctype = get_busword_type(nwords*busword)
+    if ctype is None:
         # Downstream should select appropriate `csr_[rd|wr]_buf_uintX()` pair!
         return r
-    elif size > 4:
-        ctype = "uint64_t"
-    elif size > 2:
-        ctype = "uint32_t"
-    elif size > 1:
-        ctype = "uint16_t"
-    else:
-        ctype = "uint8_t"
 
     stride = alignment//8;
     if with_access_functions:
@@ -245,7 +251,7 @@ def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, read_onl
     return r
 
 
-def get_csr_header(regions, constants, csr_base=None, with_csr_base_define=True, with_access_functions=True):
+def get_csr_header(regions, constants, csr_base=None, with_csr_base_define=True, with_access_functions=True, with_csr_introspection=True):
     alignment = constants.get("CONFIG_CSR_ALIGNMENT", 32)
     r = generated_banner("//")
     if with_access_functions: # FIXME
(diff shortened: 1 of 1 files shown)


staticfloat:
This seems to be working, but I'm not 100% certain this implementation is the best way to do this, I'm open to suggestions for improvement.

mithro:
@mateusz-holenko who did the Renode and DeviceTree generation parts might have opinions.
@xobs who did the SoCDoC and a bunch of Rust & SVD related generation might also have opinions.

xobs:
This does seem very useful! Having a device able to introspect its own registers is frequently very handy.

I do have some minor bikeshedding questions -- for example, should the `region` index be called something like `region_index` instead? Or does it make sense to replace `uint32_t` with `int`, `long`, or `size_t` instead so that it works on platforms that are not 32-bits? I wouldn't want to block merging because of this, it's just something that I notice and should be noted.

I think the bigger issue is the integration, and correct me if my interpretation of this is wrong, of "fields" with "csrs" into a single `CSRMetadata` struct. I believe it combines the two into the same array, which means iterating through the CSR metadata would result in reading registers multiple times. This could be an issue with certain registers where multiple reads invoke actions multiple times.

It might be better to break fields into its own array / struct and add a `csr_index` that specifies which `CSRMetadata` a particular field belongs to, or to annotate the `CSRMetadata` struct with something like `is_field` that determines if a particular metadata entry is a field or a whole csr.

staticfloat:
> should the region index be called something like region_index instead?

:+1:

> does it make sense to replace uint32_t with int, long, or size_t instead so that it works on platforms that are not 32-bits?

The problem is that these values can be used by host-side code as well.  For instance, in my project, these headers are consumed by a host-side program that uses PCIe to pull down the CSRs and dump them out to a file, so the most important thing is that the field offset datatypes do _not_ change depending on the configuration of the compiler.  :P

I think perhaps we would need to get the bus width of the target processor, then choose an appropriate `uintX_t`, and use that consistently.  I'm not sure if such a type definition already exists, but I'm open to suggestions here. :)

> I believe it combines the two into the same array, which means iterating through the CSR metadata would result in reading registers multiple times. This could be an issue with certain registers where multiple reads invoke actions multiple times.

Hmmm, yes I agree with you.  That does seem like a problem, and I think the right thing to do is to split the fields out into a third array, so we can have a table of regions, a table of CSRs, and a table of fields.  The fields will contain an index back into the table of CSRs, just like the CSRs contain an index back into the table of regions. I'll adjust this PR when I have time.

staticfloat:
I have pushed a new commit that implements everything brought up so far.  While I have taken a bit more care to deal with things like register sizes and whatnot, I feel that area is still quite incomplete, because this is intentionally not defining the interface for getting the registers; it just defines where to look for them.  This means that there's likely to be a mismatch between between the width of the CSR, the width of the communication channel between the host and the device, and the width of the host registers.  Because most users are probably fetching CSRs from the device through things like `litepcie_readl()` or similar, the interplay between all three sizes can be a little tricky.

In my example code that I included within the commit, I do things like the following:

```
// Read the CSR register. Note we need to explicitly handle word size here, since
// `litepcie_readl()` always returns a `uint32_t`...
unsigned long long word = 0;
for (int word_idx=0; word_idx < (csr.size + 8*sizeof(uint32_t) - 1)/(8*sizeof(uint32_t)); word_idx++) {
    word = (word << 8*sizeof(uint32_t)) | litepcie_readl(fd, csr.addr + sizeof(uint32_t)*word_idx);
}
```

This works well for me, but of course, for others it may not.  If they have CSRs that are 128 bits wide, this will truncate.

On a separate note, I am also not certain what the interplay between the memory region bus width, versus the CSR size is.  I suspect that the bus width limits the size of what a single memory access can address, limiting the device to host communication size I mentioned above.

I don't want to paralyze this PR with perfectionism, but I wonder if there aren't already tools/standards that exist in litex that I am unaware of that can help guide these pieces.

PR #1475, opened 2022-10-29
vhd2v: Use GHDL directly

shenki:
As of April 2022, GHDL can output Verilog directly without the use of yosys. This simplifies the build environment for flows that simply want to convert VHDL to Verilog.

Note that this removes some arguments from ghdl that are not required to synthesise ether Microwatt or neorv32.

### DIFF
--- a/litex/build/vhd2v_converter.py
+++ b/litex/build/vhd2v_converter.py
@@ -62,10 +62,10 @@ class VHD2VConverter(Module):
         self._force_convert = force_convert
         self._add_instance  = add_instance
 
-        self._ghdl_opts    = "--ieee=synopsys -fexplicit -frelaxed-rules --std=08 "
+        self._ghdl_opts     = ["--std=08", "--no-formal"]
+
         if work_package is not None:
-            self._ghdl_opts += f"--work={self._work_package} "
-        self._ghdl_opts += "\\"
+            self._ghdl_opts.append(f"--work={self._work_package}")
 
     def add_source(self, filename):
         """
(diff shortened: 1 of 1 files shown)


trabucayre:
LGTM.
Just a small question: depending on distro / user, ghdl may or not supporting `--synth` option, being too old, or a version before april 2022.
Why not keeping both solutions and (for example) using `ghdl --help` to know which solution to use?

It's okay: this imply more code at in a future may be removed.

shenki:
> Why not keeping both solutions and (for example) using `ghdl --help` to know which solution to use?

I considered this, but there's a few complications. Microwatt's use of `is_X` was only supported by ghdl as of July (https://github.com/ghdl/ghdl/issues/2129). And neorv32 would crash ghdl until yesterday (https://github.com/ghdl/ghdl/issues/223) (ghdl now includes part of neorv32 in it's test suite!). So we would end up with a series of special cases depending on the version.

The motivation for this work was to simplify the requirements for CI to build microwatt. We could keep both synth types if they will be used, but I would err on the side of simplifying the code and reducing the test matrix. What do you think?

enjoy-digital:
Thanks @shenki, @trabucayre. That's also something I wanted to test. Since there are only a few users for this feature, requiring a change to a newer version of GHDL seems fine; not sure we should keep the GHDL/Yosys mode if GHDL alone can now do the same.

trabucayre:
@enjoy-digital : If you are okay to move to a direct convert by GHDL I'm okay too: code looks good.
I have asked about dual support only because GHDL's last stable version (v2.0.0.) was released in March: before direct synthesis was ready. For people using GHDL from distribution's package manager this may an issue: but with a correct message to explain potential issue it's fine too :)

enjoy-digital:
Thanks @trabucayre, having to use the development version of GHDL until the next stable release seems fine since Microwatt/NeoRV32 are requiring a recent version of GHDL. Thanks a lot @shenki, this is merged.

PR #1476, opened 2022-10-29
riscv: Fix compilation with new binutils

shenki:
The csr opcodes are no longer part of the i instruction set, and must be enabled separately. This can be done by adding _zicsr to the march string, eg. -march=rv32i_zicsr. However this is not recognised by older toolchains, so we can't change over until everyone is using binutils 2.28 or later.

An alternate fix was merged for Vexriscv by patching .S files in https://github.com/enjoy-digital/litex/pull/1292. This only fixes the problem for .S files, so the usage of csr instructions in c files was still broken.

Later it was fixed for all source files for Vexriscv in https://github.com/enjoy-digital/litex/pull/1321 by telling the compiler to target the 2.0 ISA, where the csr instructions were still present.

Make the same change to all riscv cpus that specify -march=rv32. This should allow both old and new toolchains to build software.

### DIFF
--- a/litex/soc/cores/cpu/cv32e40p/core.py
+++ b/litex/soc/cores/cpu/cv32e40p/core.py
@@ -22,15 +22,15 @@ CPU_VARIANTS = ["standard", "full"]
 # GCC Flags ----------------------------------------------------------------------------------------
 
 GCC_FLAGS = {
-    #                       /-------- Base ISA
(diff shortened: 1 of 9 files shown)


shenki:
cc @zeldin

enjoy-digital:
Thanks @shenki for updating the different RISC-V CPUs.

PR #1481, opened 2022-10-30
soc/software/demo/Makefile: define WITH_CXX

alanvgreen:
Defines WITH_CXX to the C compiler when including hello.cpp in the demo. This fixes a bug wahere the the menu did not include the hellocpp item even when --with-cpp was passed to demo.py

Signed-off-by: Alan Green <alan.green@gmail.com>


enjoy-digital:
Thanks @alanvgreen!

PR #1482, opened 2022-11-01
misc openc906 fixes and enhancements

Icenowy:
This PR contains some fixes and enhancements for OpenC906.

Hardware side, it fixes AXI characteristics and added an integrated AXI2WB bridge as workaround like CVA6 does.

Software side, cache flushing is fixed and some other performance-related options are enabled.

### DIFF
--- a/litex/soc/cores/cpu/openc906/core.py
+++ b/litex/soc/cores/cpu/openc906/core.py
@@ -11,6 +11,7 @@ from migen import *
 
 from litex import get_data_mod
 from litex.soc.interconnect import axi
+from litex.soc.interconnect import wishbone
 from litex.soc.cores.cpu import CPU, CPU_GCC_TRIPLE_RISCV64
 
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @Icenowy, the issue with the DownConverter could be similar to https://github.com/enjoy-digital/litex/issues/1313. I'll try to have a closer look soon that we can avoid going through Wishbone.

enjoy-digital:
Thanks @Icenowy, I'll use it to investigate/fix #1313.

Icenowy:
> Thanks @Icenowy, I'll use it to investigate/fix #1313.

It might be easy to get a running converter on C906, but making a standard-conformant converter could be a pain because the maximum burst length of C906 is 4 while the maximum length allowed in AXI is 16, which allows converter w/o burst regroup when converting 128-bit to 32-bit.

PR #1484, opened 2022-11-02
soc/software: Support non 8bit i2c memory addresses

cklarhorst:
Add another parameter to i2c_read/write to specify the size (in bytes) of the addresses.

i2c_read/writes with 2-byte addresses were tested with a OV5647 camera.


enjoy-digital:
Thanks @cklarhorst, this looks good and I also had this limitation in the past.

PR #1485, opened 2022-11-02
soc/interconnect/csr: Fix CSR on 64-bit SoC bus width

Icenowy:
Currently the code uses the SoC bus width to calculate the alignment of CSR banks.

However when we get AXI-Lite interconnect support, the CSR bus is not directly converted from SoC bus now, instead an intermediate bus with default parameter (which means 32-bit) is created, CSR bus is converted from it, and finally this bus is attached to the main interconnect with auto converter if needed. In this case the intermediate bus is always of 32 bit bus width, eliminating the need of caring the SoC bus width when handling CSR banks.

Tested on 64-bit SoC bus width now; to make the bus further wider, other codes need to be changed either.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1201,8 +1201,7 @@ class SoC(LiteXModule):
             address_width      = self.csr.address_width,
             alignment          = self.csr.alignment,
             paging             = self.csr.paging,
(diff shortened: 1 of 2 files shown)


enjoy-digital (approved):

Issue #1025, opened 2021-09-04
Migrate from "bus master" and "bus slave" terminology


alanvgreen:
In the computing context, many find the terms "master" and "slave" problematic and there have been many parallel efforts to migrate away from them. Here's (a list](https://en.wikipedia.org/wiki/Master%2Fslave_(technology)#Possible_replacement_conventions)..

In AMBA AXI documentation, "manager" is replacing "master" and "subordinate" is replacing "slave" ([reference](https://developer.arm.com/documentation/102202/latest/)). 

Would you be open to PRs updating the LiteX API to use the terms "manager" and "subordinate"?

enjoy-digital:
Thanks @alanvgreen, I've indeed been thinking about it but haven't been able to spend the time to really think about the best replacements, the way to ensure retro-compatibility methods call/imports. (And the issue that by providing retro-compatibility, part of the problem is not fully addressed since these terms will still be in the code base... but at least not the default API).

So it's a bit more than replacing the terms; for these replacements to have a real meaning we also need to update all shared projects/examples code. So that's something I have a preference to address myself since this is more than just doing the replacement in the code base. I'll think a bit more about it and do my best to plan some time in the future to do it. Until this, I just want to make it clear that these terms have always been used in the project as technical terms (and not being native English speaker they always had a technical resonance in my mind).

With this project and the support I'm providing, I'm also trying to encourage developers to ask questions, provide feedback, contribute in a friendly and respectful manner since I think every-one can contribute and can improve the project while also enjoying working on it. And while doing so, I've been curious to see some open-source projects, companies, reacting promptly to these social events/changes while not necessarily being exemplar on other aspects that are important for in me in LiteX... 

So these topics are difficult ones if you want to do things correctly, and I'd like LiteX to follow my intuitions on this. I of course understand the concerns, will try to address this in a not too far future, but need to plan it and some time for it. I hope the community can understand this.

Regards,

Florent

alanvgreen:
@enjoy-digital Thank you for your thoughtful response. 

Agreed that retro-compatibility is an important consideration. Yes, I imagine that the new and deprecated names would coexist for some time - apologies that I didn't make that clear.

When you have had time to decide a plan, please let me know - I'd be happy to help.

alanvgreen:
@enjoy-digital I've been thinking more about your response, and realized that I didn't acknowledge some of the points you made.

[...]

PR #1486, opened 2022-11-03
litex/soc: update API to avoid unfortunate terms

alanvgreen:
Provides an alternate API to some functions of the SoCBusHandler and bus Arbiter classes to allow users of the API to avoid the terms 'master' and 'slave' in certain situations.

See discussion at #1025


enjoy-digital:
Thanks @alanvgreen!

PR #1487, opened 2022-11-04
cpu/openc906: add ethmac to memory map and misc changes

Icenowy:
The default ethmac section address conflicts with main_ram defined for openC906.

Add a custom position of ethmac to the memory map, which directly follows the internal APB.

Also fix the start address of checked IO region from 0xa0000000 to the full Region 1 in sysmap.h, and add plic and clint sections like other RISC-V CPU cores (although they're internal to the CPU and won't be usable by LiteX).


enjoy-digital:
Thanks @Icenowy!

PR #1488, opened 2022-11-04
Misc changes for a wider SoC

Icenowy:
OpenC906 core has a 128-bit data width bus, which could benefit from a wider SoC bus.

This PR includes two commits, one adding support for ROM initialization of a wider bus, the other attaching LiteDRAM to the SoC bus with a wider data port, allowing to utilize the full SoC bus width.

This increases sequential DRAM read speed of openC906 in LiteX from 110MB/s to 190MB/s when using a 128-bit SoC bus.

### DIFF
--- a/litex/soc/integration/common.py
+++ b/litex/soc/integration/common.py
@@ -40,7 +40,7 @@ def get_mem_regions(filename_or_regions, offset):
     return regions
 
 def get_mem_data(filename_or_regions, data_width=32, endianness="big", mem_size=None, offset=0):
-    assert data_width in [32, 64]
+    assert data_width % 32 == 0
     assert endianness in ["big", "little"]
 
     # Return empty list if no filename or regions.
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks a lot @Icenowy for making `get_mem_data` more generic! This was initially hardcoded to 32-bit, I quickly added 64-bit support to test NaxRiscv-64-bit and was planning to revisit it in the future; thanks for the effort.

litex/soc/integration/common.py
```diff
@@ -80,11 +80,11 @@ def get_mem_data(filename_or_regions, data_width=32, endianness="big", mem_size=
...
-                if data_width == 32:
-                    data[(base - offset)//bytes_per_data + i] = struct.unpack(unpack_order, w)[0]
-                if data_width == 64:
-                    data[(base - offset)//bytes_per_data + i]  = (struct.unpack(unpack_order, w[0:4])[0] <<  0)
-                    data[(base - offset)//bytes_per_data + i] |= (struct.unpack(unpack_order, w[4:8])[0] << 32)
+                data[(base - offset)//bytes_per_data + i] = 0
+                for filled_data_width in range(0, data_width, 32):
+                    cur_byte = filled_data_width//8
+                    data[(base - offset)//bytes_per_data + i] |= (struct.unpack(unpack_order, w[cur_byte:cur_byte+4])[0] << filled_data_width)
+                    filled_data_width += 32
```
Rongronggg9, line 87:
```suggestion
```

Hmm, this should be a line of dead code as the for-loop will always set it to be the right value.

enjoy-digital: Indeed, thanks @Rongronggg9, this is fixed with https://github.com/enjoy-digital/litex/commit/8f459a27dda79789094ce400a885ecf2a902a768.

PR #1489, opened 2022-11-04
Do not build software as PIE

shenki:
Some builds may default to creating position independent executables with a .dynamic section containing symbols that must be relocated at runtime. This section appears after .data section and is included in the output .bin, causing it to be included in the build time CRC calculation. The runtime CRC calculation stops after .data, so there was always a mismatch.

The litex bios has no support for relocating the dynamic symbols, so disable pie on all platforms. This will improve compatibility with distro toolchains.

Tested with:

 marocchino / or1k-elf-gcc (GCC 12.1.0-2 Debian 1.0.3) 12.1.0

 vexriscv / riscv64-linux-gnu-gcc (Debian 12.2.0-3) 12.2.0

 microwatt / powerpc64le-linux-gnu-gcc (Debian 12.2.0-3) 12.2.0


enjoy-digital:
Thanks @shenki! (and for verifying it with the different ISAs).

PR #1490, opened 2022-11-04
tools/litex_sim: hack to allow memory base other than 0x40000000

Icenowy:
The LiteX OpenC906 core currently uses 0x0 as the base of main memory.

Hack to allow this when preloading a binary in litex_sim by adding a command line argument.

### DIFF
--- a/litex/tools/litex_sim.py
+++ b/litex/tools/litex_sim.py
@@ -368,6 +368,7 @@ def sim_args(parser):
     verilator_build_args(parser)
     parser.add_argument("--rom-init",             default=None,            help="ROM init file (.bin or .json).")
     parser.add_argument("--ram-init",             default=None,            help="RAM init file (.bin or .json).")
+    parser.add_argument("--main-ram-init-base",   default="0x40000000",    help="(SD)RAM base assumed in init file.")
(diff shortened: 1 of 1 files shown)


Icenowy:
Sorry for my previous buggy commit -- I am not familiar with Python.

enjoy-digital:
Thanks @Icenowy, please just give me a few days to have a closer look (Maybe we should allow name it `--main-ram-base` to configure the `main_ram` address and initialization file would be filled there)

enjoy-digital:
@Icenowy: I'm merging since I also want to do other changes to litex_sim and would like to avoid conflict. I'll also have a closer look at this while doing my other changes and will post an update here.

Icenowy:
Well I hope I will see a better solution ;-)

enjoy-digital:
@Icenowy: With https://github.com/enjoy-digital/litex/commit/4a740651f0a76f25ee18b6c8405161fd40181511, we are now creating a temporary soc to avoid some workarounds in the configuration of the real one. The offset is now just directly: `offset  = conf_soc.mem_map["main_ram"]`, which should I think solve the issue you had.

PR #1494, opened 2022-11-07
litex/build/parser: check if self._target_group is not None before calling add_argument

trabucayre:
When the platform is None `add_target_group` is never called and consequently `_target_group` is None.
If a call is done to `add_target_argument` an error appear.
This may appear with `simple.py` where the plaform is known at `parse_args` time so after adding **toolchain** argument.

This PR updates `add_target_argument` to first test `_target_group`, create the group if required before using `_target_group.add_argument`


enjoy-digital:
Thanks @trabucayre!

PR #1495, opened 2022-11-08
cpus/openc906: add debug variant like vexriscv

Icenowy:
This PR adds a debug variant, which behaves like the VexRiscv debug variant, that maps the DM onto the system bus.

The DM itself is RV Debug Spec 0.13-compliant.


Icenowy:
See https://github.com/Icenowy/linux-on-litex-openc906/blob/master/litex2jtag.py for a wrapper of RV DMI.

This wrapper talks with OpenOCD with jtag_dpi protocol.

enjoy-digital:
Thanks @Icenowy! That's merged. Also nice for `litex2jtag`; this could be useful for other CPU also implementing the RV Debug Spec (at least VexRiscv-SMP/NaxRiscv and VexRiscv that should also switch to it).

PR #1500, opened 2022-11-11
core/naxriscv: Don't use os.system to execute sbt

cklarhorst:
The use of os.system together with cd is a problem because it changes the CWD for the whole python process. This breaks for example --csr-csv.
--> The csr.csv will be written to the wrong directory


enjoy-digital:
Thanks @cklarhorst!

PR #1501, opened 2022-11-12
AXICrossbar: absorb data width setting when building crossbar

bunnie:
This patch allows the AXICrossbar to absorb non-default datawidths when the crossbar is built.

e.g. in the case of a 64-bit AXICrossbar, without this patch the crossbar was only connecting the bottom 32 bits.

### DIFF
--- a/litex/soc/interconnect/axi/axi_full.py
+++ b/litex/soc/interconnect/axi/axi_full.py
@@ -574,7 +574,7 @@ class AXICrossbar(Module):
     """
     def __init__(self, masters, slaves, register=False, timeout_cycles=1e6):
         matches, busses = zip(*slaves)
-        access_m_s = [[AXIInterface() for j in slaves] for i in masters]  # a[master][slave]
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @bunnie, this is merged. I'll also apply it to AXI-Lite/Wishbone, will also do the same for address_width and will also probably add a check to be sure that all Masters/Slaves use the same parameters.

enjoy-digital:
Done with https://github.com/enjoy-digital/litex/commit/a3cc741d2653feac20971790baa81678588259be and https://github.com/enjoy-digital/litex/commit/82127043c3c6bd7b23791a03525b7c950be1f0a4.

bunnie:
thanks!

enjoy-digital:
In fact I had to disable `address_width` check with https://github.com/enjoy-digital/litex/commit/05b0c5960780466af4bdca08af73f844832f1db4 for now since it introduce regressions on some designs. More tests will be required before enabling it. But `data_width` propagation and checks are still there.

PR #1505, opened 2022-11-14
Fix Wishbone arbiter

michalsieron:
Right now, when multiple masters want to access the bus, access is granted to one of them, and is not revoked until selected master has finished all of its transactions (cyc goes low).

This state causes master starvation if access is granted to high bandwidth master, like cpu in busy loop.

This commit makes it so access to bus is revoked when pending transaction is finished (ack and cyc are high) or when selected master is idle.

### DIFF
--- a/litex/soc/interconnect/wishbone.py
+++ b/litex/soc/interconnect/wishbone.py
@@ -166,7 +166,8 @@ class Arbiter(Module):
         if controllers is not None:
             masters = controllers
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @michalsieron, we indeed need to improve arbitration and allow more complex schemes (ex priority on ports), this is a good first improvement, thanks!

PR #1506, opened 2022-11-15
RFC: Fix sensitivity dependencies in FSM for AXI-lite/AXI

bunnie:
The re-use of a left hand side (LHS) assignment in a right hand
side (RHS) evaluation inside an `always (@*)` block with non-blocking
operators leads to an infinite loop in some verilog simulators.

In particular, a form such as
```python
fsm.act("A",
   sig.eq(new_value),
   If(sig,
      # ...
   )
)
```
Causes an infinite loop when `new_value` is not 0. This is because
this gets translated in verilog to
```verilog
always (@*) begin
   sig <= 1'd0;
   ...
   sig <= new_value;
   if (sig) begin
      ...
   end
end
```
So what happens is that `sig` gets 0, then, it gets assigned
`new_value`; it is then used by the `if` to decide what to do.

When `new_value` is 0, no change is detected, and there is no
inifnite loop.

However, if `new_value` is 1, the simulator detects the 0->1
transition, and schedules the `always(@*)` for re-evaluation,
because the sensitivity list, which includes `sig` had a transition.

In this case, the simulator (at least some of them) ends up
in an infinite loop, and the simulation is unable to proceed.

This PR is more of an RFC for how to fix the underlying problem.
The commits here as-writ allow the simulation to run and it seems
correct, however, please note that I was unable to fix the
AXITimeout/AXILiteTimeout blocks, for now I am just not using them.

### DIFF
--- a/litex/soc/interconnect/axi/axi_full.py
+++ b/litex/soc/interconnect/axi/axi_full.py
@@ -343,7 +343,7 @@ class AXITimeout(Module):
                 timer.wait.eq(wait_cond),
                 # done is updated in `sync`, so we must make sure that `ready` has not been issued
                 # by slave during that single cycle, by checking `timer.wait`.
-                If(timer.done & timer.wait,
+                If(timer.done & wait_cond, # timer.wait.eq(wait_cond),
                     error.eq(1),
                     NextState("RESPOND")
                 )
@@ -360,7 +360,7 @@ class AXITimeout(Module):
                 master.w.ready.eq(master.w.valid),
                 master.b.valid.eq(~master.aw.valid & ~master.w.valid),
                 master.b.resp.eq(RESP_SLVERR),
-                If(master.b.valid & master.b.ready,
+                If((~master.aw.valid & ~master.w.valid) & master.b.ready, # timer.wait.eq(wait_cond),
                     NextState("WAIT")
                 )
             ])
@@ -375,7 +375,7 @@ class AXITimeout(Module):
                 master.r.last.eq(1),
                 master.r.resp.eq(RESP_SLVERR),
                 master.r.data.eq(2**len(master.r.data) - 1),
-                If(master.r.valid & master.r.ready,
+                If(~master.ar.valid & master.r.ready, # master.ar.ready.eq(master.ar.valid),
                     NextState("WAIT")
                 )
             ])
--- a/litex/soc/interconnect/axi/axi_full_to_axi_lite.py
+++ b/litex/soc/interconnect/axi/axi_full_to_axi_lite.py
@@ -81,7 +81,8 @@ class AXI2AXILite(Module):
             axi.r.data.eq(axi_lite.r.data),
             axi_lite.r.ready.eq(axi.r.ready),
             # Exit
-            If(axi.r.valid & axi.r.last & axi.r.ready,
+            # If(axi.r.valid & axi.r.last & axi.r.ready,   # Original semantic intent
+            If(axi_lite.r.valid & _cmd_done & axi.r.ready, # Revised so assignments not affecting always(@*) sensitivity list
                 ax_beat.ready.eq(1),
                 NextState("IDLE")
             )
@@ -105,7 +106,8 @@ class AXI2AXILite(Module):
             axi_lite.w.strb.eq(axi.w.strb),
             axi.w.ready.eq(axi_lite.w.ready),
             # Exit
-            If(axi.w.valid & axi.w.last & axi.w.ready,
+            # If(axi.w.valid & axi.w.last & axi.w.ready,    # Original semantic intent
+            If(axi.w.valid & axi.w.last & axi_lite.w.ready, # Revised so assignments not affecting always(@*) sensitivity list
                 NextState("WRITE-RESP")
             )
         )
--- a/litex/soc/interconnect/axi/axi_lite.py
+++ b/litex/soc/interconnect/axi/axi_lite.py
@@ -135,6 +135,7 @@ def axi_lite_to_simple(axi_lite, port_adr, port_dat_r, port_dat_w=None, port_we=
                 comb.append(port_we[i].eq(axi_lite.w.valid & axi_lite.w.ready & axi_lite.w.strb[i]))
         else:
             comb.append(port_we.eq(axi_lite.w.valid & axi_lite.w.ready & (axi_lite.w.strb != 0)))
+<<<<<<< HEAD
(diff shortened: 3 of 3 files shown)


enjoy-digital:
Thanks @bunnie; I'll have a closer look in the next few days.

In the meantime, can you try using `regular_comb=False` when generating the verilog:
https://github.com/enjoy-digital/litex/blob/master/litex/gen/fhdl/verilog.py#L531
https://github.com/enjoy-digital/litedram/blob/master/litedram/gen.py#L861
And try to simulate with this?

bunnie:
> 

I just tried it with `regular_comb=False` and it seems to improve some of the problematic code, but still some other issues remain and I'm not able to run the simulation all the way through.

bunnie:
Ah, but sugnificantly, `regular_comb=False` *does* allow me to use the AXITimeout blocks again!

enjoy-digital:
Thanks for the test/info, which simulator are you using BTW?

bunnie:
I'm using `xsim`, the simulator in Vivado.

enjoy-digital:
OK thanks, could you explain how to reproduce the issue? Just creating an AXI or AXI-Lite based LiteX SoC and try to simulate it with `xsim`?

bunnie:
You also have to do a read or write access to the AXI-lite bus to trigger the condition, so some firmware is needed.

I just tried to cut back my design to something minimal but because our BIOS is in Rust, I decided maybe it's not going to be worth it to adapt my whole system over as I'm guessing you don't have the Rust toolchain installed etc.

However, I suspect if you can just configure a SoC with "axi-lite" as the bus type and then simulate using `xsim` you should see an error like below:

image

If you want to figure out what lines are causing the trouble, you can run the TCL command `ptrace on`, and then run the simulator. If it's triggering very early this is pretty quick but if you have a lot of code running before you touch an AXI peripheral it's faster to run to a time just before the failure and then flip `ptrace on`.

If you can't reproduce still, I will work a bit more to try and configure a demo, but I strongly suspect you probably have something fairly easy to run. 

Here is the script I use to launch xsim, btw:

```python

        # copy over the top test bench and common code
        os.system("{} top_tb.v run".format(cpname) + os.path.sep + "top_tb.v") # "cp top_tb.v run/top_tb.v"
        os.system("{} VexRiscv.v run".format(cpname) + os.path.sep + "VexRiscv.v") # "cp top_tb.v run/top_tb.v"
        os.system("{} ".format(cpname) + os.path.normpath("sim_support/common.v") + " run" + os.path.sep) # "cp sim_support/common.v run/"

        # initialize with a default waveform that contains the most basic execution tracing
        if os.path.isfile('run/top_tb_sim.wcfg') != True:
            if os.path.isfile('top_tb_sim.wcfg'):
                os.system('{} top_tb_sim.wcfg run'.format(cpname) + os.path.sep) # 'cp top_tb_sim.wcfg run/'
            else:
                os.system('{} .'.format(cpname) + os.path.sep+'sim_support'+os.path.sep+'top_tb_sim.wcfg run' + os.path.sep) # 'cp ./sim_support/top_tb_sim.wcfg run/'

        # load up simulator dependencies
        os.system("cd run && {} ".format(cpname)+os.path.normpath("../build/gateware") + os.path.sep + "*.init .")
        os.system("cd run && {} ".format(cpname)+os.path.normpath("../build/gateware") + os.path.sep + "*.v .")
        os.system("cd run && xvlog "+os.path.normpath("../sim_support/glbl.v"))
        os.system("cd run && xvlog test_axi_soc.v -sv")
        os.system("cd run && xvlog VexRiscv.v -sv")
        os.system("cd run && xvlog top_tb.v -sv ")
        vex_dir = os.path.dirname(VEX_CPU_PATH)
        vex_dir = vex_dir.replace("/", os.path.sep)
        # copy any relevant .bin files into the run directory as well
        os.system("{} {} ".format(cpname, vex_dir + os.path.sep + "*.bin") + " run" + os.path.sep) # "{} {} run/".format(cpname, vex_dir + "/*.bin")
        os.system("cd run && xvlog {}".format(".." + os.path.sep + vex_verilog_path)) # "cd run && xvlog {}".format("../" + vex_verilog_path)

        # run user dependencies
        for cmd in os_cmds:
            os.system(cmd)

        os.system(
            "cd run && xelab -debug typical top_tb glbl -s top_tb_sim -L unisims_ver -L unimacro_ver -L SIMPRIM_VER -L secureip -L $xsimdir/xil_defaultlib -timescale 1ns/1ps")
        os.system("cd run && xsim top_tb_sim -gui")

```

Basically the steps to invoke are:

1. create a directory `run`
2. copy all the .v and .init artifacts into there
3. call `xvlog` to compile every verilog dependency
4. call `xelab` with the corresponding library arguments (second to last line of the script) to invoke

`glbl.v` can be found here: https://github.com/betrusted-io/gateware/blob/main/sim_support/glbl.v
`common.v` is just some boilerplate code: https://github.com/betrusted-io/gateware/blob/main/sim_support/common.v

Probably you know all this already, but, just trying to make it easier for you in case you rarely use `xsim`...

bunnie:
I've added a couple more cases where the simulator chokes due to a combinational loop.

All of these changes *should* be 1:1 equivalent to the original code -- the main pattern is I'm pulling purely combinational stuff out of the FSM blocks and pushing them into `comb` statements, so that the updates resulting from the combinational logic do not affect the FSM's sensitivity list and trigger an infinite loop in the simulator.

bunnie:
(sorry for the slightly messy commit record, because I also have other PRs in progress my branch is starting to diverge quite a bit with Litex mainline...I'm trying to keep the threads separate though.)

enjoy-digital:
hanks again for the investigation and detailed explanation.

I think this is now obsolete in its original form. The underlying issue was the Verilog generated for combinatorial logic: LiteX could emit non-blocking assignments inside always @(*) blocks, which could trigger xsim zero-delay/infinite-loop behavior when a signal assigned in the block was also used by the block logic.

This has since been fixed more generally with enjoy-digital/litex@5b641f51766da5a238897d48f034a8d93cacc489. Current LiteX now emits blocking assignments for combinatorial procedural logic, so the original simulator sensitivity issue should be addressed globally rather than only in the AXI/AXI-Lite blocks touched here.

I also checked the current AXI/AXI-Lite timeout and AXI-to-AXI-Lite tests, and the relevant tests pass on current master.

So I think we can close this PR now. If there is still an xsim issue with current master, it would be better to open a fresh issue/PR with a minimal reproducer against the current code.

PR #1507, opened 2022-11-15
RFC: locate reset vector in custom memory bus

bunnie:
This PR allows the reset vector to be located within a custom memory bus that is not enumerated by the SoC bus handler.

The use case is when generating a core that is:

1) to be integrated into a third party SoC, where all the memory bus wiring is handled by a different framework
2) none of the local/automatic busses are desired to be inferred with the framework.

An example of this usage would be here:

https://github.com/buncram/cram-soc/blob/cram/cram_core.py

I imagine I'm not quite doing this the way that @enjoy-digital would prefer it to be done, but at the high level, I wanted to be able to configure LiteX to not check that the CPU reset region is in a defined memory region in the case that the memory region is to be defined elsewhere for third-party integrations.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1241,7 +1241,7 @@ class SoC(LiteXModule, SoCCoreCompat):
                     cpu_reset_address_valid = True
                     if name == "rom":
                         self.cpu.use_rom = True
-            if not cpu_reset_address_valid:
+            if not cpu_reset_address_valid and not self.cpu_custom_memory:
                 self.logger.error("CPU needs {} to be in a {} Region.".format(
                     colorer("reset address 0x{:08x}".format(self.cpu.reset_address)),
                     colorer("defined", color="red")))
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @bunnie; I understand the need and will have a closer look at this very soon.

cklarhorst:
I also like this PR. Maybe `cpu_custom_memory` should be renamed to something like `reset_region_check` (with inverted logic). Then it has the same format as https://github.com/enjoy-digital/litex/blob/703bd16a9684c13417e90a117b94fe11c3924287/litex/soc/integration/soc.py#L162

enjoy-digital:
@bunnie: With https://github.com/enjoy-digital/litex/commit/4baeeed946c1a063661ede45241f329d916bd5d1 you can now disable the CPU reset address check by calling `self.cpu.disable_reset_address_check()` in your SoC.

PS: Sorry I initially wanted to do a PR for that to discuss/suggest but pushed it directly. We could still change if you  or @cklarhorst don't like this solution.

bunnie:
Works for me! I'll close this out.

PR #1510, opened 2022-11-16
FIX: Handle un-aligned edge case on AXI upconverter

bunnie:
The AXI upconverter currently always starts the demux aligned to the width of the destination bus. So, for a 32->64 bit up converter, it can correctly up-convert a single word-write with an alignment along 8-byte boundaries. However, "odd" alignments along 4-byte boundaries are not correctly aligned.

This proof-of-concept patch shows that making the demux starting address-aware fixes the issue.

However, the way it's done is pretty gross, I'm pretty sure there is a better way to patch this in.

I'll add more details once the PR is in github.


bunnie:
"Current" behavior showing incorrect byte stuffing on misaligned start address:
image

Here, you can see the value 0x0000_0002 being put into the LSB. On AXI, this would indicate to write the byte not into the intended address of 6100_0008, it would instead put it into 6100_0000.

"Working" demo:
image

This shows the value 0x0000_0002 being put into the MSB. I just noticed that this waveform has the `wrstb` as `0xff`, which is wrong, that should also be changing over but for this particular test case it "worked" because it's just an up-counting write to memory. But, it at least confirms the notion that the stride converter for AXI needs to consider the possibility of a write starting at an address that is not aligned to the destination bus width.

bunnie:
Alright, added a commit that fixes the write strobe issue. It now properly only gates the lane of words being written:

image

More testing is going to be needed to check how this performs over u8/u16 sized data, but I wanted to at least get the POC up for review in case there is something obvious I'm missing here.

enjoy-digital:
Thanks @bunnie for looking at this. The current implementation is indeed was indeed just satisfying NaxRiscv requirements (and assumptions) and needs to be fixed/improved for all use-cases.

I did some improvements to the DownConverter and think that for the UpConverter, we could increase burst's len by 1 in case of un-aligned accesses then use `strb`on the datapath to to partial accesses on downstream. I think we first need to have a good unit-test for this module and then do the changes. I'll try to have a look at this next week.

bunnie:
FYI, I've been having a lot of problems with the Litex AXI blocks, so I have moved my design to use the blocks in [`verilog-axi`](https://github.com/enjoy-digital/litex_verilog_axi_test), using adapted wrappers.

So far the verilog blocks seem to be working better, modulo some tiny tweaks. So, if you want to close this PR without accepting it, that's fine; it is no longer in my critical path. However, the bug at the root cause of this PR is still a real bug, so anyone else using the converter in the mode I was trying to use it would have the same issue.

enjoy-digital:
@bunnie: LiteX AXI support is indeed probably still not mature enough to be used on system like the one you are working on. (The subset present has been implemented to fill our need). Switching to [`verilog-axi`](https://github.com/enjoy-digital/litex_verilog_axi_test) is indeed probably a good move for now (and pretty sure I recommended it initially :)) since I was not able to be responsive enough.

Thanks a lot for your different PRs/feedback. Let's keep them open since the points raised needs to be addressed and I'll have a look at this progressively.

enjoy-digital:
@bunnie:  Thanks again for the original report and PoC.

I just updated this PR against current `master`. The native AXIUpConverter has changed quite a bit since this PR was opened and now already handles this in a cleaner way: the write-address lane is carried through the AW path and used directly when packing narrow W beats into the wider downstream AXI beat.

So I resolved the conflicts by keeping the current `master` implementation rather than keeping the old stream-level PoC changes.

I also added direct regression coverage for the original 32-bit to 64-bit odd-lane write case with:

enjoy-digital/litex@9146ab97dbf727879d09a87fa268c92d2803e07a

The PR is now mergeable again and the remaining diff is only this focused test update. This should allow us to keep coverage for the reported issue without reintroducing the older workaround.

PR #1513, opened 2022-11-17
FIX: fix wishbone arbiter

bunnie:
Commit 07184d37df9921fa2bf729f131d276402bbb3605 breaks designs in my CI testbench. The intent of the commit was to change the wishbone arbiter policy for "everyone" to solve a corner case, but in fact, it breaks other corner cases.

This fix does the following:

- makes the legacy arbiter choice the default, so we don't break designs that currently work just fine

- expose an `arbiter_policy` flag available on `soc.bus` which can be set to override the default policy for designs that require a different arbiter policy to work around a particular use case.

Even if this PR isn't what one would go with syntatically, I would strongly suggest either making the change optional or figuring out why changing the wishbone arbiter policy breaks existing designs. Really chasing down all the implications of a bus policy change like this is a big project that could take a lot of time, especially on SoCs with a single CPU core that is expected to "own" the bus.

@mtdudek it's a big change to switch the default shared fabric arbiter policy for an entire design ecosystem!

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -184,6 +184,7 @@ class SoCBusHandler(LiteXModule):
     supported_address_width = [32, 64]
     supported_addressing    = ["word", "byte"]
     supported_interconnect  = ["shared", "crossbar"]
+    supported_arbiter       = ["default", "transaction"]
 
     # Creation -------------------------------------------------------------------------------------
     def __init__(self, name="SoCBusHandler",
@@ -194,6 +195,7 @@ class SoCBusHandler(LiteXModule):
         timeout          = int(1e6),
         bursting         = False,
         interconnect     = "shared", interconnect_register=True,
+        arbiter          = "default",
         reserved_regions = None,
     ):
         self.logger = logging.getLogger(name)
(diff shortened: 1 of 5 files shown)


enjoy-digital:
@bunnie: Thanks for looking at this and reporting the issue. @mtdudek: It would indeed be interesting to understand the issue in @bunnie's testbench and if possible fix it without having to propagate the Arbiter policy.
Once understood, adding a testbench in LiteX covering this would also be interesting.

enjoy-digital:
I reverted #1505 for now with https://github.com/enjoy-digital/litex/commit/f66852b97571ce33a153383ef7afb6a328f2580d.

bunnie:
> I reverted #1505 for now with [f66852b](https://github.com/enjoy-digital/litex/commit/f66852b97571ce33a153383ef7afb6a328f2580d).

Thanks, my CI dashboard is green now, which is helpful because I'm trying to refactor out the deprecated APIs so we have a smoother transition to your next release.

mithro:
@bunnie - Do any of your tests make sense to incorporate in this upstream repository?

bunnie:
Unfortunately our tests rely on a custom BIOS written in Rust. I think the step before upstreaming the tests is upstreaming our Rust build. It's on the list of things to do but it's a very long list of things to do.

In the meantime our CI builder is publicly accessible.

enjoy-digital:
@bunnie: Thanks for the PR and sorry for the delay :)

I reworked the PR on top of current master with a smaller compatibility-safe implementation: a5531188b24b003772eccae7dabb06460f363326.

The default bus arbitration behavior is unchanged with:

  ```sh
  --bus-arbiter=default
  ```

  For the starvation case discussed here, transaction-level arbitration can now be enabled explicitly with:
  ```sh
  --bus-arbiter=transaction
  ```
For now this maps to the Wishbone arbiter and lets arbitration advance after ack, err, or when the selected master withdraws cyc. AXI/AXI-Lite are not changed; non-default arbiter modes are rejected there until implemented.

I also added regression coverage for:

  - the existing Wishbone cycle-held behavior,
  - the new transaction-level behavior,
  - SoCBus validation,
  - command-line parser forwarding.

PR #1519, opened 2022-11-21
github: Use distibution compilers for riscv and or1k

shenki:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,12 +36,8 @@ jobs:
       # Install GCC Toolchains
       - name: Install GCC Toolchains
         run: |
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Just for info, since I also want have GCC install utilities from `litex_setup.py` tested in CI, I switch GCC install in `litex_setup.py ` to distro installs (https://github.com/enjoy-digital/litex/commit/dd91c55c363d62a197552f4e73a6e011d8edb228) and switched CI to use it (https://github.com/enjoy-digital/litex/commit/2c9ddc20be60caf039f41894a6826268679ab5ab). So in the end, it's similar to what you were doing, just with `litex_setup.py` in between and tested in CI.

PR #1520, opened 2022-11-23
Zynq7000: Implement function to add axi gp slave

mkuhn99:
I implemented the add_axi_gp_slave function of the zynq7000 to use it with this [puil request](https://github.com/litex-hub/litex-boards/pull/453). The goal is to access the DDR and the IOPs of the processing system from a soft CPU on the PL.

The implementation is quite similar to the  existing `add_axi_gp_master` and `add_axi_hp_slave` functions, but I added the clock domain as an argument, because one might not want to use the ps7 clock in every case.


trabucayre:
LGTM.

PR #1522, opened 2022-11-28
interconnect/axi: map from one address space to another

mkuhn99:
When trying to connect to the PS axi gp slave to the main Bus of the LiteX SoC I came across the problem that the address space of the zybo PS and the address space/memory map of the Litex SoC are conflicting.
To solve this problem I added the `connect_mapped` function to the AXIInterface, which applies a function to the address signals before connecting the two interfaces together.

Such a function would, for example, add or subtract an offset.

### DIFF
--- a/litex/soc/interconnect/axi/axi_full.py
+++ b/litex/soc/interconnect/axi/axi_full.py
@@ -131,6 +131,13 @@ class AXIInterface:
     def connect(self, slave, **kwargs):
         return connect_axi(self, slave, **kwargs)
 
+    def connect_mapped(self, slave, map_fct):
+        comb = []
+        comb += self.connect(slave, omit={"addr"})
+        comb += [slave.ar.addr.eq(map_fct(self.ar.addr))]
+        comb += [slave.aw.addr.eq(map_fct(self.aw.addr))]
+        return comb
+
     def layout_flat(self):
         return list(axi_layout_flat(self))
 


mkuhn99:
[Here](https://github.com/litex-hub/litex-boards/pull/456) is a case where one would/could use this function.

enjoy-digital:
Thanks @mkuhn99,  this is indeed interesting and I understand the need. Please just give me some time to think about it.

enjoy-digital:
This has now been handled with a structured remapping path:

  - enjoy-digital/litex@39390177c adds `src_regions` / `dst_regions` support to AXI and AXI-Lite remappers.
  - enjoy-digital/litex@2a5ce7859 adds AXI/AXI-Lite regression coverage for region remapping.
  - enjoy-digital/litex@a0dc61851 adds the higher-level `SoCBusHandler.add_slave(..., dst_region=...)` API.

This allows a LiteX-visible decode region to differ from the downstream/external bus address window, for example:

  ```python
  self.bus.add_slave(
      name       = "ps_io",
      slave      = axi_ps,
      region     = SoCRegion(origin=0x6000_0000, size=0x0030_0000),
      dst_region = SoCRegion(origin=0xe000_0000, size=0x0030_0000),
  )
  ```

So the Zynq/PS-style remapping use case can now be expressed directly at the SoC bus level, without adding an arbitrary address callback to the low-level AXI connect() helper. The existing strip_origin behavior is kept unchanged as the map-to-zero special case.

PR #1524, opened 2022-11-29
Add octo variants for Rocket

Icenowy:
Add some Octo-width (512bit) memory bus variants for Rocket, which is useful for Xilinx 7-series FPGAs with DDR3 DIMMs (1:4, DDR, 64-bit, which results in 4*2*64=512 bits)

See https://github.com/litex-hub/pythondata-cpu-rocket/pull/5 for corresponding pythondata-cpu-rocket change.

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -55,6 +55,9 @@ CPU_VARIANTS = {
     "full4d":   "freechips.rocketchip.system.LitexFull4DConfig",
     "fullq":    "freechips.rocketchip.system.LitexFullQConfig",
     "full4q":   "freechips.rocketchip.system.LitexFull4QConfig",
+    "fullo":    "freechips.rocketchip.system.LitexFullOConfig",
+    "full4o":   "freechips.rocketchip.system.LitexFull4OConfig",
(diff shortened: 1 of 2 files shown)


gsomlo:
I have a VC707 I plan to dust off and use to test these (and
corresponding pythondata-cpu-rocket) changes over the next couple
of weeks. Thanks in advance for your patience!

mithro:
FWIW - I have some potential FPGA boards I could provide remote access to
if it would help people.

Icenowy:
Well I think with VC707 there could be full8o ;-)

gsomlo:
I added 1x, 4x, and 8x core "octo" variants with #937428b1 -- Thanks!

Icenowy:
在 2022-12-04星期日的 05:00 -0800，gsomlo写道：
> I added 1x, 4x, and 8x core "octo" variants with #937428b1 -- Thanks!

Maybe we should keep 6x just because 8x cannot fit a XC7K325T.

>

gsomlo:
At this point, it feels like maybe a better way forward, in the big scheme of things, would be to get rid of pre-generated variants and come up with a way to generate the desired Rocket configuration on the fly, as part of the LiteX SoC build.

We'd be depending on Chisel/Scala, which is a major (IMHO) hurdle, but I've started to think about ways I could make it work...

PR #1526, opened 2022-12-02
integration/soc: Warn on MemBus <-> LiteDRAM AXI width conversion

gsomlo:
CPUs with a dedicated memory port (bus) can be connected directly to the LiteDRAM port. Some models (e.g., Rocket) have otherwise equivalent variants specifically pre-generated to fit the various "standard" LiteDRAM port widths (so far, 64, 128, and 256 bits).

This patch introduces a warning when the CPU variant's dedicated memory bus width doesn't exactly match the width of LiteDRAM, requiring explicit Up- or Down-conversion.

The goal is to inform the user and give them the opportunity to pick a more suitable CPU variant (of a matching MemBus width), if available.


enjoy-digital:
Thanks @gsomlo!

PR #1527, opened 2022-12-04
software/demo: Add .got and .toc to .rodata in linker script

stone3311:
When using printf in the demo application, gcc emits .got and .toc sections which are placed in SRAM by default. This breaks printf and other functions relying on the GOT if the demo is loaded into flash or ROM. This pull request matches the demo's .rodata section to that of the BIOS.


enjoy-digital:
Thanks @stone3311, this is merged.

PR #1528, opened 2022-12-04
Pretty print Verilog constants by intelligently chosing decimal or hex

jevinskie:
I'm dusting off a bunch of my old Quality of Life patches now that I'm actively using LiteX again, in hopes to upstream them or start some discussions about the pain points.

The pain point for this particular patch is that the Verilog writer always uses decimal base in `_print_constant()` even though some constants beg for hex representation. :)

The approach is to calculate the entropy of both decimal and hex strings and pick the one with the most repetitive characters (think `d'1000000` or `h'ffffff`), and if that is a tie, the one with the least entropy. The result is cached since we print the same numbers very often.

New pain points: case statement indices change base in the middle of the sequence. I will force hex or decimal for that particular print. Do you have a preference?

### DIFF
--- a/litex/gen/fhdl/verilog.py
+++ b/litex/gen/fhdl/verilog.py
@@ -172,7 +172,7 @@ def _decimal_entropy(buf):
     return ent, nsyms
 
 def _hex_entropy(buf):
-    """Get the entropy of a hexadecimal byte-string [0-9a-z] (no 0c prefix)"""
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @jevinskie! I'll think and study this a bit more. One first possibility could also be to do this decision based on the length of the constant/signals: If length is a power or 2 and >= 4 -> force to hex. If not a power of 2 or < 4, print in decimal or binary.

PR #1529, opened 2022-12-05
build/parser: overrides set_defaults and applying default values just before args_parse()

trabucayre:
With the new `LiteXArgumentParser` some of arguments are sent to`argparse.ArgumentParser` near `parse_args()` call.
When a targets uses set_defaults, these defaults values are lost. For example, `digilent_arty_z7` set `zynq7000` as the default *CPU* but when one didn't provides `--cpu-type=zynq7000` the default `vexriscv` is used instead.

This PR overrides `default_values` method, store locally dict containing defaults args/values. These are applied after parser completely filled but before `parse_args` call resulting to the correct behaviour.


enjoy-digital:
Thanks @trabucayre!

PR #1531, opened 2022-12-06
fixes #1530: axi-lite CSR .we not obeying single-cycle interface spec

### LINKED ISSUE #1530: axi-lite CSR `.we` not obeying single-cycle interface spec
According to:

https://github.com/enjoy-digital/litex/blob/937428b1fc3953bdc1726f292f4cb9bb163bf5cd/litex/soc/interconnect/csr.py#L98-L108

The `.re` and `.we` signals should pulse high for exactly one cycle. However, in practice, I am finding that on an `axi-lite` system, the `.we` signal stays high for longer than one cycle. 

I am using the following simple test bench to create CSRs that exercise the corner cases:

```python

            # test that caching is OFF for the I/O regions
            self.sim_coherence_w = CSRStorage(32, name= "wdata", description="Write values here to check cache coherence issues")
            self.sim_coherence_r = CSRStatus(32, name="rdata", description="Data readback derived from coherence_w")
            self.sim_coherence_inc = CSRStatus(32, name="rinc", description="Every time this is read, the base value is incremented by 3", reset=0)

            self.sync += [
                If(self.sim_coherence_inc.we,
                    self.sim_coherence_inc.status.eq(self.sim_coherence_inc.status + 3)
                ).Else(
                    self.sim_coherence_inc.status.eq(self.sim_coherence_inc.status)
                )
            ]
            self.comb += [
                self.sim_coherence_r.status.eq(self.sim_coherence_w.storage + 5)
            ]
```

This is connected to a system that uses an axi-lite bus via a crossbar framework that uses axi, so the overall connectivity is CPU (axi4) -> crossbar (axi4) -> AXI2Axilite -> Axilite2Simple -> CSR banks.

Basically, the test bench will "add 3" to `rinc` every time it is read, so successive reads should generate a series of numbers like 0, 3, 6, 9, ...

This is what I'm seeing in practice:

image

The `.we` signal stays selected for much more than one cycle, so the register increments more than it should.

The `.we` signal seems this is derived from this code:

https://github.com/enjoy-digital/litex/blob/937428b1fc3953bdc1726f292f4cb9bb163bf5cd/litex/soc/interconnect/csr_bus.py#L183-L198

The `sel` comes from just checking that the address matches, and is then applied with the `If(sel, Case(...))` in the final `sync` section at the bottom. Thus it's assumed that the address provided on the bus should be valid for exactly one cycle to create a one-cycle wide `.we` pulse.

However, the address is applied for multiple cycles. The AXI->Axilite and back again conversion can't complete in a single cycle, so the address is asserted for multiple cycles as the whole round-trip completes.

I'm guessing this bug may not have been noticed because not many systems trigger errors if a CSR is read from multiple times. In practice I don't have any blocks that break if this spec is violated, but, I would really like the axi-lite CSR implementation to behave spec-identical to the wishbone stuff, it would make me sleep better at night.

That being said, I am running into myriad issues trying to rework the system to obey this discipline. There are issues both in the Axi2Axilite converter, and also in the CSR interface itself. The Axi2Axilite converter is issuing two cycles for every one AXI request, and the CSR interface behavior doesn't match what's expected of AXI due to a mismatch in cycle delays, so the fix isn't simple.

I'll open a PR to address this issue but all my code also depends on PR #1506 being accepted, since I can't run simulations without it.

bunnie:
This PR would fix issue #1530. See the issue description for more details.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1197,13 +1197,19 @@ class SoC(LiteXModule, SoCCoreCompat):
                 delattr(self, name)
 
         # SoC CSR Interconnect ---------------------------------------------------------------------
+        if self.bus.standard == 'axi-lite':
+            reg = False
+        else:
+            reg = True
         self.csr_bankarray = csr_bus.CSRBankArray(self,
             address_map        = self.csr.address_map,
             data_width         = self.csr.data_width,
             address_width      = self.csr.address_width,
             alignment          = self.csr.alignment,
             paging             = self.csr.paging,
-            ordering           = self.csr.ordering)
+            ordering           = self.csr.ordering,
+            reg                = reg
+        )
         if len(self.csr.masters):
             self.csr_interconnect = csr_bus.InterconnectShared(
                 masters = list(self.csr.masters.values()),
--- a/litex/soc/interconnect/axi/axi_full_to_axi_lite.py
+++ b/litex/soc/interconnect/axi/axi_full_to_axi_lite.py
@@ -64,12 +64,13 @@ class AXI2AXILite(Module):
         )
         fsm.act("READ",
             # ar (read command)
-            axi_lite.ar.valid.eq(ax_beat.valid & ~_cmd_done),
+            axi_lite.ar.valid.eq(ax_beat.valid & ~_cmd_done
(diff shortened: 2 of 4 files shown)


bunnie:
This is being abandoned in preference for #1539

PR #1535, opened 2022-12-11
cpu/openc906: set extended CSRs based on D1 configuration

Icenowy:
Copy all feature and performance related CSR configuration from sun20i_d1_spl, which are values Allwinner suggests for D1 SoC.


enjoy-digital:
Thanks @Icenowy!

PR #1538, opened 2022-12-15
Fix sim module build on MacOS arm64

jevinskie:
Homebrew on arm64 Mac is now in /opt/homebrew by default.


enjoy-digital:
Thanks @jevinskie! That's merged.

PR #1540, opened 2022-12-20
soc: allow regions that are not a power of 2

sensille:
Align region to its power of 2, not the region size itself.


sensille:
Wait, this still doesn't look right.

sensille:
Shouldn't it rather be origin += (size_pow2 - origin%size_pow2)?

enjoy-digital:
Thanks @sensille, sorry for the delay. There was indeed an issue when doing the alignment, fixed with https://github.com/enjoy-digital/litex/commit/5760c5ba1e3d10adfc0ca889803d8bc4adec5df6.

enjoy-digital:
Closing since should be fixed/resolved.

sensille:
Sorry for the late reply. I just tested it and this version unfortunately does not fix the problem. 
```
ERROR:SoCRegion:Origin needs to be aligned on size:
ERROR:SoCRegion:Origin: 0x80002800, Size: 0x00004800, Mode: RW, Cached: False Linker: False

```

enjoy-digital:
THank

sensille:
The use case:
 
```
 self.add_ethernet(
                phy         = self.ethphy,
                data_width  = 32,
                ntxslots    = 1,
                nrxslots    = 8,
                with_timestamp = True,
)

```

enjoy-digital:
Thanks @sensille for the use case. I've just been able to have a closer look at it and your PR looks good and has been merged.

PR #1543, opened 2022-12-23
build/altera: Fix IP integration

stone3311:
This pull request fixes an issue where adding IP files causes an error when using the Quartus build platform.


enjoy-digital:
Thanks @stone3311!

Issue #1545, opened 2022-12-25
Meson 1.0.0 build issue


AEW2015:
Tools don't like the new Meson Release Dec 23.
```
OSError: Meson version to old. Found: 1.0. Required: 0.59.
Try updating with:
- pip3 install -U meson.
```

I had to force the old Meson version 0.59.

AEW2015:
Related issue:
Meson build doesn't seem to work with the new RISC-V GCC tool chain.
```
../../../pythondata-software-picolibc/pythondata_software_picolibc/data/meson.build:35:0: ERROR: Value "c18" (of type "string") for combo option "C language standard to use" is not one of the choices. Possible choices are (as string): "none", "c89", "c99", "c11", "gnu89", "gnu99", "gnu11".
```
Do I have the wrong one on the path?

eli-schwartz:
> Tools don't like the new Meson Release Dec 23.

https://github.com/enjoy-digital/litex/blob/e8322587a06f2f3e82443cc6422ba230b11c195a/litex/soc/integration/builder.py#L256-L266

This seems like a pretty odd check to me. It's checking to see if the major version is older than 0, OR if the minor version is older than 59.

Checking the major version doesn't even make *sense*, the minimum requirement here is "0".

Anyway, this check is definitely not a real version comparison. It should use a version tuple instead:

```python
if tuple(int(i) for i in meson_version[:3]) < (0, 59, 0):
         msg = f"Meson version too old. Found ......."
```

Because it's not an either-or.

You could also do this:
```meson
from packaging.version import Version

meson_version = Version(subprocess.check_output(["meson", "-v"]).decode("utf-8"))

if meson_version < Version('0.59.0'):
    msg = f"Meson version too old. Found ......."
```

That does require the https://pypi.org/project/packaging/ module, which is reasonably common at least. It also guarantees proper "version specifiers" support.

AEW2015:
The new `apt install riscv64-linux-gnu-gcc` default to gcc 7.5 which causes the second meson error listed above.

AEW2015:
Google colab uses Ubuntu 18 and therefore only installs the riscv64 gcc 7.5.  I had to manually install the newer RISCV binaries.

enjoy-digital:
[...]

PR #1546, opened 2022-12-27
builder: Fix logic for determining meson version compatibility

tmichalak:
Recently PyPi version of meson changed to 1.0.0 which uncovered this bug which caused the error condition to triggger falsely.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -262,7 +262,7 @@ class Builder:
             msg = "Unable to find valid Meson build system, please install it with:\n"
             msg += "- pip3 install meson.\n"
             raise OSError(msg)
-        if (int(meson_version[0]) < meson_major_min) or (int(meson_version[1]) < meson_minor_min):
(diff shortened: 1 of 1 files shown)


enjoy-digital (approved):

enjoy-digital:
Thanks @tmichalak!

litex/soc/integration/builder.py
```diff
@@ -262,7 +262,7 @@ def _check_meson(self):
             msg = "Unable to find valid Meson build system, please install it with:\n"
             msg += "- pip3 install meson.\n"
             raise OSError(msg)
-        if (int(meson_version[0]) < meson_major_min) or (int(meson_version[1]) < meson_minor_min):
+        if (int(meson_version[0]) < meson_major_min) or (int(meson_version[0]) == meson_major_min and int(meson_version[1]) < meson_minor_min):
```
eli-schwartz, line 265: See https://github.com/enjoy-digital/litex/issues/1545#issuecomment-1364736945

This check is still problematic and should be majorly refactored rather than hotfixed.

PR #1547, opened 2022-12-29
Fix WID in axi versions

navan93:
There is no WID in AXI4, its there only in AXI3.

Probably was a typo in the first place


enjoy-digital:
Thanks @navan93, this was indeed probably a typo in the first place.

PR #1548, opened 2022-12-31
cores/jtag: Add more Altera part numbers

stone3311:
This pull request adds more part numbers to the JTAG primitive dictionary. It also changes `get_primitive` to return the closest match to the beginning of a part number.

This is required for parts such as `EP2S`, which shares the same signature as `EP2SGX` while using different JTAG primitives.


enjoy-digital:
Great, thanks a lot @stone3311!

PR #1549, opened 2023-01-02
build/xilinx/vivado: fix verilog include paths

michalsieron:
a286d77e introduced a bug, where `-include_dirs` parameter is incorrectly defined.

Following TCL code is being generated:
```tcl
synth_design -directive default -top digilent_arty -part xc7a35ticsg324-1L -include_dirs \{.join(self.platform.verilog_include_paths)}\}
```

Below is an explanation why it didn't work:
Python's f-strings escape curly braces using double curly braces like so `{{` instead of using backslash `\{`.
What's more, you need to alternate single and double quotations marks when using strings in curly braces expression otherwise two string objects are being generated and errors like this one can happen.


enjoy-digital:
Thanks @michalsieron!

PR #1551, opened 2023-01-04
replace Meson version check with a specification-compliant version comparator

### LINKED ISSUE #1545: Meson 1.0.0 build issue

eli-schwartz:
The current check compares the integers split out from `meson --version` one by one. This is an ad-hoc version comparison algorithm with a few flaws, notably that it doesn't truly understand how version components fit together, and that broke once Meson bumped the major version. There are other potential issues that could show up but haven't yet, such as versions with words in them (release candidates).

The packaging module is a high-quality library that provides a standard version parsing algorithm, with which you can simply say "is this version object greater than that one". Use it instead.

Fixes #1545

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -18,6 +18,8 @@ import subprocess
 import struct
 import shutil
 
+from packaging.version import Version
+
 from litex import get_data_mod
 from litex.gen import colorer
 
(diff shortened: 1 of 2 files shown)


enjoy-digital (approved):
Thanks @eli-schwartz, this looks good. Let's run it in CI and merge if passing.

PR #1558, opened 2023-01-09
liblitedram: SPD read improvements

michalsieron:
- add generic `sdram_read_spd` function, that automatically changes SPD pages
- make `sdram_spd` command read entire SPD
- print memory sizes using `print_size` function, which selects appropriate unit (`B/KiB/MiB/GiB`)
- make printed SDRAM size match, actually supported memory size by litedram. Read module size from the SPD and multiply it by the number of modules defined by litedram.
- do not choose between printing either main RAM size or SDRAM size, as they might differ. Rather print both if available

This PR requires changes in litedram from PR https://github.com/enjoy-digital/litedram/pull/320:
- definition of `SDRAM_PHY_SUPPORTED_MEMORY` as a fallback when SPD is not available
- definition of `SDRAM_PHY_[DDR3|DDR4|...]` to select appropriate code paths when reading and parsing the SPD

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -227,7 +227,8 @@ class Builder:
             from litedram.init import get_sdram_phy_c_header
             sdram_contents = get_sdram_phy_c_header(
                 self.soc.sdram.controller.settings.phy,
-                self.soc.sdram.controller.settings.timing)
+                self.soc.sdram.controller.settings.timing,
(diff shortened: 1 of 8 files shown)


enjoy-digital:
Thanks @michalsieron! This looks good and is merged (alongs with https://github.com/enjoy-digital/litedram/pull/320).

PR #1512, opened 2022-11-16
Liblitedram cleanup

michalsieron:
This PR introduces a number of changes that clean up existing liblitedram code.
It extracts existing functions like `write_rst_dqs_delay` or `write_inc_dq_delay` to separate file - `accessors.h`.
As all those functions are used in the same way:
- select module (and optionally specific line)
- increment or reset current delay
- deselect module and line

a new function was added `sdram_leveling_action`, which abstracts this away.

Support for x4 IC configuration was fixed and ability to train individual DQ lines was added.

Additionally, comments were added to `#endif`s, which indicate conditions of their `#if`s.

All changes were tested on Digilent Arty A7 board and [Antmicro Data Center DRAM Tester](https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/antmicro_datacenter_ddr4_test_board.py)

PR #1560, opened 2023-01-10
Add `sdram_hw_test` command

michalsieron:
This PR adds `sdram_hw_test` command, which uses BIST modules that are enabled when `with_bist=True` in `add_sdram` method.

It allows performing a memtest (similar to `sdram_test`), but using DMAs and bypassing system bus. This way, one can test ranges above 4 GiB, which was the limit with `sdram_test` due to address_width being limited to 32 bits.

Additionally I reworked already existing BIST functions a little bit:
- fixed `SDRAM_TEST_DATA_BYTES` definition
- extracted code to `sdram_bist_[write|read]` functions
- made global variables static
- fixed formatting

It would be best if https://github.com/enjoy-digital/litedram/pull/321 would be merged as well, although they should be independent.

### DIFF
--- a/litex/soc/software/bios/cmds/cmd_litedram.c
+++ b/litex/soc/software/bios/cmds/cmd_litedram.c
@@ -89,6 +89,46 @@ static void sdram_bist_handler(int nb_params, char **params)
 define_command(sdram_bist, sdram_bist_handler, "Run SDRAM Build-In Self-Test", LITEDRAM_CMDS);
 #endif
 
+/**
+ * Command "sdram_hw_test"
+ *
+ * Run SDRAM HW-accelerated memtest
+ *
+ */
+#if defined(CSR_SDRAM_GENERATOR_BASE) && defined(CSR_SDRAM_CHECKER_BASE)
+static void sdram_hw_test_handler(int nb_params, char **params)
+{
+	char *c;
+	uint64_t origin;
+	uint64_t size;
+	uint64_t burst_length = 1;
+	if (nb_params < 2) {
+		printf("sdram_hw_test <origin> <size> [<burst_length>]");
+		return;
+	}
+	origin = strtoull(params[0], &c, 0);
+	if (*c != 0) {
+		printf("Incorrect origin");
+		return;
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks, this looks good. 

For `sdram_bist`, have you noticed performance difference between the previous version and new one? In the previous version, we were preparing the next test in advance to avoid as much as possible inactivity periods on the DRAM, this is no longer the case here but if tests reports the same results (in sim and hardware tests), I'm fine removing this since this indeed simplify code.

Otherwise, for https://github.com/enjoy-digital/litedram/pull/321, this will have to be studied more carefully, as explained, I don't think the suggested implementation will be suitable, so let's first try to merge this PR and  https://github.com/enjoy-digital/litex/pull/1512 and we'll then see for the PRBS/LFSR improvements.

michalsieron:
I did not notice any slowdowns after my changes regarding preparing and executing DMA operations.
However, the `sdram_bist` is now slower due to my change of `SDRAM_TEST_DATA_BYTES` definition.

The original definition `#define SDRAM_TEST_DATA_BYTES (CSR_SDRAM_DFII_PI0_RDDATA_SIZE*4)` could either mean:
1. _RDDATA bytes times 4 phases_ (byte width of the DRAM port)
2. or _count of 4-byte RDDATA subregisters_ (byte width of one phase's RDDATA).

I assumed, the original definition to be (1.), in which case it was incorrect, as CSR size is defined as subregister count (also there can be more than 4 phases).

My new definition `#define SDRAM_TEST_DATA_BYTES (SDRAM_PHY_DFI_DATABITS / 8 * SDRAM_PHY_PHASES)` corrects that.

However, if the original definition was meant to be (2.), then my change modifies the `sdram_bist`'s `burst_length` parameter meaning, as it no longer is a count of phases, but rather entire transfers.

If that is undesirable, I can modify my changes to keep the original definition and turn my definition into a new one.

michalsieron:
Fixed format warnings that were already present in `bist.c`. Just like in #1574

PR #1561, opened 2023-01-10
Add Arch Linux setup support for RISC-V and OpenRISC toolchains

roby2014:
The current setup script does not install the needed dependencies for Arch Linux machines (such as Risc-V and OpenRISC).
This PR adds support, installing these packages:
- riscv : https://archlinux.org/packages/community/x86_64/riscv64-linux-gnu-gcc/
- openrisc : https://archlinux.org/packages/community/x86_64/or1k-elf-gcc/

I also found [this](https://aur.archlinux.org/packages/powerpc-linux-gnu-gcc) AUR package for the PowerPC toolchain, but since its AUR, I'm not sure if I should include it.

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -341,7 +341,6 @@ def riscv_gcc_install():
     # ------
     if sys.platform.startswith("linux"):
         os_release = (open("/etc/os-release").read()).lower()
-        print(os_release)
         # Fedora.
         if "fedora" in os_release:
             os.system("dnf install gcc-riscv64-linux-gnu")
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @roby2014! Using the AUR package for PowerPC should be fine we we could switch to official package if available in the future. You can add it if you want and I'll merge the PR.

mithro:
FYI - @shenki @antonblanchard

roby2014:
@enjoy-digital I added the PowerPC package via AUR (https://aur.archlinux.org/packages/powerpc64le-linux-gnu-gcc), all should be good now.

enjoy-digital:
@roby2014: Thanks, let's merge.

PR #1564, opened 2023-01-11
software/liblitedram: fix write leveling for x4 modules

michalsieron:
When using x4 modules, their response makes up only half a byte. We need to extract the nibble and then test it.


enjoy-digital:
Thanks @michalsieron, that's merged.

PR #1566, opened 2023-01-12
liblitedram/sdram_spd: fix invalid buffer index

michalsieron:
Fix a silly error with indexing output buffer. The first version of this function was supposed to read the entire SPD and after adding `len` parameter I forgot to edit the used index.

What's unlucky is that its current two uses (reading from the first page and reading the entire SPD) didn't expose this bug.


enjoy-digital:
Thanks @michalsieron.

PR #1574, opened 2023-01-18
liblitedram/utils: fix format warnings

### LINKED ISSUE #1573: printf size warnings on 64bit architectures
commit https://github.com/enjoy-digital/litex/commit/fb068f6e introduces a duplicate set of `print_size()` and `print_progress()` functions specific to liblitedram, but uses a combination of `uint64_t size` and `%llu` with printf, which only works on 32-bit architectures (probably vexriscv, if I had to guess :) ). On 64bit (rocket) I get  a warning that `%lu` should be used with `uint64_t`...

Can we revert this commit and update the existing (`libbase/memtest.c`)[https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/memtest.c#L168] (and friends) version to use `long long unsigned` (always 64-bit everywhere) instead?

Any other thoughts (@enjoy-digital @michalsieron)? I think it's rather dumb that `uint64_t` is `unsigned long` as far as 64bit arch library printf and gcc is concerned, but I'm sure the library and gcc people will explain why I'm an idiot for wanting it changed there, so here we are :)

michalsieron:
Use format constants for fixed width integer types to make it work on both 32-bit and 64-bit CPUs without warnings.

Fixes #1573

### DIFF
--- a/litex/soc/software/liblitedram/utils.c
+++ b/litex/soc/software/liblitedram/utils.c
@@ -14,18 +14,18 @@
 
 void print_size(uint64_t size) {
 	if (size < KIB)
-		printf("%lluB", size);
+		printf("%" PRIu64 "B", size);
 	else if (size < MIB)
-		printf("%llu.%lluKiB", size/KIB, (size/1   - KIB*(size/KIB))/(KIB/10));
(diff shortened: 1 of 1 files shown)


gsomlo (approved):
inhibits the warnings I was getting on rocket (rv64gc), so LGTM -- thanks!

enjoy-digital:
Thanks @michalsieron for the PR and @gsomlo for the review. This is merged.

PR #1580, opened 2023-01-24
soc/software/libXX/YY.h: adding extern C (required to link with cpp code)

trabucayre:
I'm trying to use LiteX libraries with a cpp code (a simulation using verilator), but at link step compilation fails with error like:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/11/../../../../x86_64-pc-linux-gnu/bin/ld: top_tb.o: in function `main':
top_tb.cpp:(.text.startup+0x1ca): undefined reference to `sdram_init()'
collect2: error: ld returned 1 exit status
make[1]: *** [Vtop_tb.mk:66: Vtop_tb] Error 1
```
This is due to a different naming between C and CPP

by adding, before functions declaration in headers:
```C
#ifdef __cplusplus
extern "C" {
#endif
```

And, these lines, at the header end
```
#ifdef __cplusplus
}
#endif
```

The problem is fixed and cpp code may be linked to LiteX libraries.

### DIFF
--- a/litex/soc/software/libbase/memtest.h
+++ b/litex/soc/software/libbase/memtest.h
@@ -1,6 +1,10 @@
 #ifndef __MEMTEST_H
 #define __MEMTEST_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #include <stdbool.h>
 
 #include <generated/csr.h>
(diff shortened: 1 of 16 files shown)


enjoy-digital:
Thanks @trabucayre, this is merged. In the long term it will be good to add a simple test that we can run in CI to report the problem in case we break it (while adding new files for example without this #ifdefs).

Issue #1612, opened 2023-02-23
Hardware interrupt handling functions not all generated in csr.h file


mohamedElbouazzati:
Hi @enjoy-digital, 

I'm using an external interrupt with my CPU named dio0, I'm trying to generate the required functions to handle my interrupt from the software. 
But the generated csr.h is limited to a couple of functions, not all the ones I'm used to generate with an older version of litex.
Do you have any idea about this situation?

Thank you
 
**Used in target :** 
```
 self.submodules.dio0 = GPIOIn(self.platform.request("dio0"),with_irq=True )
 self.irq.add("dio0")
```
**Generated csr.h :**
```
/* dio0 */
#define CSR_DIO0_BASE (CSR_BASE + 0x1800L)
#define CSR_DIO0_IN_ADDR (CSR_BASE + 0x1800L)
#define CSR_DIO0_IN_SIZE 1
static inline uint32_t dio0_in_read(void) {
	return csr_read_simple((CSR_BASE + 0x1800L));
}
#define CSR_DIO0_MODE_ADDR (CSR_BASE + 0x1804L)
#define CSR_DIO0_MODE_SIZE 1
static inline uint32_t dio0_mode_read(void) {
	return csr_read_simple((CSR_BASE + 0x1804L));
}
static inline void dio0_mode_write(uint32_t v) {
	csr_write_simple(v, (CSR_BASE + 0x1804L));
}
#define CSR_DIO0_EDGE_ADDR (CSR_BASE + 0x1808L)
#define CSR_DIO0_EDGE_SIZE 1
static inline uint32_t dio0_edge_read(void) {
	return csr_read_simple((CSR_BASE + 0x1808L));
}
static inline void dio0_edge_write(uint32_t v) {
	csr_write_simple(v, (CSR_BASE + 0x1808L));
}

/* identifier_mem */
```

**Desired csr.h( generated with a litex old version):**
```
/* dio0 */
#define CSR_DIO0_BASE (CSR_BASE + 0x0L)
#define CSR_DIO0_IN_ADDR (CSR_BASE + 0x0L)
#define CSR_DIO0_IN_SIZE 1
static inline uint32_t dio0_in_read(void) {
	return csr_read_simple(CSR_BASE + 0x0L);
}
#define CSR_DIO0_MODE_ADDR (CSR_BASE + 0x4L)
#define CSR_DIO0_MODE_SIZE 1
static inline uint32_t dio0_mode_read(void) {
[...]

PR #1613, opened 2023-02-23
Fix GPIO IRQ CSR generation

adamhlt:
Hello, in relation to this issue (#1612)  I suggest a fix for other people who may have the same problem.
It seems that the generation of CSR for GPIOIN with IRQ enable is changed with the updates.
I hope I helped.


enjoy-digital:
Thanks @adamhlt! This is merged, I just moved the `self.ev.finalize()` outside the for loop to ensure finalize is done after adding the different EventSourceProcesses.

PR #1627, opened 2023-03-01
soc/cores/clock: adding CologneChip CC_PLL

trabucayre:
This PR adds the `CC_PLL` primitive for the **CologneChip GateMate* 

- the  input clock may be a clock compatible pin or an internal signal
- 4 output clocks are available: first has a 0° phase shift, second has a 90° phase, third a 180° phase and the last 270°. First and second output provides a signal at the frequency fixed by `OUT_CLK`, last two clocks may be x1 or x2 `OUT_CLOCK`
- `CC_PLL` may or not provides a signal before DCO is locked

\cc @pu-cc

### DIFF
--- a/litex/soc/cores/clock/colognechip.py
+++ b/litex/soc/cores/clock/colognechip.py
@@ -1,0 +1,151 @@
+#
+# This file is part of LiteX.
+#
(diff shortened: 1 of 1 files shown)


enjoy-digital:

PR #1629, opened 2023-03-02
build/efinix/efinity: delays iface.py execution after project xml was written

trabucayre:
The first time an efinix's target is executed `iface.py` produces errors like:

```
Efinity project file does not exists : ../gateware/efinix_titanium_ti60_f225_dev_kit.xml
File error : Efinity project file does not exists : ../gateware/efinix_titanium_ti60_f225_dev_kit.xml
Efinity project file does not exists : ../gateware/efinix_titanium_ti60_f225_dev_kit.xml
File error : Efinity project file does not exists : ../gateware/efinix_titanium_ti60_f225_dev_kit.xml
```

The second time these errors disapears.

This is due to the order where scripts are build. Currently:
- `peri.xml` is build by `build_io_constraints()`, `iface.py` is also called here
- `project.xml` is build by `build_project()`

But since `build_io_constaints()` is called before `build_project()` `iface.py` can't have access to the project's xml. This is no more true for next build because files are already written at the previous run.

This PR delays `iface.py` call after project's xml was build to fix this dependency.

### DIFF
--- a/litex/build/efinix/efinity.py
+++ b/litex/build/efinix/efinity.py
@@ -204,21 +204,6 @@ class EfinityToolchain(GenericToolchain):
 
         tools.write_to_file("iface.py", header + gen + gpio + add + footer)
 
-        if tools.subprocess_call_filtered([self.efinity_path + "/bin/python3", "iface.py"], common.colors) != 0:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre, this makes sense.

PR #1631, opened 2023-03-04
add a proxy method to access bitstream extension

trabucayre:
When getting bitstream's filename the current behaviour for *soc/integration/builder* is to use directly `soc.platform.bitstream_ext` for **sram** mode and returning **bin** for flash. This imply, at **targets** level, for flash mode, to use `get_bitstream_filename`'s *ext* parameter to force correct extension.

This PR add, at *build/generic_platform* a method to return extension according to `mode` (may be the same extension for both case when *build/vendor/platform.py* declare `_bitstream_ext` as a string or two different extension when `_bitstream_ext` is a dict).

This also fix extension for anlogic (*fs* is for *gowin* FPGA).

### DIFF
--- a/litex/build/altera/platform.py
+++ b/litex/build/altera/platform.py
@@ -13,7 +13,10 @@ from litex.build.altera import common, quartus
 # AlteraPlatform -----------------------------------------------------------------------------------
 
 class AlteraPlatform(GenericPlatform):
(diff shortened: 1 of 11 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #1632, opened 2023-03-06
soc/cores/cpu/vexriscv_smp/core.py: fix variants for external incoming values and remove default timer0 and uart in standard mode.

jorislee:
Fix: VexRiscvSMP core variants for external incoming values and remove default timer0 and uart in standard mode.


enjoy-digital:
Thanks @jorislee, this is merged.

PR #1634, opened 2023-03-07
litex_ok.py bitstream configuration example

AEW2015:
This uses swig python binaries to use the selectmap Via USB to program the FPGA.
The *.so files are normally git ignored. Not sure how you want to handle it.
Not sure if the Opal Kelly interface should be a separate LiteX tool.
I want to include UART, Debug, Data, and Wishbone access through it later.

Here is an example output:
``` 
../../../counters.bit
 Product: XEM8320-AU25P
Firmware version: 1.53
 Serial Number: 2206000YVD
 Device ID: Opal Kelly XEM8320
FPGA configuration succeded in 0.5916 seconds
```


enjoy-digital:
Hi @AEW2015,

thanks a lot, this will indeed be very interesting to be able to use Opal Kelly tools directly from LiteX. I have a preference to not integrate these tools directly in LiteX (at least for now) but we could develop this in https://github.com/litex-hub/pythondata-misc-opal-kelly that I just created and you should also receive an invitation for LiteX-Hub.

Once this is mature enough, we could allow installation of this package with LiteX and maybe also integrate directly some of the scripts.

Unfortunately, for now I don't have Opal Kelly boards, so can't test it yet. But I'll probably need the Artix Ultrascale+ one in the future for projects.

Florent

AEW2015:
Do you have a starting template you want to use for that new repo?

enjoy-digital:
@AEW2015: We could do something similar to https://github.com/litex-hub/pythondata-misc-usb_ohci. So for now, you could create a `pythondata_misc_opal_kelly` directory in it and put your development there.

PR #1636, opened 2023-03-07
CVA6: Adding RV32 support

gatecat:
Marking this as a draft for now because:
 - ~~the most invasive change was around the debug AXI integration (removing the Xilinx IP cores and replacing them with a Pulp one where needed), which I haven't got a setup to test yet (ideally we could have a way of testing JTAG in sim?)~~ tested and confirmed working using jtagremote, also fixing for 64-bit configs at the same time
 - in general it hasn't seen much testing beyond booting the BIOS
 - for some strange reason (suspected stack corruption) the compiled verilator binary segfaults when using g++/gcc as compiler; when using clang for the verilated source it works fine

### DIFF
--- a/litex/soc/cores/cpu/cva6/boot-helper.S
+++ b/litex/soc/cores/cpu/cva6/boot-helper.S
@@ -4,6 +4,23 @@
 .global smp_ap_target
 .global smp_ap_ready
 
+#if __riscv_xlen == 32
+
+boot_helper:
+  // boot core saves args and jump target for ap cores:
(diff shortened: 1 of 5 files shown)


gatecat:
This is now tested and working on hardware.

I noticed some occasionally unreliable UART output and failing litedram calibration that was fixed by switching from a 32-bit to 64-bit system bus width (CVA6 uses a 64-bit bus even in its 32-bit config); I'm not sure if this is a bug or just not a supported mode (with things like CSR accesses not behaving well unaligned?)

PR #1641, opened 2023-03-09
feature: Support renaming default uart to uart0.

jorislee:
To keep the names aligned, it is supported to redirect the default uart name to uart0.


enjoy-digital:
Thanks @jorislee, this is indeed a limitation that need to be addressed for quite some time now. I'll have a closer look at this and try to see what would be the best approach for all peripherals.

PR #1643, opened 2023-03-11
add docstrings to bitbang

jeremyherbert:
Hi,

I asked on the discord how to add documentation to litex, and the suggestion was to start by adding some docstrings. I have had a go at a small file here (`bitbang.py`), and I would appreciate any feedback on this. I would also like to know if the following would be accepted:

- Changes to whitespace to bring the files in line with PEP8
- Inline typehints where possible

Thanks.

### DIFF
--- a/litex/soc/cores/bitbang.py
+++ b/litex/soc/cores/bitbang.py
@@ -12,29 +12,41 @@ from litex.soc.interconnect.csr import *
 # I2C Master Bit-Banging ---------------------------------------------------------------------------
 
 class I2CMaster(Module, AutoCSR):
-    """I2C Master Bit-Banging
+    """I2C bus master (bit-banged).
 
-    Provides the minimal hardware to do software I2C Master bit banging.
+    This core provides minimal hardware for use as a software controlled bit-banged I2C bus master. I2C uses a
+    tristate/open-drain output driver configuration with pull-up resistors, and this core expects that the pull-ups will
+    be provided externally.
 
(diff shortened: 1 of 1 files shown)


jeremyherbert:
I have just updated the file with PEP8 whitespace fixes (as far as I can tell) and type hints. The type hints exist mostly for IDEs like VSCode or Pycharm and shouldn't break anything because the python versions that don't support type hints are all obsolete. In particular, I have found type hints to be more useful for newbies because the IDE can at least warn you that you are probably doing something silly.

enjoy-digital:
Thanks @jeremyherbert, I now see what you meant with whitespaces, but in fact we'll keep this since I like to align things to improve code presentation. Type hints could be interesting and I'll think about it and we'll maybe wait a bit to add them (this should probably done globally)).

enjoy-digital:
Thanks @jeremyherbert, I now see what you meant with whitespaces, but in fact we'll keep this since I like to align things to improve code presentation. Type hints could be interesting and I'll think about it and we'll maybe wait a bit to add them (this should probably done globally)).

enjoy-digital:
@jeremyherbert: I merged the PR but reverted the last commit and just did minor cosmetic changes on top of your first commit with https://github.com/enjoy-digital/litex/commit/38282effd48ae6cf85c22cf6ec98b37e0f28523d. Regarding adding type hints, I'll still consider it but that's maybe not the right moment to add it to a single core.

Thanks a lot for your contribution!

PR #1646, opened 2023-03-13
soc/cores/clock/gowin_gw2a: enable GW2AR support

Icenowy:
Gowin GW2AR is just GW2A with co-packaged external RAM.

Enable using GW2APLL for GW2AR.


enjoy-digital:
Thanks @Icenowy!

PR #1647, opened 2023-03-13
build/efinix: add option use active or passive spi mode.

te-johan:
we are using spi passive mode to load our image and need to generate a bitstream for it.
adding the value to the platform since it seems the most common practice, but it can be handled by the toolchain instead.


enjoy-digital:
Thanks @te-johan!

PR #1648, opened 2023-03-14
litex_term: improve connection setup

dalegaard:
The static delay in the current implementation is not always long enough for OpenOCD to start up and find the target. This adds simple retry logic, waiting up to five seconds but still connecting as quickly as possible.


enjoy-digital:
Thanks @dalegaard!

PR #1650, opened 2023-03-16
soc/cores/clock/gowin_gw1n: improve VCO configuration

trabucayre:
The PLL configuration for **Gowin** FPGA search for the first combinaison of `IDIV`, `ODIV` and `FDIV` compatible with *VCO* limitations, and margin, to generate user's requested frequency. But for a `clkin` and `clkout` couple more than one set of parameters may respect criterions.

For example with a 27MHz input frequency and to generate a 96MHz these set are possible

| `IDIV` |  `ODIV` | `FDIV` | real freq (Hz) | difference (Hz) |
|---------|------------|----------|--------------------|---------------------|
|7         |  8          |   25     | 96428571.42857143 | 428571.4285714328 |
| 9        | 8           |   32     | 96000000.0               | 0.0                             |

As shown the best configuration isn't the first, so with current implementation the PLL produces a 96.42MHz instead of 96.00MHz

This PR, instead of returning with the first acceptable combo, store all acceptables sets into a dict, store diff also. After the end of the loop, it select best one (the one with the smallest difference and return the corresponding set of parameters).

This PR, also, modify `FDLY`: with **GW1N-1** and **GW1N-1S** a 0 delay steps is achieved by setting with input with 0 but for others models a value of 0x0F is required.

### DIFF
--- a/litex/soc/cores/clock/gowin_gw1n.py
+++ b/litex/soc/cores/clock/gowin_gw1n.py
@@ -120,9 +120,9 @@ class GW1NPLL(Module):
         self.nclkouts += 1
 
     def compute_config(self):
-        config = {}
+        configs = [] # corresponding VCO/FBDIV/IDIV/ODIV params + diff
+
         for idiv in range(1, 64):
-            config["idiv"] = idiv
             pfd_freq = self.clkin_freq/idiv
             pfd_freq_min, pfd_freq_max = self.pfd_freq_range
             if (pfd_freq < pfd_freq_min) or (pfd_freq > pfd_freq_max):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre! This looks good. As a minor detail, I think we could avoid the two different lists and just have a single `possible_configs` list that would also store `diff_freq` as an element similar to` idiv`/`odiv`/`...`.

What do you think? I could have a look at this if you agree but don't have/want to spend the time on this.

trabucayre:
I have updated the PR has requested.
now `diff` is included into the dict.

enjoy-digital:
Thanks @trabucayre, looks perfect now :)

PR #1653, opened 2023-03-20
Add mcpwm

bentwire:
Add an MCPWM Module.  

Just me figuring out how to create a new module.  Don't accept this PR if you don't want/need it.

If you do want it I can fix whatever is a mess. :)

### DIFF
--- a/litex/soc/cores/led.py
+++ b/litex/soc/cores/led.py
@@ -31,19 +31,19 @@ class LedChaser(Module, AutoCSR):
         chaser = Signal(self.n)
         mode   = Signal(reset=_CHASER_MODE)
         timer  = WaitTimer(int(period*sys_clk_freq/(2*self.n)))
-        self.leds   = Signal(self.n)
+        leds   = Signal(self.n)
         self.submodules += timer
         self.comb += timer.wait.eq(~timer.done)
         self.sync += If(timer.done, chaser.eq(Cat(~chaser[-1], chaser)))
         self.sync += If(self._out.re, mode.eq(_CONTROL_MODE))
         self.comb += [
             If(mode == _CONTROL_MODE,
-                self.leds.eq(self._out.storage)
+                leds.eq(self._out.storage)
             ).Else(
-                self.leds.eq(chaser)
+                leds.eq(chaser)
             )
         ]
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Hi @bentwire,

thanks for the PR. For multiple channels/peripherals, we generally have a preference to do this at the top level with multiple instance of the same peripherals, ex in your case, as done in Colorlite project: https://github.com/enjoy-digital/colorlite/blob/master/colorlite.py#L113-L118

```python3
# Servos -----------------------------------------------------------------------------------
from litex.soc.cores.pwm import PWM
for n in range(4):
    servo_pad = platform.request("servo", n)
    servo     = PWM(servo_pad)
    setattr(self.submodules, f"servo{n}", servo)
```

You could argue that ressource usage is probably not optimal since not shared, but that's a tradeoff I generally do these simple peripherals between simplicity/resource usage/maintainability.

So I have a preference to only keep the simple PWM core and handle the multi channel at the top level.

trabucayre:
@enjoy-digital with @bentwire's approach all output have the same (with a rising edge aligned/coherent) period. With multiple PWM instances: thjis coherency can't guaranted since all have to be configured and enabled by multiple write access (and consequently not at the same date).
But of course maybe I'm wrong

bentwire:
@trabucayre  @enjoy-digital

Isn't it more power efficient to have one counter for all the channels instead of (from your example) 16 separate counters?  I was thinking all those transitions could be power hungry.

I don't have any use for it, but what if someone wanted a 64 channel PWM for controlling something?  Won't having 64 counters all counting at the same time be noisy and power hungry from a hardware point of view?

Just asking since I am near the limit of my knowledge so far and still have lots to learn. :)

bentwire:
It looks good, and has at least one advantage over my method.

trabucayre:
@enjoy-digital : Looks good to me too.

enjoy-digital:
@bentwire, @trabucayre: Thanks for the feedback, I'll do more test/check on it. @bentwire your contributions are more than welcome, I'm just trying to avoid duplication as much as possible for maintenance. If you need more features for the PWM module, we could integrate them in the base PWM module and the Multi-Channel one will automatically get them.

bentwire:
@enjoy-digital Perfect.  I will do that instead.  Feel free to close this, or I can if you want.

PR #1654, opened 2023-03-22
cpu: rocket, cva6: Fix compilation with newer binutils

shenki:
Resolves the error:

 cpu/rocket/irq.h:23: Error: unrecognized opcode `csrr a4,mstatus',
 extension `zicsr' required

Rocket and cva6 missed this in commit 0e2a1b54a454bcbf as they are 64bit CPUs, and that change only updated the 32bit CPUs.

Tested with litex_sim and riscv64-unknown-elf-ld 2.40-2+4+b1 (Debian Bookworm).


enjoy-digital:
Thanks @shenki! That's merged.

PR #1656, opened 2023-03-22
cpu/rocket: rework variant naming convention

gsomlo:
@enjoy-digital -- please don't apply, I need to push the corresponding `pythondata-cpu-rocket` changes *first*!

The naming convention for LiteX Rocket variants has become overly complex. Simplify it while at the same time adding more flexibility.

There is a new set of instances of varying main RAM memory bus port width (1x (64bit), 2x (128bit), 4x (256bit), and 8x (512bit)), of each of the following principal LiteX specific Rocket models:

- small:  (rv64imac, no MMU, no S, no FPU)
- medium: (rv64imac, adds MMU and S-mode)
- linux:  (rv64imafdc, adds FPU, supports linux distros)
- full:   (rv64imafdcbkph[+], adds hypervisor support)

NOTE: before adding H support, the feature set of the old `full` model is now represented by the `linux` model. The old `linux` did not use to have an FPU, and is now available as `medium`.

In addition to the range of memory port widths, each model will be instantiated in 1, 2, 4, and 8 core variants. The naming convention is `LitexConfig_<model>_<num_cores>_<mem_width>`.

E.g. `LitexConfig_full_8_2` for an 8-core full model with a 128bit main RAM AXI port. On the build command line, this example would look like:

	...
	--cpu-type rocket --cpu-variant full \
	--core-count 8 --mem-width 2 \
	...

There are a total of 4 * 4 * 4 = 64 (sub-)variants: each of the four principal models can be fitted with one of four core counts, and one of four memory bus widths.

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -30,6 +30,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import os
+import logging
 
 from migen import *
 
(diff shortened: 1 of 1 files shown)


litex/soc/cores/cpu/rocket/core.py
```diff
@@ -44,89 +44,50 @@
...
+        cpu_group = parser.add_argument_group(title="CPU options")
+        cpu_group.add_argument("--core-count", default=1, help="Number of cores (1, 2, 4, 8).", type=int)
+        cpu_group.add_argument("--mem-width",  default=1, help="Width of memory port (1, 2, 4, 8).", type=int)
+
+    @staticmethod
+    def args_read(args):
+        if int(args.core_count) in [1, 2, 4, 8]:
+            Rocket.core_count = int(args.core_count)
+        else:
+            Rocket.core_count = 1 # FIXME: warn user!!!
```
gsomlo, line 78: I haven't seen an example of another CPU family logging warnings or errors, but it'd be nice to send out a LiteX-y warning here if the user requests any other core count (or memory port width, see below) than 1, 2, 4, or 8, for which we have pre-built verilog... :)

litex/soc/cores/cpu/rocket/core.py
```diff
@@ -44,89 +45,53 @@
...
+            Rocket.core_count = int(args.core_count)
+        else:
+            logger.error("Invalid core count {} (should be 1, 2, 4, or 8)".format(args.core_count))
+            quit()
+        if int(args.mem_width) in [1, 2, 4, 8]:
+            Rocket.mem_width = int(args.mem_width)
+        else:
+            logger.error("Invalid memory port width {} (should be 1, 2, 4, or 8)".format(args.mem_width))
+            quit()
+
```
gsomlo, line 87: @enjoy-digital -- since I only need logging in `args_read()` (which being `@static` may be run before the `Rocket` class `__init__` method, I figured keeping it localized like this should work, for now (and we can change it later if it turns out we need logging in more places than just here).

If this is OK (and you're otherwise OK with the overall change), I'll push both the `pythondata-cpu-rocket` update and then this one sometime later this evening, after some more testing.

Thanks!

enjoy-digital: This looks good to me. I'm also wondering if we should have a convention for the CPU parameter and start them with `cpu-` to ease parameter understanding when exposed on the target. This is not something that has been done yet on the other CPU, but have been thinking about it previously. So since we are introducing some parameters to Rocket, maybe we could think about it and do it directly?

enjoy-digital (approved):

gsomlo:
On Wed, Mar 22, 2023 at 09:05:44AM -0700, enjoy-digital wrote:
> This looks good to me. I'm also wondering if we should have a convention for
> the CPU parameter and start them with cpu- to ease parameter understanding when
> exposed on the target. This is not something that has been done yet on the
> other CPU, but have been thinking about it previously. So since we are
> introducing some parameters to Rocket, maybe we could think about it and do it
> directly?

Good idea, arguments are now `--cpu-num-cores` and `--cpu-mem-width`.

Pushed as 4b2b25d3.

PR #1657, opened 2023-03-23
WIP Allow the verilog name of the toplevel to be customized separately from the filename

hansfbaier:
I need this for MiSTeX, because there are core names like Arcade-BombJack
and verilog does not like the minus sign in the name.
But I want the project file be named like that, because that is how the bitstream
filenames appears in the MRA files.

### DIFF
--- a/litex/build/generic_toolchain.py
+++ b/litex/build/generic_toolchain.py
@@ -56,12 +56,14 @@ class GenericToolchain:
     def build(self, platform, fragment,
         build_dir      = "build",
         build_name     = "top",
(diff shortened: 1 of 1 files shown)


hansfbaier:
Do you think `toplevel_name` is a good name for the parameter, would `module_name` or `verilog_name` be  better?

hansfbaier:
put into WIP state, still needs a fix

hansfbaier:
Closing this, would need to change in all other toolchains. I think I'll better work around this

PR #1659, opened 2023-03-23
csr: add ability to express clusters

dalegaard:
This adds the ability to express clusters of registers. This can be very useful for peripherals with repeated groups of registers.

A `CSRCluster` class, emulating a simplified array, is added. Peripherals can add items to the cluster, and the CSRs will, for JSON, CSV, and HTML, be collected as normal.

For SVD, the code is extended to be able to handle outputting clusters directly in the XML structure. This way, code generation tools like `svd2rust` can generate nested array structures, making ergonomics significantly better when working with peripherals.

The code supports nesting of `CSRClusters` too.

An example usecase:
```python
class Sample(Module, AutoCSR):
    def __init__(self):
        self._before = CSRStatus()
        self.subs = CSRCluster()
        for i in range(3):
            self.subs[i] = SampleSub()
        self._after = CSRStatus()

class SampleSub(Module, AutoCSR):
    def __init__(self):
        self._nested = CSRCluster([SampleNested(), SampleNested()])
        self._after_nested = CSRStatus()

class SampleNested(Module, AutoCSR):
    def __init__(self):
        self._field_a = CSRStatus()
        self._field_b = CSRStorage()
        self._field_c = CSRStatus()

class YourSoC(SoCMini):
    def __init__(self):
        self.sample = Sample()
```

I tried to make sure that the touched codebase was as small as possible, to minimize breakage of existing tools. As far as I can tell, everything that doesn't use `CSRCluster` should work as before with no changes.

Best regards,
Lasse

### DIFF
--- a/litex/soc/doc/csr.py
+++ b/litex/soc/doc/csr.py
@@ -45,6 +45,7 @@ class DocumentedCSR:
         size                = 8,
         description         = None,
         access              = "read-write",
+        cluster             = None,
         fields              = []):
 
         self.name                = name
@@ -53,6 +54,7 @@ class DocumentedCSR:
         self.address             = address
         self.offset              = offset
         self.size                = size
+        self.cluster             = cluster
         if size == 0:
             print("!!! Warning: creating CSR of size 0 {}".format(name))
         self.description = self.trim(description)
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @dalegaard, I'll study this. I've planned some work to cleanup and document CSRs, so I'll probably want to do this first and will study your changes while doing this.

dalegaard:
Hi @enjoy-digital 

Sounds great, let me know if you need me to do any changes.

Best regards,
Lasse

PR #1670, opened 2023-04-10
soc/software: fix i2c_write

dasdgw:
commit 3fa3080e introduced an additional address length paramater to i2c_write to support non 8bit memory addresses.
But the parameter wasn't added to the call of i2c_write in i2c_send_init_cmds.

This commit fixes the issue by adding the address length parameter to the call of i2c_write.


enjoy-digital:
Thanks @dasdgw!

PR #1671, opened 2023-04-11
give human readable error messages if a connector or pin is not available

hansfbaier:
Spares the user to have to debug the litex code if a pin or
connector is not found, which happens for example on enclustra boards,
where the HDMI is connected on the baseboard, but not on the KX2 core module.


enjoy-digital:
Thanks @hansfbaier!

PR #1674, opened 2023-04-21
WIP AvalonMM interface and Avalon to Wishbone Bridge

hansfbaier:
Opening this PR as a RFC.
Goal is to implement an AvalonMM Master to Wishbone slave Bridge.

### DIFF
--- a/litex/soc/interconnect/avalon.py
+++ b/litex/soc/interconnect/avalon.py
@@ -146,6 +146,7 @@ class AvalonMM2Wishbone(Module):
         burst_counter    = Signal.like(avl.burstcount)
         burst_address    = Signal.like(avl.readdata)
         burst_read       = Signal()
+        burst_sel        = Signal.like(avl.byteenable)
 
         self.sync += [
              If  (wb.ack | wb.err, read_access.eq(0)) \
@@ -168,7 +169,6 @@ class AvalonMM2Wishbone(Module):
             wb.adr.eq(Mux(burst_cycle & last_burst_cycle,
                           burst_address, avl.address) >> word_width_bits),
             wb.dat_w.eq(avl.writedata),
-            wb.sel.eq(avl.byteenable),
             wb.we.eq(avl.write),
             wb.cyc.eq(read_access | avl.write | burst_cycle),
             wb.stb.eq(read_access | avl.write),
@@ -178,6 +178,7 @@ class AvalonMM2Wishbone(Module):
         self.submodules.fsm = fsm = FSM(reset_state="IDLE")
         fsm.act("IDLE",
             burst_cycle.eq(0),
+            wb.sel.eq(avl.byteenable),
             wb.cti.eq(Mux(avl.burstcount > 1,
                 wishbone.CTI_BURST_INCREMENTING,
                 wishbone.CTI_BURST_NONE)),
@@ -185,6 +186,7 @@ class AvalonMM2Wishbone(Module):
                 burst_cycle.eq(1),
                 NextValue(burst_counter, avl.burstcount - 1),
                 NextValue(burst_address, avl.address + word_width),
+                NextValue(burst_sel, avl.byteenable),
                 If(avl.write, NextState("BURST_WRITE")),
                 If(avl.read,
                     NextValue(burst_read, 1),
@@ -192,6 +194,7 @@ class AvalonMM2Wishbone(Module):
         )
         fsm.act("BURST_WRITE",
             burst_cycle.eq(1),
+            wb.sel.eq(burst_sel),
             wb.cti.eq(Mux(burst_counter > 1,
                 wishbone.CTI_BURST_INCREMENTING,
                 Mux(burst_counter == 1, wishbone.CTI_BURST_END, wishbone.CTI_BURST_NONE))),
@@ -200,11 +203,14 @@ class AvalonMM2Wishbone(Module):
                 NextValue(burst_counter, burst_counter - 1)),
             If(burst_counter == 0,
                 burst_cycle.eq(0),
+                wb.sel.eq(avl.byteenable),
+                NextValue(burst_sel, 0),
                 NextState("IDLE"))
         )
         fsm.act("BURST_READ", # TODO
             burst_cycle.eq(1),
             wb.stb.eq(1),
+            wb.sel.eq(burst_sel),
             wb.cti.eq(Mux(burst_counter > 1,
                 wishbone.CTI_BURST_INCREMENTING,
                 Mux(burst_counter == 1, wishbone.CTI_BURST_END, wishbone.CTI_BURST_NONE))),
@@ -215,6 +221,8 @@ class AvalonMM2Wishbone(Module):
             If (burst_counter == 0,
                 wb.cyc.eq(0),
                 wb.stb.eq(0),
+                wb.sel.eq(avl.byteenable),
+                NextValue(burst_sel, 0),
                 NextValue(burst_read, 0),
                 NextState("IDLE"))
         )
--- a/litex/soc/interconnect/avalon.py
+++ b/litex/soc/interconnect/avalon.py
@@ -27,9 +27,8 @@ _layout = [
 ]
 
 class AvalonMMInterface(Record):
-    def __init__(self, data_width=32, adr_width=30, bursting=False, **kwargs):
+    def __init__(self, data_width=32, adr_width=30, **kwargs):
         self.data_width = data_width
-        self.bursting   = bursting
         if kwargs.get("adr_width", False):
             adr_width = kwargs["adr_width"] - int(log2(data_width//8))
         self.adr_width = adr_width
@@ -127,7 +126,7 @@ class AvalonMMInterface(Record):
 
 class AvalonMM2Wishbone(Module):
     def __init__(self, data_width=32, address_width=32):
-        self.wishbone = wb  = wishbone.Interface(data_width=data_width, adr_width=address_width)
+        self.wishbone = wb  = wishbone.Interface(data_width=data_width, adr_width=address_width, bursting=True)
         self.avalon   = avl = AvalonMMInterface(data_width=data_width, adr_width=address_width)
 
         word_width      = data_width // 8
@@ -188,10 +187,10 @@ class AvalonMM2Wishbone(Module):
             wb.cti.eq(Mux(burst_counter > 1,
                 wishbone.CTI_BURST_INCREMENTING,
                 Mux(burst_counter == 1, wishbone.CTI_BURST_END, wishbone.CTI_BURST_NONE))),
-            If (~avl.waitrequest,
+            If(~avl.waitrequest,
                 NextValue(burst_address, burst_address + word_width),
                 NextValue(burst_counter, burst_counter - 1)),
-            If (burst_counter == 0,
+            If(burst_counter == 0,
                 burst_cycle.eq(0),
                 NextState("IDLE"))
         )
--- a/litex/soc/interconnect/avalon.py
+++ b/litex/soc/interconnect/avalon.py
@@ -138,9 +138,17 @@ class Avalon2Wishbone(Module):
             .Elif(avl.read,        read_access.eq(1)),
         ]
 
+        # Wishbone -> Avalon
         self.comb += [
+            avl.waitrequest.eq(~(wb.ack | wb.err)),
             readdatavalid.eq((wb.ack | wb.err) & read_access),
             readdata.eq(wb.dat_r),
+            avl.readdatavalid.eq(readdatavalid),
+            avl.readdata.eq(readdata),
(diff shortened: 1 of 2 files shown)


litex/soc/interconnect/avalon.py
```diff
@@ -9,6 +9,156 @@
 from migen import *
 
 from litex.soc.interconnect import stream
+from . import wishbone
```
enjoy-digital, line 12: Please use similar imports than other LiteX files (`from litex.soc.interconnect import wishbone`). Relative imports could be fine but I we switch to it, I have a preference to do it for the whole project. We could think about it in the future.

litex/soc/interconnect/avalon.py
```diff
@@ -9,6 +9,156 @@
...
+                else:
+                    r.append(sig.eq(pad))
+            else:
+                if direction == DIR_S_TO_M:
+                    r.append(pad.eq(sig))
+                else:
+                    r.append(sig.eq(pad))
+        return r
+
+    def bus_read(self, address, byteenable=None, burstcount=None, chipselect=None):
```
enjoy-digital, line 73: `bus_read `-> `read` for consistency with Wishbone.

litex/soc/interconnect/avalon.py
```diff
@@ -9,6 +9,156 @@
...
+        yield self.byteenable.eq(0)
+        if burstcount is not None:
+            yield self.burstcount.eq(0)
+        if chipselect is not None:
+            yield self.chipselect.eq(0)
+        while not (yield self.readdatavalid):
+            yield
+        return (yield self.readdata)
+
+    def bus_write(self, address, writedata, byteenable=None, chipselect=None):
```
enjoy-digital, line 99: `bus_write` -> `write` for consistency with Wishbone.

test/test_avalon.py
```diff
@@ -0,0 +1,40 @@
+#
+# This file is part of LiteX.
+#
+# Copyright (c) 2019 Florent Kermarrec <florent@enjoy-digital.fr>
```
enjoy-digital, line 4: I think you wrote it? :)

enjoy-digital (changes requested):
Thanks @hansfbaier! This looks good. I just added a few comment to fix some copyright/cosmetic issues.

Once fixed, we could merge and I could also help testing/improving it.

Have you been able to also test it on hardware on the MiSTeX design that was previously using the verilog AvalonMM to Wishbone bridge?

litex/soc/interconnect/avalon.py
```diff
@@ -9,6 +9,156 @@
...
+    ("byteenable",       "sel_width", DIR_M_TO_S),
+    ("read",                       1, DIR_M_TO_S),
+    ("write",                      1, DIR_M_TO_S),
+    ("waitrequest",                1, DIR_S_TO_M),
+    ("burstbegin",                 1, DIR_M_TO_S), # this is optional
+    ("burstcount",                 8, DIR_M_TO_S),
+    ("chipselect",                 1, DIR_M_TO_S), # this is optional
+]
+
+class MMInterface(Record):
```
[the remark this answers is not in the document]
hansfbaier: OK, I actually did it this way to make it consistent with wishbone.Interface,
so it would be avalon.MMInterface, but I don't have a problem with this naming also.

litex/soc/interconnect/avalon.py
hansfbaier: That does not work, because avalon has  a read signal, which gets in the way.

litex/soc/interconnect/avalon.py
hansfbaier: see above.

test/test_avalon.py
hansfbaier: Yes :) Copy and paste. Will fix.

litex/soc/interconnect/avalon.py
hansfbaier: done

litex/soc/interconnect/avalon.py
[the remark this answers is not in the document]
hansfbaier: eon

litex/soc/interconnect/avalon.py
hansfbaier: Do you have an idea for a more elegant workaround?

hansfbaier:
I have not tested this on hardware yet, and to be able to test this with ascal.vhd, I will need to implement read and write burst support.

litex/soc/interconnect/avalon.py
enjoy-digital: Ah thanks, I see. We can keep `bus_read/write` for now then. I'll think about it. Maybe we should make this separate (also for Wishbone) and make it clear it's for simulation.

enjoy-digital:
BTW, currently Wishbone in LiteX uses word addressing, so you'll probably have to change:
`wb.adr.eq(avl.address)`
to
`wb.adr.eq(avl.address[2:]) (for a 32-bit bus)`

hansfbaier:
Read and write bursts are looking good now
image

hansfbaier:
I think that wishbone_sel should be kept asserted during the duration of the read burst.
Do you agree?

hansfbaier:
Ah yes, it should:
image

hansfbaier:
image

hansfbaier:
Minor refactoring probably would be the state machine being in burst mode for once cycle too long.

enjoy-digital:
Thanks @hansfbaier, good progress. `wishbone_sel `should indeed always be valid when `wishbone_stb` is set.

hansfbaier:
Hmm, I still get a combinatorial loop now, even with `avoid_combinatorial_loop=True`
This is the toplevel file:
https://github.com/MiSTeX-devel/MiSTeX-boards/blob/main/mistex_boards/qmtech_xc7a200t_daughterboard.py
Did I wire the buses wrong somehow?
Vivado reports, these cells are part of the comb loop:
```
serv_rf_top/cpu/bufreg/builder_slave_sel_r[2]_i_1
serv_rf_top/cpu/bufreg/builder_slave_sel_r[2]_i_2
serv_rf_top/cpu/bufreg/builder_slave_sel_r[3]_i_3
serv_rf_top/cpu/bufreg/builder_slave_sel_r[3]_i_4
serv_rf_top/cpu/decode/builder_socbushandler_dat_r[63]_i_4
serv_rf_top/cpu/bufreg/builder_socbushandler_dat_r[63]_i_8
serv_rf_top/cpu/bufreg/builder_socbushandler_dat_r[63]_i_11
serv_rf_top/cpu/bufreg/main_aborted_i_2
serv_rf_top/cpu/bufreg/rreq_r_i_3
serv_rf_top/cpu/bufreg/rreq_r_i_4
serv_rf_top/cpu/bufreg/rreq_r_i_5
serv_rf_top/cpu/bufreg/sram_reg_i_82
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_1
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_2
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_3
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_4
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_5
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_12_17_i_6
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_18_21_i_1
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_18_21_i_2
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_18_21_i_3
serv_rf_top/cpu/bufreg/storage_9_reg_0_7_18_21_i_4
```
Would that mean I have to register sel and dat_r too?

hansfbaier:
Update: With the latest push I got it down to:
```
serv_rf_top/cpu/bufreg/builder_slave_sel_r[2]_i_1
serv_rf_top/cpu/bufreg/builder_slave_sel_r[2]_i_2
serv_rf_top/cpu/bufreg/builder_slave_sel_r[3]_i_2
serv_rf_top/cpu/bufreg/builder_slave_sel_r[3]_i_4
serv_rf_top/cpu/state/builder_socbushandler_dat_r[63]_i_4
serv_rf_top/cpu/bufreg/main_aborted_i_2
serv_rf_top/cpu/bufreg/rreq_r_i_3
serv_rf_top/cpu/bufreg/rreq_r_i_4
serv_rf_top/cpu/bufreg/rreq_r_i_5.
```

hansfbaier:
Weird, as soon as I switched the data buswidth from 64 to 128, all my troubles with combinatorial loops went away.

hansfbaier:
It compiles fine now, and the DDR3 init goes well in the LiteX BIOS,
but still have no video out from ascal.
Could you maybe look at my toplevel file to see if I got the address calculations wrong?
https://github.com/MiSTeX-devel/MiSTeX-boards/blob/main/mistex_boards/qmtech_xc7a200t_daughterboard.py

enjoy-digital:
@hansfbaier: This could be useful to test this in simulation to easily spot issues, this one one of the aim of creating a simulation in `litex_mister_test`. On my side, I was planning to try to integrate ASCAL like this: https://github.com/enjoy-digital/litex_mister_test/issues/3. For simulation, this also means that GHDL should correctly convert it to Verilog (which should be the case).

enjoy-digital:
Have you also try to add a LiteScope on the AvalonMM interface? Maybe not the full bus, but at least control signals and addresses, just to see the activity. And also add some control signals of the MiSTeR core to LiteScope to be able to trigger in it (ex reset).

hansfbaier:
@enjoy-digital Are all wishbone components in the LiteX SoC burst capable (Arbiter, Up/Downconverters)?

hansfbaier:
> Have you also try to add a LiteScope on the AvalonMM interface? Maybe not the full bus, but at least control signals and addresses, just to see the activity. And also add some control signals of the MiSTeR core to LiteScope to be able to trigger in it (ex reset).

Not yet. I still fretting over how to connect with Litescope on my board. Maybe use the ethernet on the QMTech daughterboard.

hansfbaier:
@enjoy-digital What is the maximum wishbone burst length supported by LiteDRAM?

hansfbaier:
@enjoy-digital Hmm, I cannot find any burst support in LiteDRAMWishbone2Native. Then I am not surprised that ascal does not work.

enjoy-digital:
@hansfbaier: Indeed, that's why I was planing to use direct LiteDRAM interface. But I would still recommend trying to work on these aspects with a system level simulation: LiteX Sim integrates LiteDRAM, so this will be convenient test ASCAL DRAM accesses.

hansfbaier:
@enjoy-digital OK I am just now writing a sim.py for ascal, to avoid duplicate effort.

enjoy-digital:
Great, such sim will be very useful for MiSTeX in the long term.

jevinskie:
> > Have you also try to add a LiteScope on the AvalonMM interface? Maybe not the full bus, but at least control signals and addresses, just to see the activity. And also add some control signals of the MiSTeR core to LiteScope to be able to trigger in it (ex reset).
> 
> Not yet. I still fretting over how to connect with Litescope on my board. Maybe use the ethernet on the QMTech daughterboard.

EtherBone would be the fastest, JTAGBone would be in the middle with pretty good speed (at least using a legitimate USB Blaster II or one built into the board), and a UARTBone works but might be a bit slow for your taste. The nice thing about JTAG/UARTbone is that they pretty much Just Work and are the two bridge transports that I get running on a new board since they're so easy to implement.

enjoy-digital:
Thanks @hansfbaier, I merged it and will play with it/will probably dome some minor changes. Thanks for the testbench, this will be very useful for this and later to catch regressions!

hansfbaier:
Thanks

enjoy-digital:
@hansfbaier: I did a first couple of changes. Can you tell more about the avoid_combinatorial_loop? If we can understand/avoid it, it would probably be possible to do other simplifications. If you can share the case that was causing trouble on your end, I could have a look. Could it also be related to `minimumResponseLatency` from the Avalon specification (P35 of http://www.audentia-gestion.fr/INTEL/PDF/mnl_avalon_spec.pdf). I think it should be possible to minimize the changes between different  `minimumResponseLatency`  to just adding a register on `readdata`/`readdatavalid`.

hansfbaier:
@enjoy-digital It has more to do with wishbone, because ack can be wired combinatorially so that for some memory types ack can come back on the same clock cycle, which would be preferred.

hansfbaier:
The tests should work with both the setting on and off.

enjoy-digital:
@hansfbaier: OK thanks, I see. I'll have another look at the code with this in mind.

PR #1677, opened 2023-04-25
bios/spiflash: fix write/ erase, add write from sdcard and range erase

mntmn:
When shipping MNT RKX7, I pre-flash the SPI flash with a LiteX bitfile
for testing. cmd_spiflash had regressed because of changed SPIFLASH defines
and didn't offer the write functions anymore. This commit fixes that, and
adds convenience functions:

- `flash_erase_range <offset> <count (bytes)>`
- `flash_from_sdcard <filename>`

The latter reuses some boot code to copy the contents of the specified
file from the boot FAT partition on the SD card to SPI flash (i.e.
a bitstream).

Some of the code is based on @norbertthiel's comment https://github.com/litex-hub/litespi/issues/52
But edited to work with W25Q128JVS flash used in MNT RKX7.

### DIFF
--- a/litex/soc/software/bios/cmds/cmd_spiflash.c
+++ b/litex/soc/software/bios/cmds/cmd_spiflash.c
@@ -8,23 +8,26 @@
 #include "../command.h"
 #include "../helpers.h"
 
+#include <libbase/progress.h>
+#include <liblitespi/spiflash.h>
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @mntmn, nice features contribution! This looks good so let's merge.

PR #1683, opened 2023-05-09
AvalonMM2Wishbone: use same addressing on avalon and wishbone, leave address translation to the user

hansfbaier:
As we discussed:
image


enjoy-digital:
Thanks @hansfbaier!

PR #1684, opened 2023-05-09
Add low res video modes

hansfbaier:
for low spec/embedded systems.


enjoy-digital:
Thanks @hansfbaier, note that for exotic video timings, it's also possible to provide them directly in the target file, for example: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/siglent_sds1104xe.py#L148-L164

hansfbaier:
Yes, thanks I used that in the ascal sim.

PR #1685, opened 2023-05-09
Assert readdatavalid on bursts in Avalon2WishboneMM test (fixes #1686)

### LINKED ISSUE #1686: AvalonMM2Wishbone: removing assert of readdatavalid breaks burst reads
This fix actually breaks burst reads:
image
In the avalon spec, readdata valid needs to be asserted whenever data is returned:
image
I also validated this working with ascal in Verilator.
What is the problem here?

hansfbaier:


enjoy-digital:
Thaknks @hansfbaier!

PR #1689, opened 2023-05-11
Avalon2Wishbone: Burst can only advance if write is high and waitrequest low

hansfbaier:
image
I just realized this was missing


enjoy-digital:
Thanks @hansfbaier.

PR #1690, opened 2023-05-15
add an option to generate without reg initializers (asic targets)

bunnie:
ASIC targets can't set a reg to a known value on boot, so for more accurate simulations it would be nice to have an option in the platform to specify generating the verilog without 'reg' initializers. The presence of these initializers can mask problems in simulations with X-prop that can lead to missing explicit reset conditions.

The PR is coded so this should have no effect on users who don't need this feature.

thanks!

### DIFF
--- a/litex/gen/fhdl/verilog.py
+++ b/litex/gen/fhdl/verilog.py
@@ -421,7 +421,7 @@ def _print_module(f, ios, name, ns, attr_translate):
 
     return r
 
-def _print_signals(f, ios, name, ns, attr_translate):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @bunnie, I perfectly understand the need for this. I'm going to merge and will just modify a little bit the changes and rename the parameter.

enjoy-digital:
The rework: https://github.com/enjoy-digital/litex/commit/9c890a0a27b60a7b38afe70a0cbfb9570acaf681

PR #1691, opened 2023-05-15
Clock keep optional for XilinxPlatform

jersey99:
Make keep attribute for add_period_constraint optional. This is because Vivado 2019 fails to generate a correct synthesis, when it tries to replicate IDELAYCTRL automatically, and the refclk to the original IDELAYCTRL is set to dont_touch.


enjoy-digital:
Thanks @jersey99.

PR #1692, opened 2023-05-16
soc/interconnect/packet: Don’t bypass dispatcher with a single slave if it can be deselected.

zyp:
A dispatcher using one hot encoding can deselect all slaves, meaning that with a single slave we effectively get one select bit that decides whether to pass or drop a packet. For this reason, a single slave dispatcher can still be useful, and we shouldn't bypass the dispatch logic when using one hot encoding.

This allows e.g. `LiteEthCrossbar` to still filter traffic when only a single port is instanced.


enjoy-digital:
Thanks @zyp!

PR #1693, opened 2023-05-17
VideoS7HDMIPhy: add option to flip the polarity of the differential pairs

hansfbaier:
I have the HDMI Pmod of Muselab, but its diff pairs  polarity
is the opposite of the FPGA pins polarity,
so the only fix is to invert the differential pair signal.


hansfbaier:
Oh that won't work, because Vivado can't route it.
Closed.

PR #1699, opened 2023-05-27
Fix frequency specification for ECPDAP on Lattice

bjonnh:
It was given as kHz but it takes Hz.

It makes the programmer fail with default speeds.


enjoy-digital:
Thanks @bjonnh.

Issue #1596, opened 2023-02-10
Radiant 3.2 tcl script error - "@E| Mal-formed command line - please check for extra quotes in macro"


slagernate:
Radiant 3.2.0.18.0 and newer spits out: "@E| Mal-formed command line - please check for extra quotes in macro"

This can be fixed by removing the `prj_set_impl_opt {include path} {""}` line ini the tcl script 

OR

commenting out line 155 in litex/litex/build/lattice/radiant.py: 
```
#tcl.append("prj_set_impl_opt {include path} {\"" + vincpath + "\")")
```

Probably a simple fix like calling `append()` only if `vincpath` is not `none` would do.

PR #1718, opened 2023-06-23
fix radiant bug 'Mal-formed command line - please check for extra quotes in macro'

slagernate:
See #1596.

### DIFF
--- a/litex/build/lattice/radiant.py
+++ b/litex/build/lattice/radiant.py
@@ -156,8 +156,6 @@ class LatticeRadiantToolchain(GenericToolchain):
         if vincpath and vincpath.strip():
             tcl.append("prj_set_impl_opt {include path} {\"" + vincpath + "\"}")
 
-        tcl.append("prj_set_impl_opt {include path} {" + vincpath + "}")
-
         # Add sources
         if self._synth_mode == "yosys":
             # NOTE: it is seemingly impossible to skip synthesis using the Tcl flow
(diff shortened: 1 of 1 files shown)


litex/build/lattice/radiant.py
```diff
@@ -153,7 +153,9 @@ def tcl_path(path): return path.replace("\\", "/")
 
         # Add include paths
         vincpath = ";".join(map(lambda x: tcl_path(x), self.platform.verilog_include_paths))
-        tcl.append("prj_set_impl_opt {include path} {\"" + vincpath + "\"}")
+        if vincpath != "": # Do not add quotes if vincpath is empty, else Radiant complains
```
trabucayre, line 156: An empty string is seen as `False` so:
````
if vincpath and vincpath.strip():
   [...]
```
is more pythoniic and catch empty string and string with only blank.

litex/build/lattice/radiant.py
```diff
@@ -153,7 +153,9 @@ def tcl_path(path): return path.replace("\\", "/")
...
+            vincpath = "\"" + vincpath + "\"" 
+        tcl.append("prj_set_impl_opt {include path} {" + vincpath + "}")
```
trabucayre, line 158: is `include path` is a mandatory option? ie when `vincpath` is empty this line may be skipped?

slagernate: Seems to build fine without `include path`

slagernate: @trabucayre see latest commit. Happy to make further changes / take feedback

trabucayre: LGTM, but l.159 must be removed.

Thanks

slagernate: Whoops. Thanks for spotting that.

trabucayre:
Applied. Thanks @slagernate !

PR #1720, opened 2023-06-28
Tx write only

sensille:
Allow liteeth txslots to be write only. Enable it in case the firmware does not need to read back the buffer. This can save a substantial amount of resources. Tested on Spartan6. Without it, txbuffers are implemented as distributed RAM, probably due to the 2 read ports. With this patch, it is implemented as BRAM.
Needs to be pulled in conjunction with the corresponding litex patch.


enjoy-digital:
Thanks @sensille!

enjoy-digital:
Applied, I just did some minor changes with https://github.com/enjoy-digital/litex/commit/7fa7a4c72a994955637cf13ac760c32e0845604a.

PR #1724, opened 2023-07-05
cores/zynqmp: Fix boot helper

stone3311:
This pull request fixes a regression where the computed goto causes an error when compiling the boot helper with newer versions of GCC. The behavior of goto has changed since GCC 12 (see https://gcc.gnu.org/gcc-12/porting_to.html).


trabucayre:
Thanks for this fix. Applied

PR #1728, opened 2023-07-09
cores/arm: Fix computed goto in boot helpers

stone3311:
This pull request fixes the computed goto of all ARM boot helpers. The issue is identical to #1724.


enjoy-digital:
Thanks @stone3311!

PR #1729, opened 2023-07-11
Currently using a lite or minimal Vexriscv config with debug and brea…

riktw:
Currently using a lite or minimal Vexriscv config with debug and breakpoints throws an error. Updated the GCC_FLAGS to include these two variants as well.


Dolu1990:
Thanks ^^

PR #1733, opened 2023-07-24
Revert "soc/interconnect/packet: Don’t bypass dispatcher with a singl…

rowanG077:
…e slave if it can be deselected."

This reverts commit 91f56aaf0ec9d5f15193fa817ba6024e76085531.

I was updating litex in my project and the udpraw core stopped working while still responding to pings. After bisecting I found that this is the offending commit.

I'm not sure whether the udpraw core code is wrong or this is wrong... For now I have reverted this locally. If this is not the correct way forward please let me know as well.

### DIFF
--- a/litex/soc/interconnect/packet.py
+++ b/litex/soc/interconnect/packet.py
@@ -62,7 +62,7 @@ class Dispatcher(Module):
     def __init__(self, master, slaves, one_hot=False):
         if len(slaves) == 0:
             self.sel = Signal()
-        elif len(slaves) == 1 and not one_hot:
+        elif len(slaves) == 1:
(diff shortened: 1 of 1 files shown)


rowanG077:
Looking a bit deeper. The reason it goes wrong is that on the udpraw core I don't use `sink_dst_port` as a filter. I assume I receive all packets and then set the `sink_dst_port` only when I have an outgoing packet. It's kind of silly to use a signal which is technically only valid when transmitting a packet as the selector on the crossbar. 

In fact if anything it should use `sink_src_port` since that is the port you are indicating is used for receiving. I will close this PR in favor of fixing the udpraw core in liteeth.

PR #1735, opened 2023-07-26
Adding a weak isr handler to handle non uart interrupts

riktw:
This would add a weak isr_handler that can be overridden in a user's application to handle interrupts that are not for the build in uart. This would make it possible to use the libbase for your own application and add peripherals with an irq.

For example, in an own application something like this can be added to handle an interrupt from an additional timer:

```
unsigned int isr_handler(int irqs)
{	
	unsigned int irqHandled = 0;
	if(irqs & (1 << TIMER2_INTERRUPT))
	{
		Timer2ClearPendingInterrupt();
		// Some own function to act on a timer interrupt
		irqHandled = 1;
	}
	return irqHandled;
}
```


hvegh:
Would like to see something like this to be added, please have a look. 

This closes https://github.com/enjoy-digital/litex/issues/1807

Tnx

Issue #1624, opened 2023-02-28
Ubuntu 22.04 riscv gcc causes issue


AEW2015:
https://github.com/enjoy-digital/litex/blob/2f5481dbb9ff22cff03f21e8e7d418442b778908/litex_setup.py#L353

```
/usr/lib/gcc-cross/riscv64-linux-gnu/11/../../../../riscv64-linux-gnu/bin/ld: ../libc/libc.a(libc_ssp_chk_fail.c.o): in function `__chk_fail':
/home/deck/xup/litex_repos/build/opalkelly_xem8320/software/libc/../../../../pythondata-software-picolibc/pythondata_software_picolibc/data/newlib/libc/ssp/chk_fail.c:13: undefined reference to `write'
collect2: error: ld returned 1 exit status
make: *** [/home/deck/xup/litex_repos/litex/litex/soc/software/bios/Makefile:72: bios.elf] Error 1
rm crt0.o
make: Leaving directory '/home/deck/xup/litex_repos/build/opalkelly_xem8320/software/bios'
Traceback (most recent call last):
  File "/home/deck/xup/litex_repos/./litex-boards/litex_boards/targets/opalkelly_xem8320.py", line 158, in <module>
    main()
  File "/home/deck/xup/litex_repos/./litex-boards/litex_boards/targets/opalkelly_xem8320.py", line 151, in main
    builder.build()
  File "/home/deck/xup/litex_repos/litex/litex/soc/integration/builder.py", line 354, in build
    self._generate_rom_software(compile_bios=use_bios)
  File "/home/deck/xup/litex_repos/litex/litex/soc/integration/builder.py", line 288, in _generate_rom_software
    subprocess.check_call(["make", "-C", dst_dir, "-f", makefile])
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['make', '-C', '/home/deck/xup/litex_repos/build/opalkelly_xem8320/software/bios', '-f', '/home/deck/xup/litex_repos/litex/litex/soc/software/bios/Makefile']' returned non-zero exit status 2.
```

I am using the https://static.dev.sifive.com/dev-tools/freedom-tools/v2020.08/riscv64-unknown-elf-gcc-10.1.0-2020.08.2-x86_64-linux-ubuntu14.tar.gz instead.

n-kremeris:
[...]

Issue #1621, opened 2023-02-28
Default bios not compiling for tang primer 20k,


L-jr-Bennett:
Error when using --with-sdcard or --with-ethernet for default target for Tang Primer 20k board.
`pythondata-software-picolibc/pythondata_software_picolibc/data/newlib/libc/ssp/chk_fail.c:13: undefined reference to `write'
collect2: error: ld returned 1 exit status `

enjoy-digital:
I'm not able to reproduce, the issue seems to be related to picolibc compilation. Can you make sure you are using up to date version of the dependencies (`./litex_setup.py --update`) and provide more info on your machine?

bentwire:
I'm experiencing this as well.  From a fresh install:
Commands I ran to install:
./litex_setup.py --init --install --user --config full
./litex_setup.py --update
pushd litex-boards/litex_boards/targets/
./sipeed_tang_primer_20k.py --dock standard --with-sdcard --with-video-terminal --with-ethernet --build

It fails here:
`CC       bios.elf
/usr/lib/gcc-cross/riscv64-linux-gnu/11/../../../../riscv64-linux-gnu/bin/ld: ../libc/libc.a(libc_ssp_chk_fail.c.o): in function __chk_fail:
/home/mike/gowin/projects/litex/litex-boards/litex_boards/targets/build/sipeed_tang_primer_20k/software/libc/../../../../../../../pythondata-software-picolibc/pythondata_software_picolibc/data/newlib/libc/ssp/chk_fail.c:13: undefined reference to write
collect2: error: ld returned 1 exit status
make: *** [/home/mike/gowin/projects/litex/litex/litex/soc/software/bios/Makefile:72: bios.elf] Error 1
rm crt0.o
make: Leaving directory '/home/mike/gowin/projects/litex/litex-boards/litex_boards/targets/build/sipeed_tang_primer_20k/software/bios
Traceback (most recent call last):
  File "/home/mike/gowin/projects/litex/litex-boards/litex_boards/targets/./sipeed_tang_primer_20k.py", line 257, in <module>
    main()
  File "/home/mike/gowin/projects/litex/litex-boards/litex_boards/targets/./sipeed_tang_primer_20k.py", line 246, in main
    builder.build(**parser.toolchain_argdict)
  File "/home/mike/gowin/projects/litex/litex/litex/soc/integration/builder.py", line 354, in build
    self._generate_rom_software(compile_bios=use_bios)
[...]

PR #1738, opened 2023-07-30
litex_setup: fix software build when liteeth or/and litesata is set (riscv toolchain issue)

trabucayre:
Starting with ubuntu 22.04, when a target is configured with ethernet and/or sata the sofware build fails with:
```
/usr/lib/gcc-cross/riscv64-linux-gnu/12/../../../../riscv64-linux-gnu/bin/ld: ../libc/libc.a(libc_ssp_chk_fail.c.o): in function `__chk_fail':
/somewhere/litex-boards/litex_boards/targets/build/digilent_arty/software/libc/../../../../../../../pythondata-software-picolibc/pythondata_software_picolibc/data/newlib/libc/ssp/chk_fail.c:13: undefined reference to `write'
collect2: error: ld returned 1 exit status
make: *** [/somewhere/litex/litex/soc/software/bios/Makefile:72: bios.elf] Error 1
```

By switching to *gcc-riscv64-unknown-elf* (bare-metal toolchain) build end successfully
Fix:
- https://github.com/enjoy-digital/litex/issues/1624
- https://github.com/enjoy-digital/litex/issues/1621

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -350,7 +350,7 @@ def riscv_gcc_install():
             os.system("pacman -S riscv64-linux-gnu-gcc")
         # Ubuntu.
         else:
-            os.system("apt install gcc-riscv64-linux-gnu")
+            os.system("apt install gcc-riscv64-unknown-elf")
 
     # Mac OS.
     # -------


enjoy-digital:
Thanks @trabucayre!

jevinskie:
Just FYI if we wanted to consolidate on the same RISC-V toolchain for every host we might be able to use [these prebuilds](https://xpack.github.io/dev-tools/riscv-none-elf-gcc/). After looking around for a bit this packaging project seems to be the most similar to other high-quality “universal” crosstool builds like Linaro’s Arm cross-compiler releases.

PR #1740, opened 2023-08-02
Add documentation arguments to builder

bayi:
# New arguments for documentation generation

This PR adds the following new arguments to builder:

 - doc-project= String: The project name that should be displayed in the documentation. Defaults to the soc name.
 - doc-author= String: The author of the documentation. Defaults to `LiteX`
 - doc-extra-config= String: the extra config options that should be included in the generated sphinx conf. ( Like setting a different theme )

The argument member property names have underscores instead of dashes in their names.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -411,7 +411,7 @@ def builder_args(parser):
     builder_group.add_argument("--soc-svd", "--csr-svd",  default=None,        help="Write SoC mapping to the specified SVD file.")
     builder_group.add_argument("--memory-x",              default=None,        help="Write SoC Memory Regions to the specified Memory-X file.")
     builder_group.add_argument("--doc",                   action="store_true", help="Generate SoC Documentation.")
-    builder_group.add_argument('--doc-project-name',      default=None,        help="Project name for SoC Documentation.")
+    builder_group.add_argument('--doc-project',           default=None,        help="Project name for SoC Documentation.")
     builder_group.add_argument('--doc-author',            default=None,        help="Author for SoC Documentation.")
     builder_group.add_argument('--doc-extra-config',      default=None,        help="Extra configuration for SoC Documentation.")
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @bayi, this is interesting. Some work has been planned on documentation in September, I'll probably wait to start this to also closely review/merge the PR.

bayi:
Well the reasoning behind my changes are:

I wanted to use a different theme for the sphinx docs, so installed the rtd theme and i had to actually copy the `interconnect/builder.py` to my project just give it the `html_theme = 'sphinx_rtd_theme'` for the extra config.

Now i can do that with args from my soc definition.

Im using it like this:

kép

Which ends up with a nice documentation generated like this:

kép

Also while we are here, do you know how can i make the CSRField names fit the WaveDrom graph if they are longer ? :)

enjoy-digital:
Thanks, the use case makes sense, especially for changing the Sphinx theme or project metadata.

I would prefer not to merge this as-is though, since it adds three new generic Builder CLI options for a fairly specialized documentation customization path. In particular `--doc-extra-config` would be raw Sphinx/Python config on the command line, which feels a bit too open-ended for the common target CLI.

The current `generate_docs()` API already supports this kind of customization directly:

  ```python
  generate_docs(
      soc,
      "build/documentation",
      project_name="My SoC",
      author="...",
      sphinx_extra_config="""
  html_theme = "sphinx_rtd_theme"
  """,
  )
  ```

It also supports from_scratch=False for users who want to keep/customize their own conf.py.

So I think the best compromise is to keep Builder --doc simple and use the Python generate_docs() API for advanced/custom documentation generation. If this becomes a common need from targets using Builder, we could later add a single programmatic Builder hook for doc kwargs rather than several new global CLI options.

So for now I suggest closing this PR.

bayi:
Thanks and i agree in this case, i will use that method instead.

PR #1742, opened 2023-08-11
build/gowin/common: Fix DDRInput

Icenowy:
The DDRInput of Gowin seems to be never used and contains a typo that prevents it from being really used.

Fix this typo.


trabucayre:
Thanks @Icenowy

PR #1743, opened 2023-08-13
gowin mcufw option

noiuynoise:
This PR adds an `mcu_firmware` option to the OpenFPGALoader `load_bitstream()` function. Adding a string to this field adds it as a `--mcufw` flag which loads it into the internal user flash of Gowin FPGAs.

Sadly this flag has to be set when the bitstream is loaded so it can't go into the `flash()` function and follow the load command pattern used for other boards.

### DIFF
--- a/litex/build/openfpgaloader.py
+++ b/litex/build/openfpgaloader.py
@@ -25,8 +25,10 @@ class OpenFPGALoader(GenericProgrammer):
         if index_chain is not None:
             self.cmd += ["--index-chain", str(int(index_chain))]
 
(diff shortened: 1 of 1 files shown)


trabucayre:
Hi @noiuynoise: This feature is now addressed by 3909b1d6117231d420ddeeab12567c96ba5cec79. The approach is more easy because possible openFPGALoader's options may be accessible without having a too large/complex method's signature.
See, also, https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/sipeed_tang_nano_4k.py#L179 for an usage example

PR #1744, opened 2023-08-15
Workarounds for utilizing Ethernet on Gowin

Icenowy:


nickoe:
Hi @Icenowy 

How much did you get ethernet to work with this? I also recently acquired a tang primer 20k with standard dock, so started to play a bit around with it.

What is the distinction between a port and a net in this case? I mean, to me it just looks like you renamed the clocks. Also the `add_false_path_constraint` is that called anywhere? Or is that in some related patches?

Icenowy:
Seems to be not valid enough, close now.

PR #1748, opened 2023-08-18
Generate DTS with XIP support for running Zephyr from flash

josuah:
This incidentally remove the `-DCONFIG_SPI_LITESPI=y` flag altogether, which I can add back in another commit to this pull request. Should there be some extra command line flag to choose if the SPI flash should be used for XIP or for user data storage like a filesystem? Or support both by default/

No support for partitions yet: this does not yet support cohabitating the bitstream and firmware yet, so the bitstream has to be loaded to CRAM directly for now (default for --load).

This allows loading larger Zephyr examples, and was required to fit the shell on the Crosslink-NX board.

See https://github.com/litex-hub/zephyr-on-litex-vexriscv/pull/13

### DIFF
--- a/litex/tools/litex_json2dts_zephyr.py
+++ b/litex/tools/litex_json2dts_zephyr.py
@@ -208,9 +208,9 @@ overlay_handlers = {
         'config_entry': 'ETH_LITEETH'
     },
     'spiflash': {
-        'handler': peripheral_handler,
-        'alias': 'spi0',
-        'config_entry': 'SPI_LITESPI'
+        'handler': ram_handler,
+        'alias': 'flash0',
+        'config_entry': 'XIP'
     },
     'sdcard_block2mem': {
         'handler': peripheral_handler,


josuah:
I notice that this can be configured from the Zephyr application and LiteX SoC entirely.
The XIP support becomes completely optional.

It is a matter of choice whether we want `spiflash` to always be used for Zephyr (implying `CONFIG_XIP`).

Also, it is necessary to overwrite that setup if we want to use a different flash address anyway, such as including an offset for the bitstream.

I am therefore closing this, as the most common use-case requires customization anyway.

josuah:
## Zephyr-side:

### `app.overlay`:
```
/ {
    chosen {
        zephyr,flash = &flash0;
    };

    # This is the flash address space  + 0x0010_0000 to have Zephyr after the bitstream
    # Zephyr's linker script will follow this nicely.
    flash0: flash@20100000 {
        reg = <0x20100000 0x100000>;
        compatible = "soc-nv-flash";
    };
};
```

## `prj.conf`:
```
CONFIG_XIP=y
```

## `CMakeLists.txt`:
```
cmake_minimum_required(VERSION 3.20.0)

set(CONF_FILE prj.conf litex.conf)
set(DTC_OVERLAY_FILE app.overlay litex.overlay)
set(BOARD litex_vexriscv)

find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
project(example_litex_project)
add_subdirectory(drivers)

target_sources(app PRIVATE src/main.c)
```

## LiteX-side:

Add the SPI flash, in whichever way work for you, I went down the route of specifying `soc.spi_flash_module = ...` from the board, and use it from `add_spi_flash()` on ZephyrSoC, to allow per-board customization reflect on ZephyrSoC:

https://github.com/josuah/zephyr-on-litex-vexriscv/blob/master/make.py#L217-L218
https://github.com/josuah/zephyr-on-litex-vexriscv/blob/master/soc_zephyr.py#L87-L89

PR #1755, opened 2023-08-31
build/xx/toolchains: allows override clock naming

trabucayre:
This PR allows to specify explicitly a clock name instead of using net/port name.
Note: most of toolchains uses a really similar method to produces sdc -> a refactoring may make sense


enjoy-digital:
Thanks @trabucayre, this looks good.

PR #1758, opened 2023-09-01
soc/cores/spi_mmap: Fix clock divider

AndrewD:
This change ensures clk_divider represents the true divider. Alternately could test: clk_count == self.clk_divider and document that the actual divide is 2x clk_divider.


enjoy-digital:
Thanks @AndrewD.

PR #1763, opened 2023-09-02
Allow use of Windows-side (.exe) Radiant toolchain

josuah:
Under WSL, it is possible to execute `.exe` files.

This patch allows to use Windows-side Radiant from WSL, so that Radiant does not need to be installed inside WSL if it already there outside.

I got feedback that this patch works from who tested it.


enjoy-digital:
Thanks @josuah!

PR #1764, opened 2023-09-03
[RFC] litex_sim: move most code from main() to SimSoC()

josuah:
I was on my way to bring support for the simulator on [Zephyr-on-LitexVexriscv](https://github.com/litex-hub/zephyr-on-litex-vexriscv), then I saw this: https://github.com/litex-hub/linux-on-litex-vexriscv/blob/master/sim.py#L70

This means that currently, `./make.py` has nothing to do with `./sim.py`, and any pull request to `./make.py` needs maintainers to always update `./sim.py`... time-consuming for little added-value?

So why not allow the simulator to be configured as a regular boards? `import litex.tools.litex_sim`

Supersedes https://github.com/litex-hub/litex-boards/pull/525

### DIFF
--- a/litex/tools/litex_sim.py
+++ b/litex/tools/litex_sim.py
@@ -153,24 +153,31 @@ class SimSoC(SoCCore):
         with_sdram            = False,
         with_sdram_bist       = False,
         with_ethernet         = False,
+        local_ip              = None,
+        remote_ip             = None,
         ethernet_phy_model    = "sim",
         with_etherbone        = False,
         etherbone_mac_address = 0x10e2d5000001,
         etherbone_ip_address  = "192.168.1.51",
         with_analyzer         = False,
         sdram_module          = "MT48LC16M16",
-        sdram_init            = [],
+        sdram_init            = None,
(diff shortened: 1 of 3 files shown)


litex/tools/litex_sim.py
```diff
@@ -545,7 +542,7 @@ def pre_run_callback(vns):
 
     builder = Builder(soc, **parser.builder_argdict)
     builder.build(
-        sim_config       = sim_config,
+        sim_config       = soc.sim_config,
```
josuah, line 545: Because `sim_config` needs to be passed to the builder but is not in `main()` anymore, `SimSoC()` exposes it as `self.sim_config`.

josuah:
Required by https://github.com/litex-hub/zephyr-on-litex-vexriscv/pull/18

I can change `#18` so that it does not rely on it and close this if there were other plans for it.

josuah:
I am closing this to let Enjoy-Digital perform refactoring freely on the Zephyr-on-LiteX-Vexriscv, as well as Linux-on-LiteX-Vexriscv, which will might happen everywhere else before here.

PR #1767, opened 2023-09-04
build/lattice/radiant: fix uname() not prefixed by 'os.'

josuah:
My bad for breaking it in https://github.com/enjoy-digital/litex/pull/1763


enjoy-digital:
Thanks @josuah, that's merged.

PR #1768, opened 2023-09-05
[RFC] new tool to produce a Verilog export

josuah:
Add an ExportToolchain, and ExportPlatform almost empty, only there to generate the Verilog output, to integrate manually.

In addition, a command line tool, `litex_export_verilog` is added to trigger a build that only produces the BIOS and Verilog.

Addresses https://github.com/enjoy-digital/litex/wiki/Export-Your-Core-SoC-To-Verilog

### DIFF
--- a/litex/build/export/platform.py
+++ b/litex/build/export/platform.py
@@ -1,0 +1,40 @@
+#
+# This file is part of LiteX.
+#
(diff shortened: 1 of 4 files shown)


josuah:
A hint from IRC discussion suggests that this kind of goes against the whole idea of LiteX.

Well... 

If trying to bypass LiteX

josuah:
I u

PR #1769, opened 2023-09-05
clock/gowin_gw5a: change allowed frequency range for GW5A- prefix

Icenowy:
When targeting GW5A-25 ES, the Gowin IDE has a more strict frequency range.

Change the range when GW5A- is matched to this.


enjoy-digital:
Thanks @Icenowy, that's merged.

PR #1775, opened 2023-09-13
gen/genlib/misc: Add CE option to WaitTimer

rowanG077:
One nice thing this allows is pipelining and combining of multiple timers. For example imagine you need to have multiple timers that are relatively long. One option now is to create a base timer and then on every base timer `done` use that has the `CE` of the other timers.

This is now possible:

```Python
self.timer1 = timer1 = WaitTimer(10)
self.timer2 = timer2 = WaitTimer(5, has_ce=True)
self.timer3 = timer3 = WaitTimer(3, has_ce=True)

self.comb += [
    self.timer1.wait.eq(~self.timer1.done),
    self.timer2.wait.eq(~self.timer2.done),
    self.timer3.wait.eq(~self.timer3.done),
]

self.sync += [
    self.timer2.ce.eq(self.timer1.done),
    self.timer3.ce.eq(self.timer1.done),
]
```

`timer2` will now `done` every 5*10 cycles and `timer3` is now done every `3*10` cycles


rowanG077:
I was not aware that existed.

PR #1778, opened 2023-09-14
gen/fhdl/verilog: Fix #1777.

### LINKED ISSUE #1777: Combinatorial signal with single slice assignment loses default ("reset") value
If a multi-bit signal is created and some bits of it is assigned _in a single assignment_, then the specified `reset` value is lost.
Consider the following module:
```
class mod1(Module):
    def __init__(self):
        self.x = Signal(8, reset=17, name_override="x")
        self.comb += [
            self.x[6:8].eq(0b01)
        ]
```
This results in the verilog
```
wire    [7:0] x;
assign x[7:6] = 1'd1;
```
Only the top two bits are assigned, and the reset value for the rest of the bits is lost.  The resulting vector will have `x` (or `z` in Icarus) in the lower 6 bits.  Not what I wanted.
_However_, if I use two assignments to set the top bits individually, like so:
```
class mod2(Module):
    def __init__(self):
        self.y = Signal(8, reset=17, name_override="y")
        self.comb += [
            self.y[6].eq(1),
            self.y[7].eq(0)
        ]
```
then the verilog becomes
```
reg     [7:0] y = 8'd17;
always @(*) begin
    y <= 8'd17;
    y[6] <= 1'd1;
    y[7] <= 1'd0;
end
```
and the vector gets the expected value of 81 (0b01010001).

I think this is a bug, especially that the result is different in the two modules, which should have completely equivalent semantics.

zeldin:


enjoy-digital:
Thanks @zeldin!

PR #1776, opened 2023-09-13
core/naxriscv provide a deployable NaxRiscv SMP

Dolu1990:
Support multi core / coherent dma, can run doom on debian.

PR #1779, opened 2023-09-16
litex_setup.py: Disable auto-update of litex_setup for CI

cklarhorst:
This adds a new config option to litex_setup (--disable-auto-update) to disable the auto-update mechanism. It further changes the CI flow to use that flag.

Why? The auto-update is a problem for the CI workflow because it will override any (especially branch-specific) changes to the litex_setup script. Then CI is testing the wrong litex_setup script.

This should fix the CI flow for https://github.com/enjoy-digital/litex/pull/1776

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       # Install (n)Migen / LiteX / Cores
       - name: Install LiteX
         run: |
-          python3 litex_setup.py --config=full --init --install --user
+          python3 litex_setup.py --config=full --init --install --user --disable-auto-update
 
       # Install GCC Toolchains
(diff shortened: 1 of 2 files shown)


Dolu1990:
<3<3

enjoy-digital:
Hi and thanks @cklarhorst,

this is in fact the intend of `--dev` mode, I'll try to see if we can merge both.

cklarhorst:
@enjoy-digital 
I tried `--dev` first but as it rewrites the repo URLs to SSH it breaks the CI flow (at least for me) because it had no credentials.

enjoy-digital:
Indeed, there is now a CI detection to avoid this with https://github.com/enjoy-digital/litex/pull/1782.

PR #1786, opened 2023-09-23
litex_json2renode: fix --bios-binary and add --opensbi-binary

### LINKED ISSUE #1785: litex_json2renode.py --bios-binary doesn't load bios to rom - wiki example doesn't work
https://github.com/enjoy-digital/litex/wiki/Generate-Renode-simulation describes using --bios-binary to load the bios as a "quick start" example, but this was broken by [json2renode: Use opensbi->base for bios binary](https://github.com/enjoy-digital/litex/commit/860ca8673e8f81231f445e17e9a56c23cc7e08fe) which changes the behaviour to assume the "bios" is actually opensbi. This means the only supported option for this script is a smp cpu with opensbi.

I suspect that commit should really have added an alternate --opensbi-binary command line option.

I nearly gave up on renode for litex after a rather time consuming exercise trying to work out how to make the documented example work.

I also tried to use  --firmware-binary to load the bios to address 0x0, but this is hard coded to only load firmware to spiflash.

I had to manually edit the generated litex.resc to load the bios and set PC.

Note that the public renode examples mentioned in [Try out Renode quickly with precompiled LiteX demos](https://workshop.fomu.im/en/latest/renode-starting.html#try-out-renode-quickly-with-precompiled-litex-demos) do work. The renode litex examples in the renode distribution also work OK. Examples include: `renode @scripts/single-node/litex_vexriscv_smp.resc`

More examples (on arch linux):
```
% ls /opt/renode/scripts/single-node/litex*
/opt/renode/scripts/single-node/litex_ibex.resc
/opt/renode/scripts/single-node/litex_microwatt.resc
/opt/renode/scripts/single-node/litex_minerva.resc
/opt/renode/scripts/single-node/litex_nexys_video_vexriscv_linux.resc
/opt/renode/scripts/single-node/litex_vexriscv.resc
/opt/renode/scripts/single-node/litex_vexriscv_linux.resc
/opt/renode/scripts/single-node/litex_vexriscv_micropython.resc
/opt/renode/scripts/single-node/litex_vexriscv_sdcard.resc
/opt/renode/scripts/single-node/litex_vexriscv_smp.resc
/opt/renode/scripts/single-node/litex_vexriscv_tftp.resc
/opt/renode/scripts/single-node/litex_vexriscv_tock.resc
/opt/renode/scripts/single-node/litex_vexriscv_verilated_cfu.resc
/opt/renode/scripts/single-node/litex_vexriscv_zephyr.resc
```

AndrewD:
--bios-binary is reverted to loading the bios into _rom_ as per the documentation and litex defaults --opensbi-binary is added to load the openSBI binary into the opensbi memory region

The examples documented in the wiki now work again

Fixes #1785

The [Generate Renode simulation](https://github.com/enjoy-digital/litex/wiki/Generate-Renode-simulation) wiki page should also have one more section:

Finally, you can ~~run~~ load the simulation by executing the command:

```
renode litex.resc
```

Then in the renode terminal window execute the `start (or s)` command to boot the bios

```
(litex_vexriscv) start
```

@MateuszKarlic I think this change makes sense but as you have done more significant changes in the past maybe you can confirm this approach.

### DIFF
--- a/litex/tools/litex_json2renode.py
+++ b/litex/tools/litex_json2renode.py
@@ -883,15 +883,22 @@ showAnalyzer sysbus.uart Antmicro.Renode.Analyzers.LoggingUartAnalyzer
 """.format(cpu_type, args.repl)
 
     opensbi_base = csr['memories']['opensbi']['base'] if 'opensbi' in csr['memories'] else None
-    if opensbi_base is not None and args.bios_binary:
-        # load LiteX BIOS to ROM
+    if opensbi_base is not None and args.opensbi_binary:
+        # load OpenSBI to opensbi base
         result += """
 sysbus LoadBinary @{} {}
-""".format(args.bios_binary, hex(opensbi_base))
-
-    if opensbi_base:
+""".format(args.opensbi_binary, hex(opensbi_base))
         for cpu_id in range(0, number_of_cores):
             result += f"cpu{cpu_id} PC {hex(opensbi_base)}\n"
+
+    rom_base = csr['memories']['rom']['base'] if 'rom' in csr['memories'] else None
+    if rom_base is not None and args.bios_binary:
+        # load LiteX BIOS to ROM base
+        result += """
+sysbus LoadBinary @{} {}
+""".format(args.bios_binary, hex(rom_base))
+        for cpu_id in range(0, number_of_cores):
+            result += f"cpu{cpu_id} PC {hex(rom_base)}\n"
 
     if args.tftp_ip:
         result += """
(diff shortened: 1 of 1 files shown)


MateuszKarlic:
@AndrewD I confirm that both changes make sense. It was my mistake to change behavior of this switch.

In your commit, I'd suggest setting PC for all cores when loading Bios, not just `cpu0`, unless this somehow breaks the examples:
```python
for cpu_id in range(0, number_of_cores):
    result += f"cpu{cpu_id} PC {hex(bios_base)}\n"
```
Since the options are not mutually exclusive, it might be helpful to print message to user when loading both, that the PC will be set to `rom_base` not `opensbi_base`.

AndrewD:
> @AndrewD I confirm that both changes make sense. It was my mistake to change behavior of this switch.
> 
> In your commit, I'd suggest setting PC for all cores when loading Bios, not just `cpu0`, unless this somehow breaks the examples:
> 
> ```python
> for cpu_id in range(0, number_of_cores):
>     result += f"cpu{cpu_id} PC {hex(bios_base)}\n"
> ```

I wasn't sure it made sense to boot the bios on all cores, but I'm only using single core ATM so either is fine with me.

> Since the options are not mutually exclusive, it might be helpful to print message to user when loading both, that the PC will be set to `rom_base` not `opensbi_base`.

I wondered about making it mutually exclusive or adding a warning. As you've also raised this I'll make it mutually exclusive as I could not think of a reason to use both together.

AndrewD:
@MateuszKarlic thanks for the review!

MateuszKarlic:
Happy to help :)

PR #1790, opened 2023-09-26
[enh] generate SVD CSR enumeratedValue based on fields values

chmousset:
The [SVD schema](https://www.keil.com/pack/doc/CMSIS/SVD/html/schema_1_2_gr.html) supports listing possible values for CSR fields, as can be seen in [Keil's example](https://www.keil.com/pack/doc/CMSIS/SVD/html/svd_Example_pg.html).

Ideally, all existing field documented should be formatted as SVD schema expects it.
However, invalid values are filtered out using the schema's regex, guaranteeing a valid SVD output will be generated regardless.

Here is a snippet of a generated SVD (from linux-on-litex-vexriscv)

```xml
        <peripheral>
            <name>SDRAM</name>
...
            <registers>
                <register>
                    <name>DFII_CONTROL</name>
...
                    <fields>
                        <field>
                            <name>sel</name>
...
                            <enumeratedValues>
                                <enumeratedValue>
                                    <name>Software</name>
                                    <description>Software (CPU) control.</description>
                                    <value>0b0</value>
                                </enumeratedValue>
                                <enumeratedValue>
                                    <name>Hardware</name>
                                    <description>Hardware control (default).</description>
                                    <value>0b1</value>
                                </enumeratedValue>
                            </enumeratedValues>
```

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -848,6 +848,8 @@ def _svd_enumerated_values(field):
     values = []
     used_names = set()
     for value in field.values:
+        if not isinstance(value, (list, tuple)):
+            continue
         if len(value) == 2:
             raw_value, description = value
             name = description
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -826,6 +826,49 @@ def _svd_cdata(description):
     # across two CDATA sections (standard XML escaping technique).
     return "<![CDATA[{}]]>".format(str(description).replace("]]>", "]]]]><![CDATA[>"))
 
+def _svd_identifier(name):
+    identifier = re.sub(r"[^A-Za-z0-9_]+", "_", str(name).strip("` "))
+    identifier = identifier.strip("_")
+    if not identifier:
+        return None
+    if not re.match(r"[A-Za-z_]", identifier):
+        identifier = "Value_" + identifier
+    return identifier
+
+def _svd_enumerated_value(value):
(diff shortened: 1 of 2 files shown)


mithro:
@xobs / @bunnie - As likely users of this, could you review it?

xobs:
There was a program that was presented at Latchup 2023 that I am having trouble finding now. It was a browser-based register explorer. When I checked before, it didn't support SVD files, but that was a pending request. I'd love to see what that does with this, and what the progress is on SVD import.

A practical and more common usecase would be svd2rust (https://docs.rs/svd-rs/0.14.2/svd_rs/enumeratedvalues/struct.EnumeratedValues.html), which appears to support enumerated values.

xobs (approved):

xobs:
That's the one! I discovered https://github.com/SystemRDL/PeakRDL-svd/blob/main/docs/dev_notes.txt but there hasn't been any movement there in a few months...

enjoy-digital:
Thanks, this is still useful.

I rebased/reworked the PR on current `master` and kept the original feature: SVD export now emits `enumeratedValues` from `CSRField.values`.

I also added a follow-up cleanup/test commit covering:
  - normal 2-tuple `CSRField.values` entries;
  - explicit 3-tuple enum names/descriptions;
  - SVD-safe name sanitizing and duplicate handling;
  - filtering unsupported/malformed values instead of generating invalid SVD;
  - CDATA escaping for enum descriptions.

PR #1791, opened 2023-09-29
Update gcc installation instruction

lianakoleva:
Current installation instruction installs an older gcc which does not support all ISA instruction types passed in `-march` flags

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,8 +166,10 @@ $ ./litex_setup.py --update
 
 3. Install a RISC-V toolchain (Only if you want to test/create a SoC with a CPU):
 ```sh
-$ pip3 install meson ninja
-$ ./litex_setup.py --gcc=riscv
+$ git clone --recursive https://github.com/riscv/riscv-gnu-toolchain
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the update. The original concern was valid: some older RISC-V GCC toolchains did not support all ISA strings used by LiteX CPU variants.

Current LiteX no longer points users to the old pinned toolchain download. The README now recommends using:

```sh
  ./litex_setup.py --gcc=riscv
```
and litex_setup.py installs the platform package-manager RISC-V GCC when available (gcc-riscv64-unknown-elf on Debian/Ubuntu, Homebrew riscv64-elf-gcc on macOS, etc.).

So I would avoid replacing the quick-start instructions with a full riscv-gnu-toolchain source build. That is useful as an advanced/manual fallback, but it is too heavy for the default LiteX setup path.

I think we can close this PR as obsolete. If a specific CPU variant still needs a newer/special RISC-V toolchain, it would be better to track that as a focused issue with the exact CPU, GCC version and failing -march string.

PR #1795, opened 2023-10-07
interconnect/wishbone: Add linear burst support to DownConverter

zeldin:
When using VexRiscv-SMP with a 64-bit data bus (required by FPU) and keeping the system bus as 32 bit, burst information would previously get lost in the downconverter, leading to poor performance.

See also zeldin/litehyperram#1


enjoy-digital:
Thanks @zeldin! That's merged.

PR #1797, opened 2023-10-10
Change EOS S3 clock names

arturkow2:
F4PGA uses new clock names which differ from names used in QORC SDK.


trabucayre:
Applied. Thanks @arturkow2 
Sorry for the delay but I have waited a bit to see if it was possible to keep backward compatibility but QORC SDK seems more or less dead and yosys now uses the new clocks name: if the SDK is updated no doubt this renaming will be applied.

PR #1799, opened 2023-10-11
Fix EOS-S3 build on latest F4PGA

arturkow2:


litex/build/quicklogic/f4pga.py
```diff
@@ -47,48 +48,59 @@ def build_io_constraints(self):
...
+                    f"{self._build_name}.v"
+                ],
+                "synth_log": "synth.log",
+                "pack_log": "pack.log",
+                "analysis_log": "analysis.log"
+            },
+            "EOS3FF512-PDN64": {
+                "default_target": "bitstream",
+                "dependencies": {
+                    "build_dir": "build",
```
trabucayre, line 81: By setting `build_dir` to *build* directory all output files are located in this sub-dir instead of directly into *gateware* (as for all toolchains). But by using `self._build_dir` instead of build `prepare_cmd` will override *.sdc* file: what do you think about renaming *xxx.sdc* (in `build_timing_constraints`) by *xxx_in.sdc* ?

litex/build/quicklogic/f4pga.py
```diff
@@ -47,48 +48,59 @@ def build_io_constraints(self):
...
 
         if subprocess.call(make_cmd) != 0:
             raise OSError("Error occured during QuickLogic Symbiflow's script execution.")
+
+        make_cmd.append("--target")
+        if subprocess.call(make_cmd + ["bitstream_openocd"]) != 0:
+            raise OSError("Error occured during QuickLogic Symbiflow's script execution.")
+
+        if subprocess.call(make_cmd + ["bitstream_jlink"]) != 0:
+            raise OSError("Error occured during QuickLogic Symbiflow's script execution.")
```
trabucayre, line 106: These two commands produces only bitstream/rules for *openOCD* and *jlink* so we lost ability to obtain *.bit.bin* and *.bit.h* files: it's required to add `bitstream_bitheader` and `bitstream_binary` too.
What do you think about using a dict to iterate on all 4 targets and improving error message with target name?

trabucayre:
Applied. Thanks @arturkow2

PR #1806, opened 2023-10-22
Test litex-boards targets

AndrewD:
This change imports test_targets from litex_boards to verify litex changes do not break those targets.

### DIFF
--- a/test/test_targets.py
+++ b/test/test_targets.py
@@ -1,0 +1,88 @@
+#
+# This file is part of LiteX-Boards.
+#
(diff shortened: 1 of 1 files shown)


AndrewD:
Thanks, i wondered if that may be the case.

My motivation was to ensure litex changes test against litex-boards,
because this catches an issue I missed with a litex change. This was the
easiest way at the time. However now I've got an idea that should be simple
and eliminate the duplication but also work when running the tests
directly.

enjoy-digital:
I think we can close this, I prefer the approach of testing ci regularly (and avoid testing litex-boards directly in litex repo). But thanks @AndrewD for the suggestion/PR.

PR #1808, opened 2023-10-23
Xilinx: Initial openxc7 toolchain support

hansfbaier:
I also tidied up the yosys+nextpnr platform,
although I hardly think anyone might want to use it,
because has to be  hand built.


enjoy-digital:
Thanks @hansfbaier, I'm sorry but haven't followed closely the different efforts on 7-series reversing/tools. Would you mind refreshing my mind on the different efforts between F4PGA, Yosys+NextPnr-Xilinx and OpenXC7? With this in mind, I'll do review of the code and try to see if we can try to avoid duplicating too much code between the different efforts.

hansfbaier:
F4pga is a tool chain made up of yosys, VPR and prjxray backend. Openxc7 is yosys, nextpnr and prjxray just as yosys+nextpnr but using a patches prjxray (with kintex support) and a slightly patched nextpnr-xilinx, deviating a bit interns of paths from the nextpnr conventions. Mainly because building the chipdbs for nextpnr for each footprint take a LOT time and space and in openXC7 that is done on demand. Especially also considering that all those contents are 100% redundant to the information in prjxray-db.

enjoy-digital:
Thanks for the information, do you know if there are plans to merge Openxc7 and yosys+nextpnr approaches in the long term?

We had a first look with @trabucayre and except the very specific parts, the build backend seems very similar to YosysNextPNR and it would probably be easier to just keep YosysNextPNR and handle the specific parts directly in it. @trabucayre did a closer review and will probably answer with more details.

hansfbaier:
How am I going to tell the yosys+nextpnr backend which variant to use? And what should the default be? No Linux distro currently packages is in 'standard' form for Xilinx.

trabucayre:
My idea is to keep `yosys+nexpnr` and adding a new `opencx7` toolchain name.

*build/xilinx/platform.py* must be updated to replace this [section](https://github.com/enjoy-digital/litex/blob/master/litex/build/xilinx/platform.py#L42-L44) by something like
```
        elif toolchain in ["yosys+nextpnr", "openxc7"] :
            from litex.build.xilinx import yosys_nextpnr
            self.toolchain = yosys_nextpnr.XilinxYosysNextpnrToolchain(toolchain)
```

Of course the toolchain must be updated too to have a constructor param with toolchain name

hansfbaier:
OK, I implemented the requested changes, ready for review.

enjoy-digital:
Thanks @hansfbaier! This looks good now.

PR #1809, opened 2023-10-23
soc/integration/soc_core: add new parameters --with-uartbone and --with-jtagbone, deprecate crossover+uartbone

trabucayre:
- `--with-jtagbone` and `--with-uartbone` are now integrated in SoCCore arguments. This class also handle `add_jtagbone` and `add_uartbone`
- when a target try to add one of this option a warning is displayed and insertion is bypassed
- `crossover+uartbone` is deprecated -> `--uart-name=crossover --with-uartbone`
- jtag capability ((un)supported) is now handled at platform level


enjoy-digital:
Thanks @trabucayre!

Issue #1807, opened 2023-10-22
RFC: register new Interrupt Service Routine (ISR) funtions for additional interrupts


hvegh:
Dear Litex community,

**Question: how to handle additional interrupts in software?**
(I don't see currently an generic way of extending this. )
Currently the [isr.c](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/isr.c) only handles the uart.

A possible API might look like:
```
int set_handler(unsigned int irq, void (*handler)(void));

void (*handler)(void) get_handler(unsigned int irq);

int remove_handler(unsigned int irq);
```

My hack for the moment is a callback function in the Litex isr routine shown below.

_My little hack to extend the isr routine:_
```
diff --git a/litex/soc/software/libbase/isr.c b/litex/soc/software/libbase/isr.c
index ac7a7920..898eb40a 100644
--- a/litex/soc/software/libbase/isr.c
+++ b/litex/soc/software/libbase/isr.c
@@ -191,6 +191,9 @@ void isr(void)
 }
 
 #else
+
+void (*isr_next)(unsigned int irqs) = NULL;
+
 void isr(void)
 {
        __attribute__((unused)) unsigned int irqs;
@@ -203,6 +206,8 @@ void isr(void)
                uart_isr();
 #endif
 #endif
+       if(isr_next)
+               (*isr_next)(irqs);
 }
 #endif
```

_Code snippet using the callback function for my pps core:_
```
extern void (*isr_next)(unsigned int irqs);
    
void isr_pps(unsigned int irqs)
{
  if (irqs & (1 << PPS0_INTERRUPT)) {
     // handle stuff
  }
}

static void pps_init(void) {
  isr_next = &isr_pps;
  irq_setmask(irq_getmask() | (1 << PPS0_INTERRUPT));
  pps0_ev_enable_write(1);
}
```

AndrewD:
There is another pr for a different approach to this.
I've got a more traditional irq handler registration (proof of concept) in our tree I've been meaning to push for discussion.

hvegh:
Now made obsolete by the pull request Andrew is working on. https://github.com/enjoy-digital/litex/pull/1815
Thanks!

PR #1815, opened 2023-10-27
soc/software: add irq_attach() / irq_detach()

AndrewD:
cleaner mechanism for other software to use interrupts

NOTE: uart_isr(void) not changed to uart_isr(int) for compatibility with cpus that don't implement this interface yet.

This is a different approach to #1807 and #1735

### DIFF
--- a/litex/soc/software/libbase/isr.c
+++ b/litex/soc/software/libbase/isr.c
@@ -191,18 +191,45 @@ void isr(void)
 }
 
 #else
+struct irq_table
+{
+	isr_t isr;
+} irq_table[NR_IRQ];
+
+int irq_attach(unsigned int irq, isr_t isr)
+{
+	if (irq >= NR_IRQ) {
+		printf("Inv irq %d\n", irq);
+		return -1;
+	}
+
+	unsigned int ie = irq_getie();
+	irq_setie(0);
+	irq_table[irq].isr = isr;
+	irq_setie(ie);
+	return irq;
+}
+
+int irq_detach(unsigned int irq)
+{
+	return irq_attach(irq, NULL);
+}
+
 void isr(void)
 {
-	__attribute__((unused)) unsigned int irqs;
+	unsigned int irqs = irq_pending() & irq_getmask();
 
-	irqs = irq_pending() & irq_getmask();
-
-#ifdef CSR_UART_BASE
-#ifndef UART_POLLING
-	if(irqs & (1 << UART_INTERRUPT))
-		uart_isr();
-#endif
-#endif
+	while (irqs)
+	{
+		const unsigned int irq = __builtin_ctz(irqs);
+		if (irq < NR_IRQ && irq_table[irq].isr)
+			irq_table[irq].isr();
+		else {
+			irq_setmask(irq_getmask() & ~(1<<irq));
+			printf("\n*** disabled spurious irq %d ***\n", irq);
+		}
+		irqs &= irqs - 1; // clear this irq (the first bit set)
+	}
 }
 #endif
 
--- a/litex/soc/software/libbase/uart.c
+++ b/litex/soc/software/libbase/uart.c
@@ -110,6 +110,8 @@ void uart_init(void)
 
 	uart_ev_pending_write(uart_ev_pending_read());
 	uart_ev_enable_write(UART_EV_TX | UART_EV_RX);
+	if (irq_attach)
+		irq_attach(UART_INTERRUPT, uart_isr);
 	irq_setmask(irq_getmask() | (1 << UART_INTERRUPT));
 }
 
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1320,6 +1320,7 @@ class SoC(LiteXModule, SoCCoreCompat):
 
         # SoC IRQ Interconnect ---------------------------------------------------------------------
         if hasattr(self, "cpu") and hasattr(self.cpu, "interrupt"):
+            self.add_constant("NR_IRQ", max(self.irq.locs.values()) + 1)
             for name, loc in sorted(self.irq.locs.items()):
                 if name in self.cpu.interrupts.keys():
                     continue
--- a/litex/soc/software/include/irq.h
+++ b/litex/soc/software/include/irq.h
@@ -1,0 +1,15 @@
+#pragma once
+#include_next<irq.h>
+
+typedef void (*isr_t)(void);
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern int irq_attach(unsigned int irq, isr_t isr) __attribute__((weak));
+extern int irq_detach(unsigned int irq) __attribute__((weak));
+
+#ifdef __cplusplus
+}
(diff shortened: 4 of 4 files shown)


hvegh:
I like it, tank you very much!

I would add: 
```
void uart_isr_new(int irq) { 
     uart_isr();
}
...
if (irq_attach)
		irq_attach(UART_INTERRUPT, uart_isr_new);
 ...
```
so we don't assume compiler will drop the irq parameter.

AndrewD:
Next step is a cleanup pass: i was either going to do something like you proposed or the more complete solution of adjusting the function signature.

One concern with my proposal is that it increases ram usage for all targets by 128 byes. If this is a concern there are some clean ways to offer something like CONFIG_DYNAMIC_IRQ and have a constant table in rodata instead of bss.

hvegh:
Could we limit the jump table to just the registered IRQ's?
The highest IRQ number can be automatically generated as a define in `soc.h`, see below.

This will limit ram demand to 4 bytes per interrupt, seams quite reasonable.

```
diff --git a/litex/soc/integration/soc.py b/litex/soc/integration/soc.py
index 2ed96f46..b71761cc 100644
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1278,6 +1278,7 @@ class SoC(LiteXModule, SoCCoreCompat):

         # SoC IRQ Interconnect ---------------------------------------------------------------------
         if hasattr(self, "cpu") and hasattr(self.cpu, "interrupt"):
+            self.add_constant("INTERRUPT_END", max(self.irq.locs.values()))
             for name, loc in sorted(self.irq.locs.items()):
                 if name in self.cpu.interrupts.keys():
                     continue
```

This will add `#define INTERRUPT_END 2` to `soc.h`

litex/soc/software/libbase/isr.c
```diff
@@ -191,18 +191,46 @@ void isr(void)
...
+	irq_setie(ie);
+	return irq;
+}
+
+int irq_detach(unsigned int irq)
+{
+	return irq_attach(irq, NULL);
+}
+
+void isr(void)
```
hvegh, line 219: Some considerations:
- Assert if irq within NR_IRQ
- NR_IRQ can be generated in soc.py see https://github.com/enjoy-digital/litex/pull/1815#issuecomment-1784151465

litex/soc/software/libbase/uart.c
```diff
@@ -110,6 +110,8 @@ void uart_init(void)
 
 	uart_ev_pending_write(uart_ev_pending_read());
 	uart_ev_enable_write(UART_EV_TX | UART_EV_RX);
+	if (irq_attach)
+		irq_attach(UART_INTERRUPT, (isr_t)uart_isr);
```
hvegh, line 114: Possibly need wrapper see https://github.com/enjoy-digital/litex/pull/1815#issuecomment-1783806810

litex/soc/software/libbase/isr.c
```diff
@@ -191,18 +191,46 @@ void isr(void)
 }
 
 #else
-void isr(void)
+#define NR_IRQ 32
```
hvegh, line 194: Generate this define using soc.py?

hvegh (changes requested):
Looking good, I have added my comments.

litex/soc/software/libbase/uart.c
AndrewD: Maybe, or just change the function signature. I just avoided this initially to minimise diffs.

litex/soc/software/libbase/isr.c
AndrewD: Yes - I like that solution.

AndrewD:
I'll make some tweaks soon, just not sure exactly when.

hvegh:
Hmm just thinking here: 

Calling `my_isr(irq)` is not that useful most of the time, you will already know which IRQ you signed up for. In simple cases 1-1 mapping beween ISR and the interrupt the `irq` parameter can be ignored.

Unless ...  you register the same interrupt handler for multiple interrupt lines:

Then you will have to scan in your driver data for the structure/context for this particular interrupt.
These scans are costly, more importantly you would want these delay's as deterministic as possible. If you have to cycle trough an array to find the interrupt context, the difference between handling the first interrupt found and the last might be several cycles.

This extra jitter introduced by scanning could be reduced if we introduce a `user_data` pointer:
- We will leave out the `irq` parameter since it is not useful in simple cases anyway.
- We will replace the `irq` parameter with a `user_data` pointer
- Downside is irq_table will be doubled.
- Upside for generic isr functions:
  - less overhead cycles for generic irq handlers
  - less jitter when handling multiple interrupts within the same handler, typical use case maybe GPIO 

We could even go the extra step to initialize a default interrupt handler at compile time removing the need for a NULL check:

Putting it all together:

`typedef void (*isr_t)(void *);
`


```
// generated by soc.py
void isr_default(void *);

static struct irq_table
{
        isr_t   isr;
        void    *user_data;
} irq_table[NR_IRQ] = {
        { isr_default, (void *)0 },
        { isr_default, (void *)1 },
        { isr_default, (void *)2 },
};
```

```
void irq_default(void *data)
{
        unsigned long irq = (unsigned long)data;
        irq_setmask(irq_getmask() & ~(1<<irq));
        printf("\n*** disabled spurious irq %ld ***\n", irq);
}

int irq_attach(unsigned long irq, isr_t isr, void *user_data)
{
     if(isr == NULL)
        return -EINVAL;
     if(irq >= NR_IRQ)
        return -EINVAL;
     irq_table[irq].isr = isr;
     irq_table[irq].user_data = user_data;
}

int irq_detach(unsigned long irq)
{
    return irq_attach(irq, isr_default, (void *)irq);
}
```

hvegh:
I forgot the disable the interrupts when updating the irq_table in the above example..

AndrewD:
Many isr interfaces include both irq and a content pointer. I'm not currently using either, but included irq as it was easy with minimal size implications and was compatible with the cpu interfaces that haven't been updated. It also allows for simple compile time integration if that is a requirement that needs to be maintained. 

Yes a context pointer could be useful one day, but chose to not add something i have no need for (currently) in a first pass at this. Following the principal of not letting perfect get in the way of good 😊

Is this something you would actually use? I'm likely to need it at some stage but thia is just a minor infrastructure tweak I want to upstream ASAP to reduce the number of patches I'm maintaining in our trees.

hvegh:
I must admit, I don't need perfect either ;) 
Having an API for ISR registration is the main point. Thank you for the effort!
Lets upstream ASAP!

AndrewD:
Final cleanup - I've dropped the unused irq parameter to minimise pointless changes. This or a content pointer can be reconsidered in the future.

enjoy-digital:
Thanks @AndrewD, I just did this change on top of it: https://github.com/enjoy-digital/litex/commit/edc6871aceefcdeef16a76e7d70afb52f7229115

AndrewD:
I think NR_IRQ is a hangover from my time following linux coding styles :)

PR #1816, opened 2023-10-27
test_csr: test cases to demonstrate a CSRStatus() issue

AndrewD:
'status' reads as 0 in simulation when CSRStatus has fields that are assigned in the design. This test case demonstrates the issue  but is wrapped in a try/catch so the rest of the tests can run without a CI failure.

I'm not sure how to fix the underlying issue, but can work around it in my core tests that discovered this issue.


enjoy-digital:
Thanks @AndrewD, it's merged and I'll look at it.

PR #1821, opened 2023-10-30
UARTBone addr width

trabucayre:
This PR contains 3 commits:
- `Stream2Wishbone`: update allowed address_width list to support 64bits (since number of iterations is based on this parameter any size multiple of 8bits may works)
- `UARTBone`: now it's possible to pass address_width to the constructor (this param is used for Stream2Wishbone)
- soc: `self.bus.address_width` is now used at `UARTBone` constructor time


enjoy-digital:
Thanks @trabucayre, this looks good!

PR #1822, opened 2023-10-30
tools: litex_server, litex_client and remote: don't hardcode bus address

trabucayre:
This PR has for goal to remove all hardcoded address width.
- For `litex_client` the address width is obtained from *csr.csv*
- For `litex_server` a new argument is added to select addr_width (default: 32bits)


enjoy-digital:
Thanks @trabucayre!

PR #1823, opened 2023-10-30
litex/soc/cores/cpu/__init__, litex/soc/integration/soc: modifying CPUNone to adapt data_width and io_regions according to bus data_width/address_width

trabucayre:
`CPUNone` has a fixed data width and io_regions.
By adding a constructor with data_width and address_width it's now possible to adapts CPU according to args provided by user.


enjoy-digital:
Thanks @trabucayre!

PR #1824, opened 2023-10-31
soc/interconnect/wishbone: InterconnectShared,Crossbar: using master adr_width for Interface constructor

trabucayre:
Allows to have masters and slaves with an address_bus != 32bits.


enjoy-digital:
Thanks @trabucayre!

PR #1759, opened 2023-09-01
Builder enhancements

AndrewD:
Various enhancements to reduce boilerplate code in target arguments and to cleanly allow a target to initialise memory with a custom firmware (for example).

Custom firmware example:

class MySimSoC(SoCCore):
    def __init__(self, firmware, **kwargs):

        self.firmware = firmware

        if self.firmware:
            self.add_config("MAIN_RAM_INIT") # target software is in ram
           self.add_constant("ROM_BOOT_ADDRESS", self.bus.regions["main_ram"].origin)

    def initialize_memory(self, software_dir, **kwargs):
        if not self.firmware:
            return

        filename  = os.path.join(software_dir, "firmware", "firmware.bin")
        data = get_mem_data(filename, endianness="little")
        # This can not generate MAIN_RAM_INIT so must set it manually in soc.__init__()
        self.init_rom(name="main_ram", contents=data, auto_size=False)

# Build --------------------------------------------------------------------------------------------

def main():
    from litex.build.parser import LiteXArgumentParser
    bool_action = argparse.BooleanOptionalAction
    parser = LiteXArgumentParser(platform=Platform, soc=MySimSoC, description="My simulation.")

    # NOTE: all target arguments have a default here, not in soc.__init__(), so they are documented.
    parser.add_target_argument("--firmware", default=True, action=bool_action,      help="compile and load firmware in main_ram.")

    # Create SoC.
    soc = MySimSoC(**parser.soc_argdict)

    # add custom firmware: compiled by connecting here and stored in initialize_memory()
    src="firmware"
    src_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), src)
    builder = Builder(soc, **parser.builder_argdict)
    builder.add_software_package(src, src_dir)

    # Build/Run SoC
    builder.build(sim_config=sim_config, **parser.toolchain_argdict, run=True)

PR #1840, opened 2023-11-14
Parser set defaults improvements

AndrewD:
Improvements to LiteXArgumentParser when using set_defaults() in a target to ensure args_fill() for the new cpu_type is called and to report using set_default(arg=default) for an argument that doesn't exist.

### DIFF
--- a/litex/build/parser.py
+++ b/litex/build/parser.py
@@ -217,7 +217,8 @@ class LiteXArgumentParser(argparse.ArgumentParser):
                 self._platform.fill_args(self._toolchain, self)
 
         # Intercept selected CPU to fill arguments.
-        cpu_cls = cpu.CPUS.get(self.get_value_from_key("--cpu-type"), None)
+        default_cpu_type = self._args_default.get("cpu_type", None)
+        cpu_cls = cpu.CPUS.get(self.get_value_from_key("--cpu-type", default_cpu_type))
         if cpu_cls is not None and hasattr(cpu_cls, "args_fill"):
             cpu_cls.args_fill(self)
 
--- a/litex/build/parser.py
+++ b/litex/build/parser.py
@@ -225,6 +225,13 @@ class LiteXArgumentParser(argparse.ArgumentParser):
         # Injects arguments default values
         if len(self._args_default):
             argparse.ArgumentParser.set_defaults(self, **self._args_default)
+            # Catch defaults which do not match any arguments - typos?
+            remaining = list(self._args_default.keys())
+            for action in self._actions:
+                if action.dest in remaining:
+                    remaining.remove(action.dest)
+            if len(remaining) > 0:
+                raise ValueError(f"set_default() for invalid argument(s): {remaining}")
 
         # Parse args.
         self._args = argparse.ArgumentParser.parse_args(self, args, namespace)


AndrewD:
This is required for this to work:

```
parser.set_defaults(cpu_type="vexriscv_smp")
parser.set_defaults(cpu_variant="linux")
parser.set_defaults(with_fpu=True)
```

litex/build/parser.py
```diff
@@ -217,13 +217,21 @@ def parse_args(self, args=None, namespace=None):
...
         # Injects arguments default values
         if len(self._args_default):
             argparse.ArgumentParser.set_defaults(self, **self._args_default)
+            # Catch defaults which do not match any arguments - typos?
+            remaining = list(self._args_default.keys())
+            for action in self._actions:
+                if action.dest in remaining:
+                    remaining.remove(action.dest)
+            if len(remaining) > 0:
+                raise ValueError(f"set_default() for invalid argument(s): {remaining}")
```
trabucayre, line 234: Having a test and/or a message looks good, but raising an exception may, potentially being a problem for users: 
- if in a target I set `cpu_type=vexriscv_smp` and `dtlb_size=xxx`
- if a user decide to use `--cpu-type=naxriscv`

An exception will occur because `dtlb_size` is something specific to the `vexriscv_smp`.

Maybe it's better here to displays a message (similar to [compat_notice](https://github.com/enjoy-digital/litex/blob/master/litex/compat/__init__.py#L6)) with unused/unknown arguments list with and a delay to warn user about situation. Otherwise user has to modifiy target to remove/comment default and it's not really user friendly.

trabucayre:
To complete my previous message (because I have, maybe,seen why you have added the `raise KeyError`).
If my understanding is true it's to mimic, for `set_defaults`, argparse's behavior, no?
In fact I see three cases:
- a user/dev adds some `set_defaults` in the target (for CPU configuration and/or to select/configure toolchain): here we assume he know what it do, and with a warning/delay it will able to fix a potential typo.
- a user select some CLI's arguments: if there is a typo or unknown argument: it make sense to stop because it's an unwanted situation: it's not possible to build gateware with or without parameters selected by user, due to a typo.
- last possibily is when user select/change, with CLI, some parameters affecting `set_defaults`: we can consider user know why this change, and if a defaults  value/argument is no more relevant, it make sense to warn but its not an error (in my mind).

AndrewD:
Thanks for the detailed review.

From what I've seen, there are many cases where the litex behaviour is to raise an exception of something goes wrong so this pattern is not unusual.

However I don't remember my exact thought process on this as I added it a long time ago. I don't mind how the error is reported, I simply wanted it to be clear that the arguments were incorrect instead of being silently ignored as a typo in set_default() is otherwise far harder to find and I lost some time on several occasions due to this, hence I added something to detect the issue and stop.

Ultimately this exception is only thrown if there is a bug in the target file.

For example I guard set_defauts() for cpu specific arguments based on cpu_type so this exception can be resolved by writing a sane target 😀

AndrewD:
For example this is how I use set_defaults() for cpu_type to guarantee the defaults are valid while also generating a target that will, by default, build with our required features:
```
if not any("--cpu-type" in arg for arg in sys.argv):
    parser.set_defaults(cpu_type="vexriscv_smp")
    parser.set_defaults(cpu_variant="linux")
    parser.set_defaults(with_coherent_dma=True)
    parser.set_defaults(with_rvc=True)
    parser.set_defaults(with_fpu=True)
    parser.set_defaults(with_privileged_debug=True)
    parser.set_defaults(with_wishbone_memory=True)
    parser.set_defaults(expose_time=True)
    parser.set_defaults(hardware_breakpoints=4)
```

There are a range of other arg parsing refinements I also made a while ago in an PR here: #1759. This small piece was an attempt to break the original larger change set into smaller chunks to be easier to review and merge.

trabucayre:
> For example this is how I use set_defaults() for cpu_type to guarantee the defaults are valid while also generating a target that will, by default, build with our required features:
> 
> ```
> if not any("--cpu-type" in arg for arg in sys.argv):
>     parser.set_defaults(cpu_type="vexriscv_smp")
>     parser.set_defaults(cpu_variant="linux")
>     parser.set_defaults(with_coherent_dma=True)
>     parser.set_defaults(with_rvc=True)
>     parser.set_defaults(with_fpu=True)
>     parser.set_defaults(with_privileged_debug=True)
>     parser.set_defaults(with_wishbone_memory=True)
>     parser.set_defaults(expose_time=True)
>     parser.set_defaults(hardware_breakpoints=4)
> ```
Yes it's always possible to do something like that, but it adds a bit more complex and break the idea behind `LiteXArgumentParser`: reducing complexity at targets level. It also possible to think to a `set_defaults` variant with a condition to only inject this when cond is True (but it's beyond this PR scope).
> 
> There are a range of other arg parsing refinements I also made a while ago in an PR here: #1759. This small piece was an attempt to break the original larger change set into smaller chunks to be easier to review and merge.

Yes it always better to have a small patches serie (ie multiple PR) with small modifications/fix/new features and a commit message enough to see why this PR (something similar to a primary school's math demonstration) : it's simplify the task for maintainers/reviewers and allows to be merged more quickly.

For this PR, first part looks good because it fix a specific issue where cpu-type is not taken into account when it must: with some potential issue for CPU configuration. For the second part I'm a bit mitigated because it make sense to help developer to fix some typos but in the same time it will introduce a potential build failure when user select parameters different than default:
- developer know what it do and if a warning is displayed may fix is code
- user is not always aware about targets content and may not be able to see/know why build fails due to an option not provided in the command line.

AndrewD:
I can drop the second patch from this pr if that helps?
But hardly any targets use set_default at the moment so it's safe from that regard 😂

enjoy-digital:
@AndrewD, @trabucayre: I think we can merge as is since already fixing issues with CPU defaults and since mostly experienced developpers will be using the set_defaults for fixed configuration, raising an error when the argument is not supported seems fine. 

I'm merging and will do a more global review when I'll have a bit more time.

AndrewD:
Thanks @trabucayre and @enjoy-digital !

PR #1845, opened 2023-11-21
patch litex soc gen

jwfaye:


enjoy-digital:
Thanks for the PR. I agree this points at a real limitation of `litex_soc_gen`: today it exposes one fixed `mmap_s` and one fixed `mmap_m`, with the `mmap_m` region hardcoded.

I would not merge this implementation as-is though. It feels too ad-hoc for upstream:
  - it changes the default external interface behavior instead of preserving compatibility;
  - the JSON naming is confusing, since `masters` become SoC slaves and `slaves` become SoC masters;
  - `base`/`size` are present for slave entries but are not used;
  - there is no validation/test/doc, and the example file looks more like a local use case than a stable LiteX interface.

If we improve this, I think it should be done with a smaller compatibility-preserving change: either make the existing `mmap_m` region configurable, or design one clean optional external-interface config format with validation and tests while keeping the current defaults.

So I would prefer not to merge this PR directly. If this is still needed, we can track a cleaner `litex_soc_gen` external bus configuration separately.

PR #1850, opened 2023-12-07
Efinix PLL calc when feedback != INTERNAL

trabucayre:
Efinix python API has a bug when feedback is different than "INTERNAL" with, as a result, a wrong configuration. *LOCAL* or *CORE* mode is required for LVDS/serdes rx to align input clk and clk used by LVDS blocks.

This PR currently  only address Trion devices


enjoy-digital:
Thanks @trabucayre!

PR #1851, opened 2023-12-08
Add AXI/AXILite 64 bus support

trabucayre:
This PR mainly adds code to propagate bus.address_width.

It also fix a multiple read to CSR


enjoy-digital:
Thanks @trabucayre!

PR #1852, opened 2023-12-13
soc/software/bios/boot: serialboot: increase CMD_TIMEOUT_DELAY

trabucayre:
When using *jtag_uart* with a big file, `serialboot` fails due to a timeout. since JTAG is slower than classic UART, the timeout delay must increased to avoid this failure.
Tested with an arty, femtorv and 1.2MB file


enjoy-digital:
Thanks @trabucayre!

PR #1856, opened 2023-12-22
soc/builder: add initialize_memory()

AndrewD:
Allows the target soc to override memory initialization:
  rom and/or ram.

Includes documentation for example usage to compile soc target specific software and load it to main_ram. This mechanism
can also be used to load a custom bios into ROM instead of the standard Litex bios.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -388,8 +388,11 @@ class Builder:
                 self._prepare_rom_software()
                 self._generate_rom_software(compile_bios=use_bios)
 
+                # Allow soc to override the memory initialisation.
+                self.soc.initialize_memory(self.software_dir, **kwargs)
+
                 # Initialize ROM.
-                if use_bios and self.soc.integrated_rom_size:
+                if use_bios and self.soc.integrated_rom_size and not getattr(self.soc, "rom").mem.init:
                     self._initialize_rom_software()
 
         # Translate compile_gateware to run.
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Sorry for the delay in the review @AndrewD. I'll review it now and will probably modify things a bit. I also have a preference to avoid that much documentation/example directly in the code and prefer having it in a wiki. I'll do the change and will provide an update here.

enjoy-digital:
@AndrewD: I've reworked things a bit on top of your commit with:
- https://github.com/enjoy-digital/litex/commit/f181eabebb6a01b2a171cdcb3abfd18d3232de59
- https://github.com/enjoy-digital/litex/commit/9f49b13f8c8ac48b4cf23746be27bc63a21495d7
- https://github.com/enjoy-digital/litex/commit/175aab73a835d777d1be99bc8679a124a1e55e35
- https://github.com/enjoy-digital/litex/commit/06083c7863b0063c06df2b11b34b75c2a739a402
- https://github.com/enjoy-digital/litex/commit/31508ddfa43d1fe277050c1dc6a57a1d0ee545cd
- https://github.com/enjoy-digital/litex/commit/38a8a171dd04d6c6a635709c651b098a94783902
Can you check it still allow you to do what you wanted? 

For an example of the usage, we could add it to the wiki or to a specific design requiring it.

AndrewD:
I was thinking of documenting some usage in the wiki, for example this can improve the demo firmware integration.
I used this for much cleaner integration of unit testing firmware, but it can also be used to load, for example, a customised bios for a target.

PR #1874, opened 2024-01-18
soc/cores/cpu/naxriscv fix arch definition and small adjust

trabucayre:
With recent gcc software fails with:
```
litex/litex/soc/cores/cpu/naxriscv/system.h:12: Error: unrecognized opcode `fence.i', extension `zifencei' required
```
`fence.i` seems have been moved out of **i** ISA

`rvXXima` **arch** must be replaced by `rvXXi2p0_ma`


trabucayre:
The second commit is to avoid some issue when target build is done into a python script. Before the git clone  `os.chdir` is called. But after this step current directory remains `pythondata-cpu-naxriscv` and caller script fails if some task must be done after without using absolute path.

enjoy-digital:

PR #1877, opened 2024-01-24
soc/integration/soc: add_etherbone: allowing to specify local/remote IP

trabucayre:
For `add_etherbone` user, at target level, may specify **eth_ip**.
But for `add_ethernet` there is no way to provides interface IP and/or remote IP: user must add piece of code to override constants.

This PR add `local_ip` and `remote_ip` to pass these informations at build time, and adds required piece of code to update constants at `add_ethernet` level.


enjoy-digital (approved):
Thanks @trabucayre.

PR #1878, opened 2024-01-27
cpu/rocket: change load path of generated verilog files

inochisa:
Since rocket-chip upstream switched to mill build, it can not generate a single verilog file, instead, it generate a lots of file in build directory. To save space, we do deduplication for the common file. This means that we need to load verilog file from different directory (common code and config specific).

Add all the path needed for generated verilog files.

related change: https://github.com/litex-hub/pythondata-cpu-rocket/pull/6

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -29,6 +29,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+import glob
 import os
 import logging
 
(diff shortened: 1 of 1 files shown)


gsomlo:
Hi @inochisa,

Thanks for the PR. I'll have to review this, but more importantly,
your pythondata-cpu-rocket PR at
https://github.com/litex-hub/pythondata-cpu-rocket/pull/6
and evaluate it for any additional implicit dependencies (on top of
just `mill` itself), which might take me a bit of time, so thanks in
advance for your patience.

inochisa:
@gsomlo Great, let's wait the [litex-hub/pythondata-cpu-rocket#6](https://github.com/litex-hub/pythondata-cpu-rocket/pull/6) and see what is needed.

gsomlo:
An experimental branch of `pythondata-cpu-rocket` has been created to support `mill`/`firtool` based builds, here: https://github.com/litex-hub/pythondata-cpu-rocket/compare/master...mill-experimental

The corresponding LiteX changes will, for the time being, be available here: https://github.com/enjoy-digital/litex/compare/master...gsomlo:litex:rocket-mill-experimental

To reiterate some of the issues preventing the main/master branches from simply being updated:
- the latest upstream rocket results in "bloated" verilog that requires significantly more LUTs than before (e.g., 4-core Linux variant no longer fits on the digilent nexys video board)
- the latest upstream results in elaborated `*.sv` files that can't be loaded into `yosys`, which would remove support for ECP5 and the FOSS yosys/trellis/nextpnr toolchain.

Until these issues are resolved, we can experiment with the latest rocket upstream, but can ***not*** adopt it in an "official" capacity.

PR #1888, opened 2024-02-13
core/vexriscv_smp add reset vector support

Dolu1990:
--cpu-reset-address  can now be used to set the reset vector of vexriscv-smp

For instance, you can skip the litex bios and go directly to opensbi via  :
litex_sim --cpu-type=vexriscv_smp --with-sdram --sdram-init boot.json --with-rvc  --with-privileged-debug --hardware-breakpoints 4  --cpu-reset-address 0x40f00000 --integrated-rom-size 0


enjoy-digital:
Thanks @Dolu1990!

PR #1894, opened 2024-02-21
Send serial data to sim uart only if rx_ready is set.

mdejw:
This prevents that characters are lost in case of back pressure on sim UART RX.


enjoy-digital:
Thanks @mdejw!

PR #1895, opened 2024-02-22
tools/litex_json2dts_linux.py: adding ip= bootarg when localip and remoteip are provided

trabucayre:
Configuring network interface may be done by three different ways:
1. using `ifconfig` to configure interface and `route` for gateway
2. updating */etc/network/interfaces with all required informations
3. using `ip=` bootargs (requires to have a kernel configured with `CONFIG_IP_PNP=y`

First option needs a user's manual intervention and must be done for each boot. Second is hardcoded and this file must be updated when user change ip configuration.

Last option is more easy since, when ethernet is enabled, json contains all required informations to fill the `ip=` bootargs.


enjoy-digital:
Thanks @trabucayre!

PR #1897, opened 2024-02-29
support TQFP144 ECP5 package

machdyne:
Hello. I had to add this in order to use the TQFP144 ECP5 package, with the [Vanille](https://github.com/machdyne/vanille) board.

This detects if the device is set to "LFE5U-12F-6TG144", etc. I don't think there is a 144-pin BGA package for ECP5.


enjoy-digital:
Thanks @machdyne!

PR #1902, opened 2024-03-07
Rocket update

trabucayre:
This PR try to update rocket core and litex_json2dts_linux to mimic has much as possible dts available in *pythondata-cpu-rocket*.

### DIFF
--- a/litex/soc/cores/cpu/rocket/core.py
+++ b/litex/soc/cores/cpu/rocket/core.py
@@ -340,20 +340,20 @@ class Rocket(CPU):
         soc.add_config("CPU_MMU",   "sv39")
 
         # Constants for Cache so we can add them in the DTS.
-        soc.add_config("CPU_DCACHE_SIZE",       4096) # CHECKME: correct/hardwired?
-        soc.add_config("CPU_DCACHE_WAYS",          2) # CHECKME: correct/hardwired?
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre!

gsomlo:
On Thu, Mar 07, 2024 at 03:35:54AM -0800, Gwenhael Goavec-Merou wrote:
> This PR try to update rocket core and litex_json2dts_linux to mimic has much as
> possible dts available in pythondata-cpu-rocket.

The updated values happen to match "Full" and "Linux" variants at the
current time -- which I guess is an improvement, but no guarantees the
numbers won't change if we ever decide to update pythondata-cpu-rocket
to newer upstream sources.

The (new) numbers will be wrong if building with Medium or Small
(which can also run linux, just not an actual *distro* like Fedora
or Debian).

Note to self: think harder about how to parse the provided `dts`
samples in pythondata-cpu-rocket for the purpose of 'cut'n'pasting`
these values directly from there into the output of json2dts, without
hardcoding any of them anywhere in the sources... :)

trabucayre:
True and these changes are a bit boring to maintain this CPU in the long term.
Knowing verilog file name I see two possibilities to have dynamically access to these informations:
- by using *pylibfdt* to parse the dts
- maybe (if both are sync) using the associated json file.

PR #1906, opened 2024-03-20
Avoid extra timing delays for NXLRAM path

jdavidberger:
The large block ram is pretty slow as is, so avoid the conditional assignment of dataout. 

I'm new to migen so apologies if this isn't correct; but the change does generate the verilog I expected and relieved some timing pressure so I think it should be good.


enjoy-digital (approved):
@jdavidberger, this indeed make perfect sense. Thanks for the PR!

enjoy-digital:
@jdavidberger: We just had to revert while doing more tests on CertusPro-NX FPGAs. The change was only working in all cases. This is reverted with https://github.com/enjoy-digital/litex/commit/ecd0f0e5486ee542d1aca96ec4082f6443394f3e. We'll see to improve timings if problematic by adding a register stage.

PR #1908, opened 2024-03-23
Update NeoRV32 CPU to v1.9.7

Irvise:
This updates the NeoRV32 CPU to the latest (current) version.

This also fixes a constant issue with Whishbone that I have been having where the data is always corrupted.

I have tested this update on my ULX3S using the latest oss-cad-toolchain build.


enjoy-digital:
Thanks @Irvise!

PR #1909, opened 2024-03-26
sim/video: Change pixel format to RGBA

davidar:
The framebuffer gets filled out in RGB order (https://github.com/enjoy-digital/litex/blob/master/litex/build/sim/core/modules/video/video.c#L156-L159), but the texture format was set to BGR, swapping the colour channels.

I also added a `--with-video-colorbars` option to `litex_sim` to make this easier to test.

Before:

Screenshot 2024-03-26 200834

After:

Screenshot 2024-03-26 202011

Note that the order is supposed to be (https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/video.py#L341-L348):

White Yellow Cyan Green Purple Red Blue Black


enjoy-digital:
Thanks @davidar!

PR #1911, opened 2024-03-31
sim/video: Add option to render only on frame vsync

davidar:
By default the video simulator renders a single scanline at a time, which throttles the speed of the simulation quite a lot in cases where it would otherwise be capable of rendering multiple frames per second. This adds an argument that allows it to render only full frames, improving performance (and reducing tearing). I've left the default behaviour unchanged as I can see that would be useful for more complex designs that take multiple seconds to simulate each frame.


enjoy-digital (approved):
Thanks @davidar, this looks good.

PR #1913, opened 2024-04-02
Auto format constants in decimal or hex

AndrewD:
Automatically format constants in decimal or hex for readability in verilog and generated headers. A simple implementation inspired by #1528.

### DIFF
--- a/litex/gen/fhdl/verilog.py
+++ b/litex/gen/fhdl/verilog.py
@@ -29,6 +29,7 @@ from migen.fhdl.specials    import Instance, Memory
 from litex.gen import LiteXContext
 from litex.gen.fhdl.namer     import build_signal_namespace
 from litex.gen.fhdl.hierarchy import LiteXHierarchyExplorer
+from litex.gen.format import format_constant
 
 from litex.build.tools import get_litex_git_revision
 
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks @AndrewD, the approach seems simpler/easier than https://github.com/enjoy-digital/litex/pull/1528. I'll study it more carefully and see if it also matches other need I have (and will then complement it).

enjoy-digital:
Thanks again for the proposal. The idea makes sense and I reworked it on top of current `master` with a more conservative implementation and regression tests:

 https://github.com/enjoy-digital/litex/pull/2488

The new version keeps the same goal: make generated constants easier to read without changing their value, width, signedness or C type. Small/count-like values remain decimal, while masks, powers of two, aligned/address-like values and clearer large constants are emitted in hexadecimal.

Examples:

  ```verilog
  assign count   = 8'd42;
  assign address = 32'h80000000;
  assign mask    = 16'hffff;

  #define CONFIG_CLOCK_FREQUENCY 100000000
  #define CONFIG_CPU_RESET_ADDR 0x80000000
  #define CONFIG_MASK 0xffff
  #define CONFIG_64BIT 0x100000000ULL
  ```

I also added tests for both generated Verilog and soc.h, including preservation of the existing 64-bit ULL/LL suffix handling.

So I suggest we continue with #2488 and close this PR.

PR #1915, opened 2024-04-04
soc/integration: Ensure the video framebuffer fits in main RAM

davidar:
On targets with less than 20MiB of SDRAM, the default framebuffer region overflows the available address space, leading to errors like this:

```
litex> mem_list
Available memory regions:
ROM                0x00000000 0x20000 
SRAM               0x10000000 0x2000 
SPIFLASH           0x00800000 0x800000 
MAIN_RAM           0x40000000 0x800000 
VIDEO_FRAMEBUFFER  0x40c00000 0x800000 
CSR                0xf0000000 0x10000 

litex> mem_test 0x40000000 0x800000
Memtest at 0x40000000 (8.0MiB)...
  Write: 0x40000000-0x40800000 8.0MiB     
   Read: 0x40000000-0x40800000 8.0MiB     
Memtest OK

litex> mem_test 0x40c00000 0x800000
Memtest at 0x40c00000 (8.0MiB)...
memtest_access error @ 0x40c00000, exiting memtest.
```

This change resizes the framebuffer to occupy no more than half the available space and relocates it to before the end of the main_ram region on such targets.

### DIFF
--- a/test/soc/test_soc.py
+++ b/test/soc/test_soc.py
@@ -17,6 +17,7 @@ from migen import ClockDomain, Record, Signal
 from migen.sim import run_simulation
 
 from litex.soc.cores.hyperbus import HyperRAM
+from litex.soc.cores.video import video_framebuffer_size
 from litex.soc.interconnect import axi, wishbone
 
 from litex.soc.integration.soc import (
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @davidar, this make sense. I'll review it very soon.

enjoy-digital:
Thanks @davidar, this was indeed a real issue.

The actual framebuffer placement logic has since been fixed on `master`: the default framebuffer size is now derived from the selected video mode/pixel format and, when `main_ram` is known, the region is placed near the end of `main_ram` instead of using the old fixed 8 MiB region at `0x40c00000`.

I updated this PR on top of current `master` and kept it as focused regression coverage for the original small-RAM case: https://github.com/enjoy-digital/litex/commit/99fb77a251998d6222282c8b8e64c2afe46f0ad9

The added test covers an 8 MiB `main_ram` setup with the default 800x600 RGB framebuffer and checks that the generated framebuffer region stays inside `main_ram`.

PR #1918, opened 2024-04-07
Reset register for edge catching at the same time as pending register

nrndda:
If trigger source stays the same, no interrupt will arise even when clear is set to 1. Resetting trigger_d register will allow such interrupts arise. Seen such behavior with liteuart: after reset sink.ready always stays at 1 and no interrupt will appear.
Also combine register control under one If/Elif block, which makes more correct verilog code.


enjoy-digital:
Thanks @nrndda, I'll do a closer review very soon. One thing I can see with the updated code is that the IRQ will be lost if the edge appears during the IRQ clear.

nrndda:
Fixed simultaneous interrupt and clear.

nrndda:
I'll close this MR and create another with all changes my changes. 
In this one probably I was wrong. I'll describe shortly.

PR #1919, opened 2024-04-07
Uart tx irq handling fix

nrndda:
This follow RX path. Handle interrupt before releasing it, otherwise will stuck at loop if cannot handle interrupts fast enough.


enjoy-digital:
Thanks @nrndda!

trabucayre:
I have reverted this commit since it break `jtag_uart` (see this [issue](https://github.com/litex-hub/litex-boards/issues/554))

PR #1921, opened 2024-04-09
CSR ordering support in generated files

AndrewD:
Add CSR_ORDERING constants and update access functions. Unchanged behaviour with default csr_ordering="big".

Can use `--csr-ordering little` on command line or `parser.set_defaults(csr_ordering="little")`.

Note that this only affects the addressing of CSR that are 64bit or wider.

This is a proof of concept implementation - only confirmed by inspection so far.

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -199,7 +199,7 @@ def _get_csr_addr(csr_base, addr, with_csr_base_define=True):
     else:
         return f"{hex(csr_base + addr)}L"
 
-def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, read_only, csr_base, with_csr_base_define, with_access_functions):
+def _get_rw_functions_c(reg_name, reg_base, nwords, busword, alignment, ordering, read_only, csr_base, with_csr_base_define, with_access_functions):
     r = ""
 
     addr_str = f"CSR_{reg_name.upper()}_ADDR"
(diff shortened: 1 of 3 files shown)


AndrewD:
This only affects the way csr longer than 32bits work. Currently those csr are ordered big endian on little endian (by default) CPUs like riscv.

For example:

- add a field to the end an existing csr that results in a > 32 bit CSR. The word for the new field is placed first so the binary API of the peripheral changes. This resulted in a bug to track down and is just not intuitive to me: a longer csr should be compatible (but will still shift subsequent csr).
- 64 bit values such as the dma address for sdcard need the words to be written in non-native order resulting in ugly code in Linux/nuttx drivers.

Does this help clarify the intention?

AndrewD:
> updated/rebased/squashed this PR [here](https://github.com/trabucayre/litex/commit/c63e747d3f78fc5f00b2764dc5a5b8e41aabb325).

Thanks! I looked at rebasing this yesterday but decided to leave it until later due to the conflicts.

AndrewD:
Also note that by default this has no effect on the generated gateware, but it will be require driver changes if the csr_ordering is changed.
I have a separate change that adds the correct information to device tree so the Linux SD card driver could support both options.

enjoy-digital:
Thanks, this was useful. Most of the original CSR-ordering support is now already covered by the later merged work in #2270: generated CSR accessors, imported CSR JSON, BIOS/common helpers and the  `--csr-ordering` option now handle `CONFIG_CSR_ORDERING_BIG/LITTLE`.

I reworked the remaining useful LiteSDCard DTS metadata part here: https://github.com/enjoy-digital/litex/pull/2489

This adds the generated Linux DTS `big-endian;` / `little-endian;` property for the LiteSDCard node, derived from the existing CSR-ordering constants. The default remains the historical big ordering, and the change is generator-only.

I also added focused tests for default big ordering, explicit little ordering, and conflicting generated constants.

A matching Linux driver change is still needed if a non-default CSR ordering is used, since the DTS metadata alone does not change driver behavior.

PR #1928, opened 2024-04-16
Add bios disassembly file generation rules.

nrndda:
Added bios disassembly file generation. Very useful for debugging.

### DIFF
--- a/litex/soc/software/bios/Makefile
+++ b/litex/soc/software/bios/Makefile
@@ -50,7 +50,7 @@ else
 LSCRIPT = linker.ld
 endif
 
-all: bios.bin
+all: bios.bin bios.dis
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks, this is useful for debugging BIOS/early boot issues.

I updated the PR a bit with a follow-up commit: 1edc341f5.

The disassembly generation is now opt-in instead of being part of the default BIOS build. This avoids generating an extra file on every normal build/CI run, while still making it easy to use when needed.

After building the BIOS, the user can generate the disassembly from the BIOS build directory with:

  ```sh
  make dis
  ```

  or:

  make bios.dis

This creates bios.dis from bios.elf using the target objdump.

PR #1937, opened 2024-04-22
tools/litex_term: fix DeprecationWarning

AndrewD:
litex/litex/tools/litex_term.py:557: DeprecationWarning: setDaemon() is deprecated, set the daemon attribute instead
  self.reader_thread.setDaemon(True)
litex/litex/tools/litex_term.py:586: DeprecationWarning: setDaemon() is deprecated, set the daemon attribute instead
  self.writer_thread.setDaemon(True)


enjoy-digital:
Thanks @AndrewD!

PR #1938, opened 2024-04-22
soc/export: add guard for access_functions

AndrewD:
Disable access functions with -DLITEX_ACCESS_FUNCTIONS=0 when compiling. Useful when using litex generated headers in an environment where the access functions generate warnings and are not used.

### DIFF
--- a/litex/soc/integration/export.py
+++ b/litex/soc/integration/export.py
@@ -187,7 +187,12 @@ def get_soc_header(constants, with_access_functions=True):
 
     if with_access_functions:
         r += "\n#ifndef __ASSEMBLER__\n"
+        r += "#ifndef LITEX_ACCESS_FUNCTIONS\n"
+        r += "#define LITEX_ACCESS_FUNCTIONS 1\n"
+        r += "#endif\n"
+        r += "#if LITEX_ACCESS_FUNCTIONS\n"
         r += funcs
+        r += "#endif // !LITEX_ACCESS_FUNCTIONS\n"
         r += "#endif // !__ASSEMBLER__\n"
 
     r += "\n#endif\n"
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Hi @AndrewD,

sorry I reverted for now with https://github.com/enjoy-digital/litex/commit/fd477703db13cce1020d496497c0f044a21230e8 since I first wanted to do a closer review and discuss it. W

hile I understand the use case, this add a bit of complexity to code that is already complex to maintain and we also already have the possibility to generate the header with or without the access functions. What I generally do when I don't want to want to have the access funtions is to generate the header without them with the parameter already available. Could this also work for you? If you need the two versions, we could also enable generation of the two versions with different suffixes.

enjoy-digital:
OK I see, thanks for the PR. I'll have a closer look and think about it.

AndrewD:
> OK I see, thanks for the PR. I'll have a closer look and think about it.

One alternative implementation would be to move the generated functions to a separate header: csr_access.h or similar and only include that header when desired.

enjoy-digital:
@AndrewD, this indeed seems a better approach. I was off this week but will look at this next week.

enjoy-digital:
Hi @AndrewD,

with https://github.com/enjoy-digital/litex/pull/1953, we are no longer generating by default the field accessors (that are, from what I remember, only used by a few projects). So by default we now only have the CSR definitions and simple access functions. Could you check if you still have warnings with this? If so, I could continue and try to split things on different files, but if not, we could just keep things like that.

AndrewD:
Hi @enjoy-digital,

Unfortunately this cleanup doesn't cater for the case I was solving: generate `csr.h` once and:

1. use in the bios => requires access functions
2. use in nuttx => doesn't use access functions

My change provided a predecessor define to allow us to selectively disable access functions with `-DLITEX_ACCESS_FUNCTIONS=0` when compiling `nuttx` , otherwise we need to provide a dummy `system.h` header which defines `csr_read_simple()` and `csr_write_simple()` and other tweaks to suppress warnings from the unused access functions in the `csr.h`.

enjoy-digital:
Thanks for the feedback @AndrewD, I'll look at it today or tomorrow.

enjoy-digital:
@AndrewD: This should now be good with https://github.com/enjoy-digital/litex/commit/4502edd33e9cad3347b0599f44fad467d7c5ca1f, https://github.com/enjoy-digital/litex/commit/5b297f5601ac106c177c904593645ce191f2b31f and https://github.com/enjoy-digital/litex/commit/05030990b28ef4faef3f49b19c50cec3fb5c18c5. This should be almost functionally almost identical to your solution but avoid too much complexity in generated code and should still allow a human to look at it.

There's maybe the `system.h` include that you'll want to get rid of: We could make this include optional.

PR #1939, opened 2024-04-24
yosys+nextpnr: fix error message

hansfbaier:
This was supposed to be a format string


trabucayre:
Applied. Thanks @hansfbaier

PR #1942, opened 2024-04-26
Target hooks

AndrewD:
This allows the target to supply custom target_init() and target_boot() functions, via a library linked with the bios. eg:

```
    # add custom target specific library to the bios:
    src="libtarget"
    src_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), src))
    builder.add_software_package(src, src_dir)
    builder.add_software_library(src)
```

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -580,7 +580,7 @@ void netboot(int nb_params, char **params)
 /* Flash Boot                                                            */
 /*-----------------------------------------------------------------------*/
 
-#ifdef FLASH_BOOT_ADDRESS
+#if defined(FLASH_BOOT_ADDRESS) || defined(MAIN_RAM_BASE)
 
 static unsigned int check_image_in_flash(unsigned int base_address)
 {
@@ -603,9 +603,10 @@ static unsigned int check_image_in_flash(unsigned int base_address)
 
 	return length;
 }
+#endif
 
-#if defined(MAIN_RAM_BASE) && defined(FLASH_BOOT_ADDRESS)
-static int copy_image_from_flash_to_ram(unsigned int flash_address, unsigned long ram_address)
+#if defined(MAIN_RAM_BASE)
+int copy_image_from_flash_to_ram(unsigned int flash_address, unsigned long ram_address)
 {
 	uint32_t length;
 	uint32_t offset;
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks @AndrewD, this looks good. I'll review it soon.

litex/soc/software/bios/boot.h
```diff
@@ -12,5 +12,8 @@ void flashboot(void);
 void romboot(void);
 void sdcardboot(void);
 void sataboot(void);
+extern void target_init(void) __attribute__((weak));
+extern void target_boot(void) __attribute__((weak));
```
trabucayre, line 16: why not something like:
```
void target_nop() {}
void target_init(void) __attribute__((weak, alias("target_nop")));
void target_boot(void) __attribute__((weak, alias("target_nop")));
```
to avoid having to test if function exist

litex/soc/software/bios/boot.c
```diff
@@ -631,6 +633,7 @@ static int copy_image_from_flash_to_ram(unsigned int flash_address, unsigned lon
 }
 #endif
 
+#if defined(FLASH_BOOT_ADDRESS)
```
trabucayre, line 636: I don't see `#endif` for this `#if` ?

trabucayre:
For `target_boot` and `target_init` I think its a good idea to allows target/specific libraries to do some tasks, at some points, without having to modify existing code.
But for `check_image_in_flash` and `copy_image_from_flash_to_ram` its less clear. Is it to allows custom code to check and load firmware for a custom boot mode more or less similar to `flashboot()` ?

AndrewD:
> But for `check_image_in_flash` and `copy_image_from_flash_to_ram` its less clear. Is it to allows custom code to check and load firmware for a custom boot mode more or less similar to `flashboot()` ?

Yes: for example a target_boot() that can validate a dtd, copy a kernel and copy then run opensbi is one such usage.

AndrewD:
I would also like to rename for `check_image_in_flash` and `copy_image_from_flash_to_ram` but didn't to minimise the diff

litex/soc/software/bios/boot.h
AndrewD: Yes that's another alternative I onsidered.I like the fact that the test makes it's clear that it's optional, but that's personal preference.
There doesn't appear to have been any goal to minimise code size in the bios so i opted for clarity.

PR #1943, opened 2024-04-26
Fix false path clock constraints

nrndda:
If two clock already included in false path list then swapped tuple pair of clocks won't be added. Fix this by adding temporary variable with tuple

### DIFF
--- a/litex/build/generic_toolchain.py
+++ b/litex/build/generic_toolchain.py
@@ -180,5 +180,6 @@ class GenericToolchain:
         if keep:
             from_.attr.add("keep")
             to.attr.add("keep")
-        if (to, from_) not in self.false_paths:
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks, this was pointing at a real issue: false-path constraints can be directional, so `(from, to)` and `(to, from)` must not be collapsed in the generic storage path.

This has since been fixed on `master` with enjoy-digital/litex@94f816d6b9830ef5f8908edfecd6b5f59674aa15, and later simplified further by keeping the generic toolchain storage as:

  ```python
  self.false_paths.add((from_, to))
  ```

So both directions are now preserved when they are explicitly added. The PR branch is now conflicted/obsolete, and the same issue is already covered in current LiteX. I think we can close this PR.

PR #1944, opened 2024-04-28
When part skipped do not threat it as sent

nrndda:
Do not mask skipped part of transaction. Otherwise will stack without slave.aw.valid and slave.w.valid.


enjoy-digital:
Thanks, this has now been reworked on top of current master.

The PR keeps the original fix, but the implementation has been cleaned up so the AXI-Lite down-converter only considers a skipped narrow write beat complete on real `valid & ready` handshakes, instead of treating an idle high `ready` as progress.

I also added a regression test covering the failing case: a 32-bit to 16-bit AXI-Lite down-conversion where the low halfword is skipped by `strb=0b1100` and the slave keeps `aw.ready`/`w.ready` asserted while idle.

PR #1946, opened 2024-05-02
Revert LitexModule for AXILiteSRAM as well. 

nrndda:
Follows revert d021564fca8ac6d239dad2b18b5531cc9ee249c6 for wishbone.

Original bug is within multi-drive of some memories.  At least sdram is twice assigned from different always blocks. This happens because in csr_bus.py function CSRBankArray.scan scans all objects with get_memories attribute, which now present in every LiteXModule instance. Thus making some memories appear in slave regions of interconnect and in CSR banks! Don't know how to fix this duplicate correctly but at least revert to allow AXI-Lite as top interconnect.

### DIFF
--- a/litex/soc/interconnect/axi/axi_lite.py
+++ b/litex/soc/interconnect/axi/axi_lite.py
@@ -218,7 +218,7 @@ def axi_lite_to_simple(axi_lite, port_adr, port_dat_r, port_dat_w=None, port_we=
 
 # AXI-Lite SRAM ------------------------------------------------------------------------------------
 
-class AXILiteSRAM(LiteXModule):
+class AXILiteSRAM(Module):
     def __init__(self, mem_or_size, read_only=None, init=None, bus=None, name=None):
         if bus is None:
             bus = AXILiteInterface()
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @nrndda, I'll look at it.

Dolu1990:
Hi,

This change breaks VexiiRiscv peripheral accesses in some subtil ways (sdcard DMA goes crazy)
The issue only start to appear during the linux boot when it start to use sdcard (it seems).
--cpu-type=vexiiriscv --bus-standard axi-lite 
I do not have a way to reproduce the issue in simulation.

I had a exactly the same issue somehow by the past when i was not using --bus-standard axi-lite XD

Not sure how we should operate from here.

enjoy-digital:
Thanks @Dolu1990, I'll have a look at this, hopefully tomorrow.

Dolu1990:
I have to say, this is a tricky bug, the bios is fine with the sdcard, but things turn wrong later on with the linux driver.

nrndda:
This should be reverted and LiteXModule should be fixes for SRAM functions instead. I observed multidrive for sram, but if it's only me, this could be some weirdness in my core.py.

enjoy-digital:
The issue with LiteXModule has been fixed with https://github.com/enjoy-digital/litex/commit/fbf03ec74c7d3c941e4db01f02a37ff7d58651fe, thanks for the analysis @nrndda.

As @Dolu1990 figured out, his issue was in fact related to the additional CSRs that were moving the mapping and required a different `.dts/.dtb` file.

PR #1948, opened 2024-05-04
Add missing 32-bit variants of RISC-V target triples

Liamolucko:
I had to waste another 10-15 minutes building a `riscv64-none-elf` GCC instead after LiteX didn't like my `riscv32-none-elf` one, so I thought I'd quickly add it. (I was using that target triple because it's what Nix's `pkgsCross.riscv32-embedded` package set uses).

I also added 32-bit variants of the rest of the target triples which didn't have them.


enjoy-digital:
Thanks @Liamolucko!

PR #1949, opened 2024-05-07
The case when AWVALID and WVALID are not asserted at the same cycle

alexey-morozov:
This is a fix for the case when AWVALID and WVALID signals are not asserted at the same clock cycle (a valid case according to the AXI/ACE protocol specification, also for the Lite version). The problem was discovered with the use of Xilinx AXI Interconnect IP which asserts WVALID one cycle later after AWVALID assertion.


FlyGoat:
I'm also hit by this problem on a proprietary hard core.
@enjoy-digital Is it possible to get this merged?

Thanks

enjoy-digital:
Thanks @alexey-morozov and @FlyGoat for the feedback. This is merged.

PR #1955, opened 2024-05-16
cores/cpu/naxriscv: fix 64 bits IRQ support

Dolu1990:
The IRQ routine for NaxRiscv was only supporting 32 bits cpu. Now it is ok for 64 bits aswell.


enjoy-digital:
Thanks @Dolu1990!

PR #1956, opened 2024-05-17
Zynqmp aximaster eth i2c uart

trabucayre:
This serie address a some improvement / new features for zynqMP:
- a new method `set_preset` to fix the *PSU* default with a tcl script (from Vivado bd)
- a rework to the `add_axi_gp_master` method, similar to zynq7000 way. All signals are explicitly connected instead of using a layout
- zynqMP contains peripherals (ethernet, i2c, uart, ...) able to be connected to PS_MMIO pins (dedicated to PSU) or available at PL side (EMIO mode). This PR contains a serie of commit to enable to activate and route ethernet, i2c, and uart to the logic


enjoy-digital:
Thanks @trabucayre!

PR #1958, opened 2024-05-18
Simple mistake fix

nrndda:


trabucayre:
Applied. Thanks @nrndda

PR #1959, opened 2024-05-21
Add dynamic mux for UARTBone / normal UART

jwise:
One problem with debugging a single-UART system is that the system will always block if a UARTBone client is not connected.  This is particularly annoying if, for instance, your vendor's RAM update tools do not work right (I'm looking at you, Efinix...), and you need the UARTBone to iterate on firmware -- but, once you have a build, you would like to have a UART that does not block.  (In my system, I have two FPGAs talking to each other, and I am probably developing one or the other at a time, and I would like to be able to distribute a bitstream to the customer that just happily boots without any drama, for instance.)

This PR solves this by modifying the `UARTCrossover` to have a `debug_en` CSR, and a `pad_phy` input.  When `debug_en` is turned on (the default, on reset), then the system behaves exactly like it previously did.  However, switching off `debug_en` means that the crossover internal UART and UARTbone are disconnected, and the `UARTCrossover`'s `source` and `sink` are actually connected to the pads of the outer world -- basically, passing the UART straight through.

To use this, I do something like this:

```c
int main(void)
{
#ifdef CONFIG_CPU_HAS_INTERRUPT
        irq_setmask(0);
        irq_setie(1);
#endif
        // As soon as the CPU is running, before we try to write to the UART
        // (and potentially block the system), then we disable uart_debug_en.
        // Now, just connect to the system with picocom to get a console.
        uart_debug_en_write(0);
/* ... system setup ... */

/* ... main loop ... */
        str = readstr();
        if(str == NULL) return;
        token = get_token(&str);
/* ... */
        // Have a CLI command to reenable the uartbone.  As soon as this
        // command is issued, you will need to start up a litex-server and
        // litex-term.
        else if(strcmp(token, "debug") == 0) 
                uart_debug_en_write(1);
```

There is a little bit of WIP here: the simplest way to wire this up was through the old `crossover+uartbone` path that is now deprecated.  I undeprecate this by commenting out the deprecation!  But obviously I would like feedback on what you think the better way is to do this.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -474,12 +474,53 @@ class UARTCrossover(UART):
 
     Creates a fully compatible UART that can be used by the CPU as a regular UART and adds a second
     UART, cross-connected to the main one to allow terminal emulation over a Wishbone bridge.
+
+    If a pad_phy is provided, a debug_en register is created that enables a
+    mux to interpose the pads.  When debug_en is 0, then the internal UART
+    is routed to the pads (and the crossover UART is unused); when debug_en
+    is 1, then the pad_phy is passed through to its original destination,
+    and the crossover is also passed through as usual (i.e., the behavior as
+    if no pad_phy were provided).
     """
-    def __init__(self, **kwargs):
+    def __init__(self, pad_phy=None, **kwargs):
         assert kwargs.get("phy", None) == None
         UART.__init__(self, **kwargs)
         self.xover = UART(tx_fifo_depth=1, rx_fifo_depth=16, rx_fifo_rx_we=True)
-        self.comb += [
-            self.source.connect(self.xover.sink),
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Thanks again for the original proposal and detailed explanation.

I opened a refreshed implementation in #2506 based on the same UARTBone/normal UART mux idea. Compared with this PR, the new version keeps the user-facing path simpler:

- the mux is selected automatically when UARTBone and the regular serial/crossover UART are both enabled;
- the mux resets to UARTBone mode;
- the UART `uartbone` CSR selects UARTBone vs regular UART mode;
- the BIOS now has a `uartbone` command to switch back to UARTBone from the regular UART console;
- focused tests were added for the mux routing and SoC integration.

The commit message also credits the original idea here. If this direction looks good, we can continue from #2506 as the current integration point.

PR #1962, opened 2024-05-23
Add support for the Efinix reconfiguration interface

m-byte:
This adds low level support for the reconfiguration interface (sometimes also called remote update by Efinix)


enjoy-digital (approved):
Thanks @m-byte.

PR #1968, opened 2024-05-28
liblitespi: Fix #1967

### LINKED ISSUE #1967: bios: #1953 breaks bios when spiflash is configured "with_master=True"
Since #1953 (or more specifically https://github.com/enjoy-digital/litex/pull/1953/commits/46911d5078b81038dceb010b0b786bcb07a7a4d1), bios can't be built when spiflash is configured as "with_master = True". This is due to the code in libspilite assuming that the csr functions are available.

I see three options to resolve this:
a) revert https://github.com/enjoy-digital/litex/pull/1953/commits/46911d5078b81038dceb010b0b786bcb07a7a4d1
b) modify builder.py so that with_fields_access_functions is automatically set to true when at least one spiflash is configured as with_master
c) rewrite libspilite to not make use of these convenience functions

@enjoy-digital

m-byte:
Make liblitespi independent from field_access_functions, since they were removed in 46911d5078b81038dceb010b0b786bcb07a7a4d1

### DIFF
--- a/litex/soc/software/liblitespi/spiflash.c
+++ b/litex/soc/software/liblitespi/spiflash.c
@@ -74,23 +74,42 @@ void spiflash_dummy_bits_setup(unsigned int dummy_bits)
 
 #ifdef CSR_SPIFLASH_CORE_MASTER_CS_ADDR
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @m-byte.

PR #1971, opened 2024-05-29
efinix: add synthesis options to a dictionary in the EfinityToolchain

jwise:
Sometimes the synthesis tool needs some help to successfully map a design, in the form of different synthesis options.

In my particular case, this didn't actually save me for my design, but it would be nice to have it available in the future if I need it again.  I use this in my `Platform` like this:

```python
class Platform(EfinixPlatform):
    default_clk_name   = "clk50"
    default_clk_period = 1e9/50e6

    def __init__(self, toolchain="efinity"):
        EfinixPlatform.__init__(self, "T20F256C4", _io, iobank_info=_bank_info, toolchain=toolchain)
        self.toolchain.map_opts['mode'] = 'area'
```

### DIFF
--- a/litex/build/efinix/efinity.py
+++ b/litex/build/efinix/efinity.py
@@ -12,6 +12,7 @@ import sys
 import site
 import inspect
 import datetime
+import itertools
 
 from xml.dom import expatbuilder
 import xml.etree.ElementTree as et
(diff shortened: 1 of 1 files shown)


trabucayre:
Hi,
Why not implementing `efinity_args` and `efinity_argdict` to allows users to specify/modify these parameters directly from the command line? See [trellis toolchain](https://github.com/enjoy-digital/litex/blob/master/litex/build/lattice/trellis.py#L158) and associated call from [the platform](https://github.com/enjoy-digital/litex/blob/master/litex/build/lattice/platform.py#L59)?

Also one remark: using a dict and itertools allows to have a code more compact but its increase difficulties to read/update/maintain code.

enjoy-digital:
Thanks @jwise, I agree with @trabucayre about using something closer to what is already present on other platforms (and to favor readability vs code compactness).

trabucayre:
Hi @jwise.
I have written this [PR](https://github.com/enjoy-digital/litex/pull/2009), with remarks.
With this one it's possible to change synthesis parameters directly from the CLI at build time instead of having to modify the target or the platform.
It also keep the original structure, more easy to read/maintain.

Thanks

enjoy-digital:
Thanks @jwise, @trabucayre. #2009 has been merged so we can close.

PR #1974, opened 2024-05-30
DTS zephyr updates

AndrewD:
This PR improves the support for indentation so those functions can be reused to generate a nicer Linux DTS. See #1960.

It also changes the formatting of the reg in dts_reg() to be dts schema compliant but I haven't tested if the old format is mandated by zephyr. Can someone please confirm, maybe @maass-hamburg or @trabucayre?

PR #1973, opened 2024-05-30
Dts schema compliance

AndrewD:
This fixes several linux schema validation issues. Closes #1905, see #1960

This branch is on top of the branch for #1972 but that is not a dependency.

PR #1975, opened 2024-05-30
DTS target integration

AndrewD:
This set of changes follows on from #1972, #1973, #1974 and adds a mechanism to specify DTS information for a core from anywhere within litex.

With these changes we can generate a valid DTS for standard litex cores and custom cores without needing to hard code any support into litex_json2dts_linux.py and implements most of the concepts detailed in #1960.

@enjoy-digital any feedback on style or details would be appreciated.

### DIFF
--- a/litex/gen/common.py
+++ b/litex/gen/common.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 from migen import *
+from typing import Union, List
 
 # Coloring Helpers ---------------------------------------------------------------------------------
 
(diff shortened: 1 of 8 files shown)


enjoy-digital:
Thanks @AndrewD, I'll have to think about the global approach for this. We also have been thinking with @trabucayre  about switching to a python lib ease generation of the dts files, which could also help here.

AndrewD:
Hi @enjoy-digital,
I didn't find a python lib for dts generation. All I found was libraries to fixup a DTS (a lib from zephyr) and a few others with similar scope.

Ultimately I needed very little that is DTS specific and most was already in json2dit_zephyr so it seemed easiest to  follow the pattern and leverage that.

trabucayre:
Hi @AndrewD I have tried to use [pylibfdt](https://pypi.org/project/pylibfdt/). I'm able to produces a valid dtb/dts but I haven't found how to attach a label for example.

AndrewD:
@enjoy-digital I've removed the python 3.10 dependency. I did notice that the [litex:setup.py](https://github.com/enjoy-digital/litex/blob/329bd36f7f3659a26c8f1fb1a77e535d4fb42af6/setup.py#L23) specifies python >= 3.7 but the CI build is with 3.9?

enjoy-digital:
Thanks @AndrewD, I'll have a closer look at the changes but haven't had the time to think more about it yet (In the past, I merged some features in the project because I was not able to closely review things and to avoid frustrating developers, but this also created lots of headaches for me later to undo some things or do things differently to allow the project to go further). For this change, I first prefer that we have a clear vision of the direction before we start doing changes.

I would first like to rework the way .dts files are generated (with helper code or a python lib), then think about the mechanism to collect information and generate the .dts information to the .json file and then how to reuse this information to generate the .dts dynamically. If we apply this approach to the peripherals, we should also probably apply it to the CPUs. Let me think about this, discuss with @trabucayre internally and we'll get back to you.

AndrewD:
@enjoy-digital we are actively using this and have made further refinements for our own usage. I've pushed an update of the current state that has been squashed down to a few commits (without the initial experiments). Do you have any further thoughts yet?

While the impact on Litex is isolated this is now quite solid and could replace most of the hard coded nodes for modules in json2dts_linux: json2dts_zephyr could also leverage this.

enjoy-digital:
Thanks @AndrewD. We haven't yet been able to fully think about the right approach (busy with other things and other tasks we are working on together). We'll definitly spend more time on this in September and will find a solution that will satisfy both of us. Sorry for being slow on this.

PR #1978, opened 2024-06-03
soc/integration/builder: use output_dir for csr

AndrewD:
Currently `csr` files are always generated in the current working directory. This changes the behaviour to ensure all output from the build is relative to `output_dir` if it is set.

Otherwise the existing behaviour is preserved.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -109,9 +109,9 @@ class Builder:
         self.build_backend    = build_backend
 
         # Exports.
-        self.csr_csv  = csr_csv
-        self.csr_json = csr_json
-        self.csr_svd  = csr_svd
+        self.csr_csv  = os.path.join(self.output_dir, csr_csv)  if output_dir and csr_csv  else csr_csv
+        self.csr_json = os.path.join(self.output_dir, csr_json) if output_dir and csr_json else csr_json
+        self.csr_svd  = os.path.join(self.output_dir, csr_svd)  if output_dir and csr_svd  else csr_svd
         self.memory_x = memory_x
 
         # BIOS.


litex/soc/integration/builder.py
```diff
@@ -109,9 +109,9 @@ def __init__(self, soc,
         self.build_backend    = build_backend
 
         # Exports.
-        self.csr_csv  = csr_csv
-        self.csr_json = csr_json
-        self.csr_svd  = csr_svd
+        self.csr_csv  = os.path.join(self.output_dir, csr_csv)  if output_dir and csr_csv  else csr_csv
```
gsomlo, line 112: I like the idea of this patch, but when I tested it, it doesn't seem to work without *explicitly* specifying `--output-dir` as well. The output dir defaults to `./build/<boardname>/` (with further `software/` and `gateware/` subdirectories beneath that) in the absence of an explicit `--output-dir` value.

Am I missing something? If not, would be nice if this worked w/o requiring an explicit output dir specified on the build command line...

trabucayre:
One thing i doesn't understand: if you use `--csr-csv=csr.csv` I'm agree the file is written in the current directory, But if you uses `--csr-csv=/tmp/csr.csv` instead, the file is present in */tmp* directory:
```bash
$ date && ./digilent_arty.py --build --no-compile --csr-csv=/tmp/csr.csv
Tue Jun  4 04:59:55 PM CEST 2024
[...]
$ ls -l /tmp/csr.csv
-rw-r--r-- 1 gwe gwe 4412 Jun  4 16:59 /tmp/csr.csv

gsomlo:
On Tue, Jun 04, 2024 at 08:02:51AM -0700, Gwenhael Goavec-Merou wrote:
> One thing i doesn't understand: if you use --csr-csv=csr.csv I'm agree the file
> is written in the current directory, But if you uses --csr-csv=/tmp/csr.csv
> instead, the file is present in /tmp directory:

Right. So then implementing this change correctly gets a bit *more*
complex :)

1. If --csr-csv=filename.csv, then put it in the final resolved value
of the gateware dir (or in the output dir, which containes the
gateware and software directories, I'm really OK with either :)

2. If --csr-csv=path/to/filename.csv, then keep the current behavior
(if path is not fully specified, i.e., doesn't start with `/`, then
it should be relative to the *current* directory, at least IMHO, to
minimize the "surprise factor" of it ending up "somewhere else").

trabucayre:
In fact I don't understand because solution seems redundant to default behavior:
- currently user may provides a relative or absolute path for files (I'm agree the path must be provided for all files)
- with this PR the user must provides a relative or absolute path using extra argument.

gsomlo:
On Tue, Jun 04, 2024 at 12:09:53PM -0700, Gwenhael Goavec-Merou wrote:
> In fact I don't understand because solution seems redundant to default
> behavior:
> 
>   • currently user may provides a relative or absolute path for files (I'm
>     agree the path must be provided for all files)

When there is *no* path provided, the current assumption is "./"
(i.e., current directory, that's different from
./build/<boardname>/... where all other artifacts are placed).

It looked (to me) that this is what the PR is meaning to address, and,
if that's indeed the case, I agree with the overall goal :)

>   • with this PR the user must provides a relative or absolute path using extra
>     argument.

AndrewD:
> When there is *no* path provided, the current assumption is "./" (i.e., current directory, that's different from ./build/<boardname>/... where all other artifacts are placed).

Correct, that's the goal. The logic just had to be a bit convoluted to maintain existing behaviour but use output_dir when it is provided.

enjoy-digital:
Hi,

I addressed the issue a bit differently with https://github.com/enjoy-digital/litex/commit/8d8dd117b6d734657f3f3acddb8c39712f1f6fa6. Now by default all the the exports are generated to output_dir directory with default name, unless user explicitly provide an explicit export.

This should address @AndrewD's initial need and could also the common case where export has not be forgotten and build command need to be re-executed with it.

PR #1983, opened 2024-06-10
linux dts: add vexii clint support

Dolu1990:
With this change, the DTS generated when using VexiiRiscv is good out of the box and can be used with upstream generic opensbi

### DIFF
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -328,7 +328,7 @@ def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, root_devic
 
     # Interrupt Controller -------------------------------------------------------------------------
 
-    if (cpu_arch == "riscv") and ("rocket" in cpu_name):
+    if (cpu_arch == "riscv") and ("rocket" in cpu_name or "vexiiriscv" in cpu_name):
         # FIXME  : L4 definitiion?
         # CHECKME: interrupts-extended.
         dts += """


litex/tools/litex_json2dts_linux.py
```diff
@@ -328,7 +328,7 @@ def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, root_devic
 
     # Interrupt Controller -------------------------------------------------------------------------
 
-    if (cpu_arch == "riscv") and ("rocket" in cpu_name):
+    if (cpu_arch == "riscv") and ("rocket" in cpu_name or "vexiiriscv" in cpu_name):
```
trabucayre, line 331: Why not using:
```python
if (cpu_arch == "riscv") and (cpu_name in ["rocket",  vexiiriscv"]):
```

Dolu1990: i was afraid of breaking some weird stuff as : 
"rocket" in cpu_name isn't equivalent to cpu_name in ["rocket"]
As there may eventualy be a "rocket_on the moon" as cpu_name ?

So i prefered not disrupting the old behaviour. As they could have done it with cpu_name == "rocket" but didn't XD

trabucayre: make sense but `cpu_name` is used with `cpu_architecture` so if there is a "rocket_somewhere" a keyError will be produces. :)

gsomlo:
> make sense but cpu_name is used with cpu_architecture so if there is a
> "rocket_somewhere" a keyError will be produces. :)

Does cpu_name get populated from the name of the directory under
`litex/soc/cores/cpu/` ?

If so, the folder is named `rocket` and we could use `==` instead of
`"string" in cpu_name` -- I have no memory of how/why that got
started.

IOW, is there *currently* a possibility that cpu_name will be
"[some-optional-prefix]rocket[some-optional-trailer]" and not just
simply "rocket" ?

If not, then we should probably

	`s/if "rocket" in cpu_name/if cpu_name == "rocket"/g`

and make life simpler :)

litex/tools/litex_json2dts_linux.py
Dolu1990: I pushed as you proposed just now ^^

Dolu1990:
> Does cpu_name get populated from the name of the directory under
`litex/soc/cores/cpu/` ?

To me it get named from 

```
class Rocket(CPU):
    category             = "softcore"
    family               = "riscv"
    name                 = "rocket" # <---- this 
```

Dolu1990:
I checked, it comes from core.py "name                 = "rocket" # <---- this "

So, i did the `s/if "rocket" in cpu_name/if cpu_name == "rocket"/g` and pushed it to this PR.

I kinda need this to be merged before https://wiki.f-si.org/index.php/FSiC2024, (before next week) as it is the last thing which prevent out of the box exployment for Vexii.
Then this "how to try debian on vexii" would work out of the box : 
https://github.com/SpinalHDL/VexiiRiscv/tree/dev/doc/litex/debian

Let's me know if anything prevent the merge :)

trabucayre:
Applied thanks @Dolu1990 !

Dolu1990:
Thanks :)

PR #1984, opened 2024-06-11
litex_json2renode.py: add option for elf bios file and correct vexriscv variants

maass-hamburg:
- add option for elf bios file
- correct vexriscv variants

for the linux variant I also removed `'privilegeArchitecture: PrivilegeArchitecture.Priv1_10'` as that is already the default value in https://github.com/renode/renode-infrastructure/blob/5904a437c60ea376b679e8de92b71b29a62c8f8a/src/Emulator/Cores/RiscV/VexRiscv.cs#L20


enjoy-digital:
Thanks @maass-hamburg.

PR #1991, opened 2024-06-17
Add json excludes

AndrewD:
In add_json() the interrupt numbers from a downstream SOC are not relevant to the main SOC so filter them out by default.


enjoy-digital:
Thanks @AndrewD, this is merged (thanks also for the origin/name swap fix).

PR #1992, opened 2024-06-17
integration/export: Fix MockCSRRegion base definition.

AndrewD:
MockCSR in the downstream SOC are not related to csr_base in the main SOC. This fixes the addresses in generated csr files.


enjoy-digital:
Thanks @AndrewD, this is merged.

PR #1994, opened 2024-06-18
soc/cores/cpu/zynqmp/core.py: added csr into mem_map, added M_AXI_HPM0_FPD by default

trabucayre:
By default and unlike others CPUs, for zynq (7000 and MP) based board, CSR and corresponding bus  are added at target level.
Also, for these CPU family, a converter between AXI and wishbone is added: this is no more required: LiteX is now able to convert between native bus type and `bus_standard`.

This PR has for goal to change this behavior: csr base address is now directly added at CPU level and the corresponding bus is added in peripheral_buses: no more to add these hook at board level.

Targets update is done with this [PR](https://github.com/litex-hub/litex-boards/pull/590).


enjoy-digital:

PR #1995, opened 2024-06-19
soc: add_spi_master: make spi_clk_freq an int

maass-hamburg:
convert spi_clk_freq to an int. this way the constant is also an int.

this is important as that constant is used in the dts converters and there it should not end with a `.0`


enjoy-digital:
Thanks @maass-hamburg.

PR #1997, opened 2024-06-22
integration/soc: data_width_convert: Inherit more bus properties

FlyGoat:
For data_width converter we need to ensure that adapted interface have same properties as it's parent interface, so that id and user signals for AXI will be propagated properly from parent to adapted interface.


enjoy-digital:

PR #1990, opened 2024-06-17
Initial MIPS CPUs support

FlyGoat:
Hi,

This PR implemented initial support for MIPS based CPUs, it implemented `bios_map` hook for CPU to deal with problem mentioned in #1903, fixed AHB bus handling then added two CPUs as examples.

Please review.

Thanks

PR #1998, opened 2024-06-23
soc/integration/soc.py: Fix creation of AHB2Wishbone bridge

FlyGoat:
Don't do bus_addressing_convert as it's being handled in AHB2Wishbone logic.

Add addressing parameters for AHBInterface constructor as required by soc code.

This is split from #1990, I would like to get those independent features merged first.


enjoy-digital:
Thanks @FlyGoat!

PR #1999, opened 2024-06-23
csr_bus: Honour re signal from the upstream bus

FlyGoat:
Currently CSR bus assumed that ~we means reading, that created a problem that when for a CSR if reading has side effects and adr parked unintentionally at that CSR, the reading side effect will be triggered.

For SoCs, this happened when upstream bus issued a write transaction with wishbone.sel, then on CSR bus it will be translated as adr = addr, we = 0, which will be interpreted as a read to such address, and trigger undesired side effect for such CSR.

Such upstream transaction will be generated by our bus width converter.

Given that we signal already presents in CSR Interface, the easiest way to handle such situation is to generate re signal at bus bridges and propagate it all the way down to the Interface.

### DIFF
--- a/litex/soc/interconnect/axi/axi_lite.py
+++ b/litex/soc/interconnect/axi/axi_lite.py
@@ -148,7 +148,7 @@ class AXILiteRemapper(LiteXModule):
 
 # AXI-Lite to Simple Bus ---------------------------------------------------------------------------
 
-def axi_lite_to_simple(axi_lite, port_adr, port_dat_r, port_dat_w=None, port_we=None):
+def axi_lite_to_simple(axi_lite, port_adr, port_dat_r, port_dat_w=None, port_re=None, port_we=None):
     """Connection of AXILite to simple bus with 1-cycle latency, such as CSR bus or Memory port"""
     bus_data_width = axi_lite.data_width
     adr_shift      = log2_int(bus_data_width//8)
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks @FlyGoat, the CSR bus is indeed too simple and not having a validity signal for the reads can cause these side effects. (We tried to avoid actions on register read for this, but that's indeed no longer the case in some modules) so let's merge your PR with the `re` additional signal.

m-byte:
@FlyGoat @enjoy-digital it seems like this PR broke a few things (e.g. crossover uart), since quite a few current cores assume that the we signal could be used for this purpose (this introduces the re signal on the CSR bus which wasn't there before). Even though I agree that this is a more sensible approach, it is quite literally a breaking change and I feel like it should have been clearly communicated and at least some effort should have been made to at least check compatibility with the standard litex cores.

enjoy-digital:
Thanks @m-byte for the feedback, I'll be back at work tomorrow and will look at it.

FlyGoat:
Sorry, I only experimented with my daily configurations using AXI/Wishbone interconnect.
I checked most CSRBus masters and they are all generating proper `re` signal.

For etherbone I think it's accessing Bus with Wishbone, so it might be a Wishbone protocol violation at CSR Bridge or bone side.

m-byte:
In simulation, I can see that apparently the strobe and cyc signals are generated by the etherbone wb master. They can also be seen on `shared_cyc`/`_strobe` and `interface0_cyc`/`_stb`. However, while I do see `csr_interconnecr_dat_r` changing when `ack` is active on these buses, there's no change on `csr_interconnect_re`/`_we`.

m-byte:
I was able to narrow it down. When reading, sel is apparently set to 0 by the etherbone wb master.

m-byte:
Found it..  source.be isn't set in the `RECEIVE_READS` state of `LiteEthEtherboneReceiver`. Sorry @FlyGoat for mistakenly assuming your PR broke something, while it actually just showed that some older code was broken. I'm assuming it is something similar in the case of picorv32. I'll open a PR against liteeth to fix the etherbone master.

PR #2005, opened 2024-07-02
litex_json2dts_zephyr.py: Remove unnessesary configs

maass-hamburg:
Remove all configs, that are enabled by default
in zephyr based on the devicetree.

In zephyr drivers are automatically enabled based on status of device in the devicetree.


enjoy-digital:
Thanks @maass-hamburg.

PR #2012, opened 2024-07-10
soc/cores/video: Add additional color formats

machdyne:
This PR adds rgb332, mono8 and mono1 color formats.

This may be useful for improving video performance when more color depth isn't required.

Related to litex-hub/linux-on-litex-vexriscv#396


enjoy-digital:
Thanks @machdyne!

PR #2018, opened 2024-07-20
Add i2c master

AndrewD:
This PR imports the I2CMaster from misoc and adapts it to Litex.

It also includes the history of the various fixes that were required to make it actually work for READ: it appeared from our testing that while read was implemented it had never been used...

This core has limitations but it works and is far better than bitbang.

### DIFF
--- a/litex/soc/cores/i2c.py
+++ b/litex/soc/cores/i2c.py
@@ -1,0 +1,235 @@
+from migen import *
+from misoc.interconnect import wishbone
+
+
+__all__ = [
+    "I2CMaster",
+    "I2C_XFER_ADDR", "I2C_CONFIG_ADDR",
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @AndrewD, @rtucker85, this is merged. I'll do some test with it soon.

AndrewD:
@enjoy-digital I have a python app that can do an i2c scan over etherbone: maybe we can add this to litex in some form?

AndrewD:
@enjoy-digital thanks for merging so quickly 👍

enjoy-digital:
@AndrewD: The I2C scan script over Etherbone could be interesting yes. We could maybe integrate it to litex_cli, but you can just provide the script and I will eventually update/rework the integration while testing.

PR #2021, opened 2024-07-25
build/altera/common,platform: added ddrinput/ddrout primitives

trabucayre:
For *Agilex5* **ALTDDIO_OUT** and **ALTDDIO_IN** are not supported, and replaced by **tennm_ph2_ddio_out** and **tennm_ph2_ddio_in** (according to plaform designer generated code).

This PR is not fully tested due to errors at build time:
```
Error (23098): One or more blocks are configured incorrectly and will not have the desired functionality. --BCM instance name: hvio_0_1                                                                                                                                                                                        
Error (12274): A critical error occurred while the periphery placement was committed to the atom netlist. The atom netlist is now invalid and the Fitter must be restarted.                                                                                                                                                    
```

### DIFF
--- a/litex/build/altera/common.py
+++ b/litex/build/altera/common.py
@@ -99,7 +99,6 @@ class AlteraDDROutputImpl(Module):
             o_dataout  = o,
         )
 
-
 class AlteraDDROutput:
     @staticmethod
     def lower(dr):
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre, let's merge and eventually update when the build issue will be fixed (if related).

PR #2024, opened 2024-07-30
build/altera/common.py: implement SDRTristate for Agilex5 family

trabucayre:
`SDRTristate` is required to have SDCard in SDIO Mode working.


enjoy-digital:
Thanks @trabucayre.

PR #2028, opened 2024-08-01
soc: add add_spi_ram function

maass-hamburg:
After https://github.com/litex-hub/litespi/pull/70 is merged, the function to add a spi ram should be added to soc.py.

### DIFF
--- a/litex/build/openocd.py
+++ b/litex/build/openocd.py
@@ -27,12 +27,13 @@ class OpenOCD(GenericProgrammer):
         ])
         self.call(["openocd", "-f", config, "-c", script])
 
-    def flash(self, address, data, set_qe=False):
+    def flash(self, address, data, set_qe=False, init_commands=[]):
         config      = self.find_config()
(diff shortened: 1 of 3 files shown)


enjoy-digital:
Hi @maass-hamburg,

thanks a lot. Would you mind also sharing an example of SoC integration to ease our review?

maass-hamburg:
@enjoy-digital  adding the following to the init of the `class BaseSoC(SoCCore)` should be  enough. 
```
        self.add_spi_ram(name="psram", mode="4x", clk_freq=30e6, module=PSRAM(Codes.READ_1_4_4, program_cmd=Codes.PP_1_4_4))

        self.bus.add_region("main_ram", SoCRegion(
            origin = self.bus.regions["psram"].origin,
            size = self.bus.regions["psram"].size,
            linker = True
        ))
```

trabucayre:
Applied. Thanks @maass-hamburg 
It's a good starting point to integrate PSRAM with a more easy to use / intuitive way.

PR #2030, opened 2024-08-04
build/gowin/common.py: re-add tristate impl and SDRxxx for GW5A/Arora family (required for SDRAM use)

trabucayre:
This PR adds some GW5A primitives for SDRInput/SDROutput and SDRTristate.
Input/Output are simple `DFFSE` (no clocked in / out for arora) and may (or not) be inferred for sync blocks but without these SDRAM are not working.

Tested with @sipeed Tang Primer 25k and Tang Mega 138k pro


enjoy-digital:
Thanks @trabucayre!

PR #2035, opened 2024-08-11
Fix litex.build.gowin's __init__.py; closes #2034

### LINKED ISSUE #2034: `litex.build.gowin` is missing a proper `__init__.py`
The [`__init__.py` of the `litex.build.gowin` module](https://github.com/enjoy-digital/litex/blob/6309f30e0b3ff807249a6003730d3920bc88d0c0/litex/build/gowin/__init__.py) is miss-configured, as it contains no code at all. Looking at other modules [like the one for xilinx](https://github.com/enjoy-digital/litex/blob/6309f30e0b3ff807249a6003730d3920bc88d0c0/litex/build/xilinx/__init__.py) it can be seen that there's ment to be code in it.

The missing code leads to a problem when a programmer tries to use anything gowin related. I.e. following code will fail to be executed:
```py
from litex.build.gowin import GowinPlatform
```

While when using xilinx on the other hand it works perfectly fine:
```py
from litex.build.xilinx import XilinxPlatform
```

Mai-Lapyst:
This fixes the empty `__init__.py` in `litex.build.gowin`: https://github.com/enjoy-digital/litex/blob/6309f30e0b3ff807249a6003730d3920bc88d0c0/litex/build/gowin/__init__.py / https://github.com/enjoy-digital/litex/blob/master/litex/build/gowin/__init__.py

### DIFF
--- a/litex/build/gowin/__init__.py
+++ b/litex/build/gowin/__init__.py
@@ -1,0 +1,5 @@
+# Platforms.
+from litex.build.gowin.platform import GowinPlatform
+
+# Programmers.
+from litex.build.gowin.programmer import GowinProgrammer


trabucayre:
Applied. Thanks @Mai-Lapyst !

PR #2036, opened 2024-08-12
Adds apicula toolchain to gowin platform

Mai-Lapyst:
This PR:
- Adds GowinApiculaToolchain by using yosys + nextpnr
- Moves constraints generation from GowinToolchain.build_io_constraints to _build_cst to be reuseable
- Modifies NextPNRWrapper.get_call so command only contains flag for constraint fmt if it is non-empty. This is necessary as the `nextpnr-himbaechel` needs to be used, where the constaints file needs to be specified via `--vopt cst=file`, so the intended way of NextPNRWrapper for constraint-passing does not work anymore.

### DIFF
--- a/litex/build/gowin/apicula.py
+++ b/litex/build/gowin/apicula.py
@@ -1,0 +1,46 @@
+from litex.build.generic_platform import *
+from litex.build import tools
+from litex.build.gowin.gowin import _build_cst
+from litex.build.yosys_nextpnr_toolchain import YosysNextPNRToolchain
+
+class GowinApiculaToolchain(YosysNextPNRToolchain):
+    family     = "gowin"
+    synth_fmt  = "json"
+    pnr_fmt    = "report"
+    packer_cmd = "gowin_pack"
+
+    def __init__(self):
+        super().__init__()
+        self.additional_cst_commands = []
+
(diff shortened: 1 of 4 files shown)


litex/build/gowin/apicula.py
```diff
@@ -0,0 +1,46 @@
...
+    pnr_fmt    = "report"
+    packer_cmd = "gowin_pack"
+
+    def __init__(self):
+        super().__init__()
+        self.additional_cst_commands = []
+
+    def build_timing_constraints(self, vns):
+        # nextpnr-himbaechel for gowin currently doesnt supports timing constaints
+        pass
```
trabucayre, line 18: Not required since implemented by `GenericToolchain` module (and must return `("","")` when unused).

trabucayre:
LGTM.
Just a small remark: `build_timing_constraints` must be removed to let default behavior or a return must be added.
Thanks!

Mai-Lapyst:
> LGTM. Just a small remark: `build_timing_constraints` must be removed to let default behavior or a return must be added. Thanks!

Thanks for the response! Have removed the method you mentioned and also added the missing license header to apicula.py

trabucayre:
Applied.
Thanks @Mai-Lapyst

PR #2042, opened 2024-08-16
Gowin: add dual use gpio options

pepijndevos:
In the Gowin platforms these options are set to use dual-use pins.
In Apicula these are passed as command line arguments.

Counterpart to https://github.com/litex-hub/litex-boards/pull/603


trabucayre:
Applied. Thanks @pepijndevos !

PR #2043, opened 2024-08-19
fixes for apicula support

pepijndevos:
This maps Gowin IDE device families to Apicula database names, and updates the PLL generator to use the "revised PLL" primitive, which is basically the same as the PLL primitive but I guess they made some breaking changes to the implementation.

Stand by while I test more Gowin boards on Apicula in https://github.com/litex-hub/litex-boards/pull/603/files

### DIFF
--- a/litex/build/gowin/apicula.py
+++ b/litex/build/gowin/apicula.py
@@ -25,16 +25,27 @@ class GowinApiculaToolchain(YosysNextPNRToolchain):
         return (self._build_name + ".cst", "CST")
 
     def finalize(self):
+        devicename = self.platform.devicename
+        # Non-exhaustive list of family aliases that Gowin IDE supports but don't have a unique database
+        if devicename == "GW1NR-9C":
+            devicename = "GW1N-9C"
+        elif devicename == "GW1NR-9":
+            devicename = "GW1N-9"
+        elif devicename == "GW1NSR-4C" or devicename == "GW1NSR-4":
+            devicename = "GW1NS-4"
+        elif devicename == "GW1NR-4C" or devicename == "GW1NR-4":
+            devicename = "GW1N-4"
+
         pnr_opts = "--write {top}_routed.json --top {top} --device {device}" + \
             " --vopt family={devicename} --vopt cst={top}.cst"
         self._pnr_opts += pnr_opts.format(
             top        = self._build_name,
             device     = self.platform.device,
-            devicename = self.platform.devicename
+            devicename = devicename
         )
 
         self._packer_opts += "-d {devicename} -o {top}.fs {top}_routed.json".format(
-            devicename = self.platform.devicename,
+            devicename = devicename,
             top        = self._build_name
         )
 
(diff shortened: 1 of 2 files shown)


litex/build/gowin/apicula.py
```diff
@@ -25,16 +25,27 @@ def build_io_constraints(self):
...
+        # Non-exhaustive list of family aliases that Gowin IDE supports but don't have a unique database
+        if devicename == "GW1NR-9C":
+            devicename = "GW1N-9C"
+        elif devicename == "GW1NR-9":
+            devicename = "GW1N-9"
+        elif devicename == "GW1NSR-4C" or devicename == "GW1NSR-4":
+            devicename = "GW1NS-4"
+        elif devicename == "GW1NR-4C" or devicename == "GW1NR-4":
+            devicename = "GW1N-4"
+
```
trabucayre, line 38: why not using a dict as done for xilinx ?

trabucayre:
for `PLL`vs `rPLL` it make sense to also test with gowin's tools to verify for non regression.

litex/build/gowin/apicula.py
pepijndevos: Because there is a many to one relationship and I doubt a dict is faster for small numbers but also because I ported it from C. I don't really have strong feelings one way or the other.

pepijndevos:
It makes sense but I don't wanna 😭 

Our rPLL has exactly the same ports and parameters as the Gowin one, as documented in ther clock primitive docs.

The more important thing to verify is that there aren't any other parameter changes that need adjustment.

pepijndevos:
I did some digging and found an old datasheet with the original PLL documentation, which allowed me to diff the actual changes in rPLL:

```diff
--- pll.vhdl    2024-08-21 10:29:48.268678073 +0200
+++ rpll.vhdl   2024-08-21 10:21:43.595398786 +0200
@@ -1,5 +1,5 @@
 architecture Behavioral of Four_Bit_Adder is
-  component PLL
+  component rPLL
     generic
     (
       FCLKIN           : string  := "100.0";
@@ -27,15 +27,13 @@
     );
     port
     (
-      CLKIN      : in std_logic_vector;
+      CLKIN      : in std_logic;
       CLKFB      : in std_logic;
       IDSEL      : in std_logic_vector(5 downto 0);
       FBDSEL     : in std_logic_vector(5 downto 0);
       ODSEL      : in std_logic_vector(5 downto 0);
-      RESET      : in std_logic := '0';
-      RESET_P    : in std_logic := '0';
-      RESET_I    : in std_logic := '0';
-      RESET_S    : in std_logic := '0';
+      RESET      : in std_logic;
+      RESET_P    : in std_logic;
       PSDA, FDLY : in std_logic_vector(3 downto 0);
       DUTYDA     : in std_logic_vector(3 downto 0);
       LOCK       : out std_logic;
```

as you can see the main change is the removal of those two reset parameters, and CLKIN is no longer a vector(???), other than that they are the same primitive with the same parameters and ports.

pepijndevos:
That's all for currently working boards

trabucayre:
I will try to build/test bitstreams with gowin toolchain and your PLL's modification to check messages/behavior

pepijndevos:
Confirmed working with Gowin IDE

trabucayre:
Applied. Thanks @pepijndevos

PR #2050, opened 2024-08-27
efinix: Fix PLL with external clock input ifacewriter

Dolu1990:
Efinity tool was complaining of bad pin voltage for the PLL clock input.
The issue was only comming up when the pll input clock was using a different bank voltage than the default.

In practice this PR will add pin definition to the generated iface.py : 

```
# This line is still the same
design.gen_pll_ref_clock("pll0", pll_res="PLL_BL0", refclk_src="EXTERNAL", refclk_name="clk100", ext_refclk_no="1")

# add the following line : 
design.set_property("clk100","IO_STANDARD","3.3_V_LVCMOS")
```


trabucayre:
Applied. Thanks @Dolu1990 !

PR #2056, opened 2024-09-03
build/altera/common: added special AsyncResetSynchronizer based on altera_std_synchronizer_nocut

trabucayre:
For Agilex5: replace default/basic AsyncResetSynchronizer by a dedicated primitive.


enjoy-digital:
Thanks @trabucayre.

PR #2057, opened 2024-09-05
core/usb_ohci: fix SDRTristate clock

Dolu1990:
Fix USB OHCI SDRTristate using the wrong clock.

Wasn't always noticable, but when it does, this were going quite bad ^^


enjoy-digital:
Thanks @Dolu1990!

PR #2058, opened 2024-09-05
bios: add spiram

maass-hamburg:
add spiram in bios, so it can enable QPI for the spiram.

Its partially copied from `spiflash.c`, with things removed that are not needed.

For the freq_int it uses the memtest_data() function instead of the crc, as we can also write over mmap.


enjoy-digital:
Thanks @maass-hamburg!

PR #2059, opened 2024-09-05
soc/cores/vexiiriscv: update clocks + add video framebuffer support

Dolu1990:
The clock rework allows the CPU to work with a different clock than litex (higher fmax)
the video support add a memory coherent video output (will need to add more parametrization later, currently it is ycbcr 4:2:2 800x600@60hz


enjoy-digital:
Thanks @Dolu1990!

PR #2060, opened 2024-09-05
build/efinix: add a few IO primitives, IO constraints, sdc rework

Dolu1990:
add a few IO primitives, IO constraints, but mainly it rework how the SDC are handled.

The main issue with how the efinix flow was handled, is that the constraints for the syncronous IO were missing. This break stuff for high frequency design in large FPGA (Ti375)
The efinity tool generate a template outflow/xxx.pt.sdc. With this patch, this xxx.pt.sdc is combined with the one generated from litex.

!!! WARNING !!! 
There is one dirty thing is this PR.
This is the changes in efinix.py. when you setup a clock domain with a pll, this replace the clock domain "clk" with the new one created by the PLL.
This is done to ensure the constraints match with the efinix pll toplevel IO.

An alternative would be to provide a way to specify that a given litex clock domain should use the given name in the SDC file (instead of the clock domain "clk" name)
Let me know what you think about this, this can be reworked.

let's me know if the PR need to be splitted.

### DIFF
--- a/litex/build/efinix/common.py
+++ b/litex/build/efinix/common.py
@@ -139,24 +139,7 @@ class EfinixTristate(Module):
     def lower(dr):
         return EfinixTristateImpl(dr.platform, dr.target, dr.o, dr.oe, dr.i)
 
-# Efinix SDRTristate -------------------------------------------------------------------------------
 
-class EfinixSDRTristateImpl(Module):
-    def __init__(self, platform, io, o, oe, i, clk):
-        _o  = Signal()
-        _oe = Signal()
-        _i  = Signal()
-        self.specials += SDROutput(o, _o, clk)
-        self.specials += SDRInput(_i, i, clk)
-        self.submodules += InferedSDRIO(oe, _oe, clk)
-        tristate = Tristate(io, _o, _oe, _i)
-        tristate.platform = platform
-        self.specials += tristate
-
-class EfinixSDRTristate(Module):
-    @staticmethod
-    def lower(dr):
-        return EfinixSDRTristateImpl(dr.platform, dr.io, dr.o, dr.oe, dr.i, dr.clk)
 
 # Efinix DifferentialOutput ------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


litex/build/efinix/common.py
```diff
@@ -331,6 +431,7 @@ def lower(dr):
     Tristate               : EfinixTristate,
     DifferentialOutput     : EfinixDifferentialOutput,
     DifferentialInput      : EfinixDifferentialInput,
+    SDROutput              : EfinixSDROutput,
     SDRTristate            : EfinixSDRTristate,
     DDROutput              : EfinixDDROutput,
     DDRInput               : EfinixDDRInput,
```
maass-hamburg, line 437:
```suggestion
    DDRInput               : EfinixDDRInput,
    DDRTristate            : EfinixDDRTristate,
```

maass-hamburg (changes requested):

litex/build/efinix/common.py
Dolu1990: Ahhh right ^^
Thanks !

trabucayre:
PLL's modifications look good to be able to use a ClockSignal instead of a string but I see one issue:  When more than one module with a PLL are added to a gateware clk_name is not unique. Maybe it's required to append name with id (PLL id ?)

Dolu1990:
> Maybe it's required to append name with id (PLL id ?)

Either that, and this also mean that the clock name get some extra "noise", 
Either removing this whole trick around "cd.clk = clk_out" and instead reworking the emited SDC file by replacing all occurance of cd.clk names into its PLL input pin counter part. I think this would kinda be less hack. Wanna me to try ? Could probably be done here :  https://github.com/enjoy-digital/litex/pull/2060/files#diff-da6a1f772f6f2d1119b40ece412e6e70650c0bddfa9ba76ef8cd3d9887df77f5R327
It would just need a dictionary populated with all the replacement required. That dictionary could be populated in the create_clkout function.

trabucayre:
I already did some try with an id. The problem is more when the script is executed to avoid conflict between signals. I'm agree this adds a bit of noise. @enjoy-digital has also propose to adds primitive type+id when a clock is created to see if it's a PLL, or clkin,...

trabucayre:
Applied with few addtions. Thanks @Dolu1990 !

litex/build/efinix/common.py
```diff
@@ -261,13 +244,130 @@ class EfinixDifferentialInput:
...
 
+
+
+
+
+# Efinix DDRTristate ---------------------------------------------------------------------------------
+
+class EfinixDDRTristateImpl(Module):
+    def __init__(self, platform, io, o1, o2, oe1, oe2, i1, i2, clk):
+        assert oe1 == oe2
```
maass-hamburg, line 255: this assert doesn't work right if `oe1` and `oe2` are a `_Slice`, like in the litespi ddr phy

Dolu1990: If i remember well, one issue is that the efinix primitive only support 1 output enable signal (in ddr mode).
- Either when oe1 != oe2 it could fall back to DDR tristate emulation (et get very bad timings)
- Either it could be ignored and hope for the best
- Either it can keep the assert

Not sure what is the best.

Dolu1990: Another options would be to update litespi to avoid the usage of 2 oe signals

maass-hamburg: it was already with changes like: https://github.com/litex-hub/litespi/pull/60

The problem is this: `oe1 = dq_oe[i], oe2 = dq_oe[i],` with `dq_oe = Signal(len(pads.dq))`
where `oe1` and `oe2` are only a `_Slice``od a Signal and not a whole Signal. the error comes from here: https://github.com/m-labs/migen/blob/832a7240ba32af9cbd4fdd519ddcb4f912534726/migen/fhdl/structure.py#L41

maass-hamburg: this should fix it https://github.com/m-labs/migen/pull/292

PR #2065, opened 2024-09-11
bios: boot.c: fix warnings

maass-hamburg:
this fixes the warnings, when compiling.


enjoy-digital:
Thanks @maass-hamburg!

PR #2066, opened 2024-09-11
soc.py: add_ethernet: add mac addr constant

maass-hamburg:
add mac addr constant to add_ethernet,
so it matches the one from add_etherbone.


enjoy-digital:
Thanks @maass-hamburg!

PR #2068, opened 2024-09-12
build: io: make oe2 of DDRTristate optional

maass-hamburg:
make oe2 of DDRTristate optional.


maass-hamburg:
this https://github.com/m-labs/migen/pull/292 should be merged in migen before, so it wont raise a `TypeError`, when `oe1` and `oe2` are a `_Slice`

enjoy-digital:
Thanks @maass-hamburg for looking at this. We are currently not using upstream migen due to change to `pyproject.toml` for the installation so would also need to fix this if https://github.com/m-labs/migen/pull/292 is merged to use it.

maass-hamburg:
> Thanks @maass-hamburg for looking at this. We are currently not using upstream migen due to change to `pyproject.toml` for the installation so would also need to fix this if [m-labs/migen#292](https://github.com/m-labs/migen/pull/292) is merged to use it.

@enjoy-digital and what exactly is the problem with the `pyproject.toml` in migen? 
We need that fix also for https://github.com/enjoy-digital/litex/pull/2060 so we can use the DDRTristate when `oe1` and `oe2` are of the type `_Slice`. It being that, would be the most common  way to use it wenn it is use for something with multiplie lanes, like in LIteSPI.

maass-hamburg:
@enjoy-digital the part that needs https://github.com/m-labs/migen/pull/292 and https://github.com/enjoy-digital/litex/pull/2073 is now separated into https://github.com/enjoy-digital/litex/pull/2074

the difference is that: 46be90ea8177774f0664a614f134c3798e886d3f

maass-hamburg:
@enjoy-digital you might want to take another look. This now don't require the change in migen.

enjoy-digital:
Thanks @maass-hamburg!

PR #2069, opened 2024-09-12
sim: add HW_PREAMBLE_CRC for ethernet

### LINKED ISSUE #2067: Libliteeth is incorrectly configured in simulation
When running a simulation with default settings and "--with-ethernet" as the only flag, tftp boot doesn't work. For some reason, `CSR_ETHMAC_PREAMBLE_CRC_ADDR` is not set, even though it is necessary for proper operation.

m-byte:
This fixes the behavior of `ethernet_phy_model` `"sim"`. As the preamble is automatically attached by the tap, there is no need to add it from the BIOS. To let the BIOS know, `HW_PREAMBLE_CRC` needs to be set.

Fixes #2067


m-byte:
@enjoy-digital

enjoy-digital:
Thanks @m-byte!

PR #2070, opened 2024-09-12
BIOS: Add BOOTP support

m-byte:
Here are a few commits that enable BOOTP support in BIOS.

With this, it is possible to set IPs and filename from a server/computer. This simplifies usage of multiple devices in the same network.

### DIFF
--- a/litex/soc/software/bios/boot.c
+++ b/litex/soc/software/bios/boot.c
@@ -21,6 +21,8 @@
 #include "sfl.h"
 #include "boot.h"
 
+#include <bios/helpers.h>
+
 #include <libbase/uart.h>
 
 #include <libbase/console.h>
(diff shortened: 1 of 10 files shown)


m-byte:
@enjoy-digital

enjoy-digital:
Thanks @m-byte, this looks very interesting. I haven't yet closely looked at the code, but by default, do we still have the previous behavior with static IPs and no requirement for a BOOTP server on the other side? If not, I would prefer keeping it and enabling BOOTP through a build parameter.

m-byte:
@enjoy-digital it falls back to previous behavior after 3 failed attempts, with a relatively short time between those attempts (relative to the tftp timeout).

m-byte:
@enjoy-digital please let me know if the behavior described above would be fine for you, or whether this should still be behind a build parameter. If this were to be behind a build flag, it might make sense to still enable it by default if no IP is given by the user. I.e. to only disable it by default if the user sets a default IP and otherwise enable BOOTP, followed by assignment of the current fallback values.

PR #2073, opened 2024-09-16
litex_setup: use current version of migen

maass-hamburg:
use current version of migen. Use already install migen via pip (line 286), which supports `setup.py` and `pyproject.toml`, so it should make no difference, what migen is using.

Reverts: 6c30cb8695aa3f0760c627993031a1f4fa772416


enjoy-digital:
Thanks @maass-hamburg, this PR is now obsolete since we switched to current Migen with another PR.

PR #2075, opened 2024-09-18
build/efinix/common.py: ClkInput: added ClockSignal support

trabucayre:
`ClkInput` specials is currently only uses as PLL's clkin. The current approach is to uses a string for `create_clkin`. This way is valid because both `ClkInput`'s o name and PLL's `clkin` must match at *iface* level and nothing is known in a generated verilog file. But when this specials is used for a `ClockDomain` this way is no more valid and it's requires to connect `.o` to the `ClockSignal`.
The first commit address this limitation by allowing the use of a `ClockSignal`, converted to a signal when the class contructor is called. In this situation, the constructor adopt the same approach done for PLL's `create_clkout` method. string is kept as backward compatibility to avoid breaking all existing code.

Second commit address another issue: when a gpio out (ddio f.i.) is used through specials: `ClockSignal` is resolved and the Signal contains an `name_override` attribute. But when this primitive is instanciated directly by a module, cd remains a `ClockSignal` and conversion between signal and name must be done in ifacewriter script. Also the `ClockSignal` only provides `ClockDomain` name (not resolved) so a first search for the name is done, but if no results are found: the `_clk` suffix is added to the name.

### DIFF
--- a/litex/build/efinix/common.py
+++ b/litex/build/efinix/common.py
@@ -66,19 +66,34 @@ class EfinixAsyncResetSynchronizer:
 # Efinix Clk Input ---------------------------------------------------------------------------------
 
 class EfinixClkInputImpl(Module):
+    n = 0
     def __init__(self, platform, i, o):
-        o_clk  = platform.add_iface_io(o) # FIXME.
+        self.name = f"clk_input{self.n}"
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #2078, opened 2024-09-24
build: efinix: add function to add ip

maass-hamburg:
add function to add efinix IP.


enjoy-digital:
Thanks, this is merged.

PR #2086, opened 2024-09-30
build: io: don't use mutable object as default value

maass-hamburg:
don't use mutable object (here: ClockSignal()) as default value, because they will be the same object.
Leeds to problems, when for example two `SDRInput` are used in two different Modules and one of them is used with a `ClockDomainsRenamer()`, then both are changed.

more infos: https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil


enjoy-digital:
Thanks @maass-hamburg!

PR #2087, opened 2024-10-01
build: efinix: allow clk inverting and different in clk on reg tristates

maass-hamburg:
efinix:
- allow clk output inverting
- allow clk inverting on registered gpios

all:
- allow a different clock for input on SDR and DDR Tristates


maass-hamburg:
@enjoy-digital you might want to take a look?

maass-hamburg:
@enjoy-digital what do you think about this?

PR #2090, opened 2024-10-08
build: efinix: use ifacewriter to set bank voltage

maass-hamburg:
use efinix python api to set bank voltage,
instead of editing the peri.xml file.


trabucayre:
Applied. Thanks @maass-hamburg !

PR #2093, opened 2024-10-08
Fixes #2092: provides support for riscv gcc installation on Alpine Linux.

### LINKED ISSUE #2092: litex_setup.py does not install riscv gcc on Alpine Linux
Running the command:

`./litex_setup.py --gcc=riscv`

on Alpine Linux, it does not install **riscv** _gcc_.

It is useful to create Alpine-based Docker development containers.

mgaggero:
This PR adds support for installing RISC-V GCC in Alpine Linux environments, which is useful for creating Docker development containers based on Alpine images.

**WARNING**: To test the fix, you must run the command **litex_setup.py** with the _--dev_ option to prevent the script from auto-updating (the checksum is validated against the main branch of the original repository):

`./litex_setup.py --gcc=riscv --dev`

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -349,6 +349,9 @@ def riscv_gcc_install():
         # Arch.
         elif "arch" in os_release:
             os.system("pacman -S riscv64-linux-gnu-gcc")
+        # Alpine.
+        elif "alpine" in os_release:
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied. Thanks @mgaggero

PR #2095, opened 2024-10-09
soc/cores/cpu/zynqmp/core.py: added support for SGMII via PL with optional PTP support

trabucayre:
GEM controler may be used via PL (FPGA) and connected to RGMII, GMII or SGMII.
This PR adds required code + module needed for SGMII.


enjoy-digital:

PR #2099, opened 2024-10-18
vexiiriscv: add options and conditions

maass-hamburg:
- have opensbi things behind a condition

### DIFF
--- a/litex/soc/cores/cpu/vexiiriscv/core.py
+++ b/litex/soc/cores/cpu/vexiiriscv/core.py
@@ -57,6 +57,7 @@ class VexiiRiscv(CPU):
     with_rva         = False
     with_dma         = False
     with_axi3        = False
+    with_opensbi     = False
     jtag_tap         = False
     jtag_instruction = False
     with_cpu_clk     = False
@@ -163,6 +164,7 @@ class VexiiRiscv(CPU):
         VexiiRiscv.vexii_args += " --relaxed-branch"
 
         if args.cpu_variant in ["linux", "debian"]:
+            VexiiRiscv.with_opensbi = True
             VexiiRiscv.vexii_args += " --with-rva --with-supervisor"
             VexiiRiscv.vexii_args += " --fetch-l1-ways=4 --fetch-l1-mem-data-width-min=64"
             VexiiRiscv.vexii_args += " --lsu-l1-ways=4 --lsu-l1-mem-data-width-min=64"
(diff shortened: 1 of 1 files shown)


Dolu1990:
Hi,
For the with-rva side of things, because there is so many option possible to configure the CPU, the idea is more to pass them directly to VexiiRiscv SpinalHDL code itself through the --vexii-args="..." command line argument.
So : --vexii-args="--with-rva" for instance.

It is a bit hidden, but then, those --vexii-args parameters will also make SpinalHDL generate a .py which provide more information to litex. See pythondata-cpu-vexiiriscv/pythondata_cpu_vexiiriscv/verilog/xxxx.py. here is an example of this generated .py : 
```
VexiiRiscv.xlen = 32
VexiiRiscv.with_rvm = 1
VexiiRiscv.with_rva = 1
VexiiRiscv.with_rvf = 0
VexiiRiscv.with_rvd = 0
VexiiRiscv.with_rvc = 0
VexiiRiscv.with_lsu_software_prefetch = 1
VexiiRiscv.with_lsu_hardware_prefetch = "rpt"
VexiiRiscv.internal_bus_width = 128
``` 

Adding another variant for baremetal opensbi would be a better solution ?

maass-hamburg:
@Dolu1990 I understand the with-rva thing, even-though I thing that it would be preferable to have the option of the cups similar, so enabling it would be the same.

for the opensbi thing, we want to use this cpu with zephyr, so there is no need opensbi. Also wo don't like our csr positions be changed.

Dolu1990:
> I thing that it would be preferable to have the option of the cups similar, so enabling it would be the same.

Yes right. Overall i did it like this to reduce the codebase to its minimum and avoid redondancy. (Human stack overflow ^^)

> if VexiiRiscv.with_opensbi:

I'm totaly fine with your proposal. 
Note, any option which is added in the core.py need to be hashed through https://github.com/enjoy-digital/litex/pull/2099/files#diff-53633644687426d427b8a3569f41f700fbc8ba5f50a3493f56f35e0d47566a71R358 to ensure the netlist cache get an unique name. So with_opensbi would need to be hashed aswell.

I would say, just remove the RVA related stuff from the PR, add the with_opensbi into the hash and that is good :)

note that VexiiRiscv has no PMP support for now. I got some funding to implement, it is in my todo list, likely in the next few months it will come.

maass-hamburg:
> > I thing that it would be preferable to have the option of the cups similar, so enabling it would be the same.
> 
> Yes right. Overall i did it like this to reduce the codebase to its minimum and avoid redondancy. (Human stack overflow ^^)
> 
> > if VexiiRiscv.with_opensbi:
> 
> I'm totaly fine with your proposal. Note, any option which is added in the core.py need to be hashed through https://github.com/enjoy-digital/litex/pull/2099/files#diff-53633644687426d427b8a3569f41f700fbc8ba5f50a3493f56f35e0d47566a71R358 to ensure the netlist cache get an unique name. So with_opensbi would need to be hashed aswell.
> 
> I would say, just remove the RVA related stuff from the PR, add the with_opensbi into the hash and that is good :)
> 
> note that VexiiRiscv has no PMP support for now. I got some funding to implement, it is in my todo list, likely in the next few months it will come.

done

Dolu1990:
Thanks :D

PR #2100, opened 2024-10-23
efinix: gpio: use constant output option

maass-hamburg:
use constant output option, when the output is a `Constant`


maass-hamburg:
@enjoy-digital you might want to take a look. This improves open drain IOs, which are used for I2C. It also does some cleanup in the efinix ifacewriter

maass-hamburg:
@enjoy-digital what do you think?

enjoy-digital (approved):
Thanks @maass-hamburg and sorry for the delay, this looks good and that's  nice you see you managed to share common code for Input/Output.

PR #2101, opened 2024-10-24
bios: litespi: clear rx queue after write Beta

maass-hamburg:
clear rx queue at the end of spiflash_master_write().


enjoy-digital:
Thanks @maass-hamburg!

PR #2102, opened 2024-10-24
Vexiiriscv update

Dolu1990:
This PR bring improve the VexiiRiscv core with : 
- Better timings
- all RV32/64  x   RVF RVD legal combination supported
- VexiiRiscv SoC now support the ethernet rgmii with DMA accesses. 

The ethernet RGMII driver can be found here : 
- https://github.com/Dolu1990/litex-linux/tree/spinal-sgmac/drivers/net/ethernet/spinal
- Getting around 766 Mbits/s RX and 621 Mbits/s TX in linux with iperf3


enjoy-digital:
Nice, thanks @Dolu1990!

PR #2104, opened 2024-10-25
Fixes #2103: calculate  memory depth for WS2812

### LINKED ISSUE #2103: Unable to create WS2812 of nleds = 1
In led.py, `Memory` is used with depth = nleds.

https://github.com/enjoy-digital/litex/blob/c1225736a8c3dc3fb46dfafcaff763bd5fe9b0a2/litex/soc/cores/led.py#L135-L141

While migen's `Memory` type does not allow a depth of 1, which means if your are addressing a single word, no address signal is needed.

https://github.com/m-labs/migen/blob/c19ae9f8ae162ffe2d310a92bfce53ac2a821bc8/migen/fhdl/specials.py#L291-L298

See-also: https://github.com/m-labs/migen/pull/295

andelf:
This fixes #2103 .

When `Memory` depth is 1, a register should be used. In this fix, use memory of depth 2 to simplify this situation.

See-also: https://github.com/m-labs/migen/pull/295

### DIFF
--- a/litex/soc/cores/led.py
+++ b/litex/soc/cores/led.py
@@ -137,7 +137,8 @@ class WS2812(LiteXModule):
             self.bus  = bus = wishbone.Interface(data_width=32, address_width=32, addressing="word")
         else:
             # Memory.
-            mem = Memory(32, nleds, init=init)
+            mem_depth = max(nleds, 2)
+            mem = Memory(32, mem_depth, init=init)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @andelf, the simple solution is indeed perfectly fine for now to avoid different architectures for this specific case.

PR #2105, opened 2024-10-25
build: io: add multibit/bus variants of SDR and DDR

maass-hamburg:
add bus variants of SDR and DDR, which
support more than one pin.

### DIFF
--- a/litex/build/altera/common.py
+++ b/litex/build/altera/common.py
@@ -91,13 +91,14 @@ class AlteraDifferentialOutput:
 
 class AlteraDDROutputImpl(Module):
     def __init__(self, i1, i2, o, clk):
-        self.specials += Instance("ALTDDIO_OUT",
-            p_WIDTH    = 1,
-            i_outclock = clk,
-            i_datain_h = i1,
-            i_datain_l = i2,
-            o_dataout  = o,
-        )
+        for j in range(len(o)):
+            self.specials += Instance("ALTDDIO_OUT",
+                p_WIDTH    = 1,
+                i_outclock = clk,
+                i_datain_h = i1[j],
+                i_datain_l = i2[j],
+                o_dataout  = o[j],
+            )
 
 class AlteraDDROutput:
     @staticmethod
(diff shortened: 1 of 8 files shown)


enjoy-digital:
Thanks @maass-hamburg for the different PRs! Would you mind explaining the use-cases in the PR to be able to review the PR having the use-cases in mind?

maass-hamburg:
@enjoy-digital
for example in the sdr phy in litespi this: 
```
            for i in range(len(pads.dq)):
                self.specials += SDRTristate(
                    io = pads.dq[i],
                    o  = dq_o[i],
                    oe = dq_oe[i],
                    i  = dq_i[i],
                )
```

could be replaced with:

```
            self.specials += SDRTristateBus(
                    io = pads.dq,
                    o  = dq_o,
                    oe = dq_oe,
                    i  = dq_i,
                )
```

maass-hamburg:
@enjoy-digital IMO the prefered way to go instead of this would be to have the regular SDR/DDR variants support longer lengths than 1, just like the migen `Tristate`, but that would need changes to all vendor-specific implementations. But I also don't know, if that's the way litex should go.

enjoy-digital:
@maass-hamburg: That's indeed what I also tought when having a first look at the PR. If it's too much work to do it for all vendor initially, we could also just have assert in implementations to only build for the supported cases.

maass-hamburg:
@enjoy-digital I implemented this now also for all of the other vendors, but in a generic way, which should function, but might not be the most efficient way for that platform.

maass-hamburg:
@enjoy-digital rebased it, due to conflicts

maass-hamburg:
@enjoy-digital please take a look, would be nice to get this merged soon.

enjoy-digital:
@maass-hamburg: Thanks, I try to do this soon.

enjoy-digital:
Thanks @maass-hamburg, this looks fine. I think we should be able to simplify things and just use for the following constructs: `i[j]`  instead of `i[j] if len(i) > 1 else i` and same for all Input/Output/InOut?

enjoy-digital:
@maass-hamburg: Great, thanks. This looks fine. I'll just try to build some design with it and if OK, will merge.

enjoy-digital:
@maass-hamburg, this is merged. I had to do a few changes on top to avoid verilog slicing issues: https://github.com/enjoy-digital/litex/commit/0bfaf39d579f38bae37e0f31e18fe49a43a8b18a, https://github.com/enjoy-digital/litex/commit/e71e404ef45911d7d29b927bc372f115398e5156 and https://github.com/enjoy-digital/litex/commit/279b758584a64a3cdea7da5cfa543f1355292548.

PR #2115, opened 2024-11-04
Fix SOC region range check

### LINKED ISSUE #2114: SoC range check boundary issue
The check at https://github.com/enjoy-digital/litex/blob/0170462fe8a69a4d2de6f85bf8bf9e77e0771ceb/litex/soc/integration/soc.py#L321 does not properly consider the upper boundary of regions.

For example, a region starting at 0xFC00_0000 with a size of 0x0400_0000 will not fit in a container region starting at 0x8000_0000 with a size of 0x8000_0000, because both regions "end" at 0x1_0000_0000.

CKeilbar:
Closes #2114

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -318,7 +318,7 @@ class SoCBusHandler(LiteXModule):
         is_in = True
         if not (region.origin >= container.origin):
             is_in = False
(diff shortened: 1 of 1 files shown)


enjoy-digital (approved):
Thanks!

PR #2123, opened 2024-11-09
build/vhd2v_converter.py: allows users to pass a list of libraries files to compile before convert HDL.

trabucayre:
`VHD2VConverter` is used to convert one or more VHDL file to one verilog file. It is more or less similar to `Instance` with a dict of parameters, input, ouput,  inout signals and the list of files.
Based on **p_xx** a ghdl command is build and the result is one file verilog file without parameters, and with all signals with a fixed size based on generic.

In most of the case this behavior is fine, at least for generic IPs where no hardware primitives or vendor's blackbox are used.

When primitives are present, *ghdl* fails to convert sources because it don't know entity structures. To fix this use case vendor's libraries must be compiled as `.o` before conversion step.

This PR fix this issue by adding a list of libraries (the file path or a tuple(work, file path), and a first step before conversion consisting to call ghdl to produces associated `.o`.

### DIFF
--- a/litex/build/vhd2v_converter.py
+++ b/litex/build/vhd2v_converter.py
@@ -43,13 +43,16 @@ class VHD2VConverter(Module):
         force use of GHDL even if the platform supports VHDL
     _ghdl_opts: str
         options to pass to ghdl
+    _libraries: list of str or tuple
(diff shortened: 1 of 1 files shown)


enjoy-digital:

PR #2125, opened 2024-11-17
build/lattice/common.py: added Tristate support for ECP5 when build with diamond

trabucayre:
Unlike nextpnr/Trellis specials implementation, ECP5 has no dedicated implementation for Tristate Special.
This PR adds this special by using `BB` primitive.


enjoy-digital:

Issue #2126, opened 2024-11-20
No compatibility with Efinity 2024.2.294


m-byte:
There are multiple changes in 2024.2 that seem to break compatibility:
1. the `pt/db/device_*.xml` files are gone. This might be intentional, however, I just copied them over from the previous version for myself, which seems to help (will investigate further at some later point in time)
~~2. the vdb file is placed in the `outflow` directory, while litex expects it to be in `work_syn`. As a workaround, I added a check into my local codebase, whether the file exists in `work_syn` and use the `outflow` directory if it doesn't.~~

enjoy-digital:
Thanks @m-byte, we can probably close this since your PR has been merged.

m-byte:
@enjoy-digital not quite, as I only took care of the secind point above.

enjoy-digital:
@m-byte: OK thanks, that's re-opened then.

jwise:
Good idea on copying over the xml file from the previous version :) I looked briefly to see if the name had changed, and it appears that not just the name has changed but also the structural content.

jwise:
Thanks so much, @m-byte!

PR #2127, opened 2024-11-21
Fix vdb path for Efinity 2024.2

m-byte:
With Efinity 2024.2, the path of the vdb file has changed from `work_syn` to `outflow`. To fix this, first check if the file exists in `work_syn`. If it doesn't `outflow` is used.

(Fixes half of #2126)


trabucayre:
Applied. Thanks @m-byte !

PR #2128, opened 2024-11-26
Add efinix SEU interface

m-byte:
This enables instantiation of the Efinix single event upset interface.


enjoy-digital:
Thanks @m-byte!

PR #2129, opened 2024-11-26
liblitespi: add 4k erase function

m-byte:
Currently, only range erase is supported. In some cases, it might be necessary, to erase blocks with a finer granularity than 64k. With this function, erasing individual 4k blocks is possible.


enjoy-digital:
Thanks @m-byte!

PR #2130, opened 2024-11-26
cores/cpu/vexiiriscv: Add PMP support

Dolu1990:
The RISC-V PMP feature can now be enabled via --vexii-args="--pmp-size=8" for instance.

TOR support can be disabled via --pmp-tor-disable to save area / timings


enjoy-digital:
Thanks @Dolu1990!

PR #2133, opened 2024-12-03
Fixed for eliminating picolibc in package writing

kaolpr:
While trying to make a reproducible LiteX build using Nix I found that `libc` `Makefile` copies CPU family files to `$PICOLIBC_DIRECTORY/newlib/libc/machine/$(CPUFAMILY)/`. This directory is a package directory and should not be altered during build (what was the cause of build failure due to insufficient permissions in my case).

This PR suggests a walk around that copies all contents of `$PICOLIBC_DIRECTORY` to `picolibc_src` in current build directory and copies `libc` CPU family files there.

### DIFF
--- a/litex/soc/software/libc/Makefile
+++ b/litex/soc/software/libc/Makefile
@@ -39,11 +39,13 @@ cross.txt:
 	@echo "$$CROSSFILE" > $@
 
 __libc.a: cross.txt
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @kaolpr, sorry for the delay. This is merged.

PR #2139, opened 2024-12-13
cores/clock/intel: add reset to Intel PLLs

piotro888:
`self.reset` is not connected in `IntelClocking` PLLs.
I think it was simply missed, or is there some reason for it? 

It makes declared reset input in some current Intel board definitions not effective (like [here](https://github.com/litex-hub/litex-boards/blob/a3b38163e82c14348fdfbd2adbc0a26b96c70952/litex_boards/targets/sitlinv_a_e115fb.py#L36)) and `soc_rst` that works via CRG ignored.

### DIFF
--- a/litex/soc/cores/clock/intel_common.py
+++ b/litex/soc/cores/clock/intel_common.py
@@ -105,8 +105,15 @@ class IntelClocking(LiteXModule):
             return best_config
         raise ValueError("No PLL config found")
 
+    def add_reset_delay(self, cycles):
+        for _ in range(cycles):
+            reset = Signal()
+            self.specials += Instance("DFFE", i_clk=self.clkin, i_d=self.reset, o_q=reset, i_ena=1, i_clrn=1, i_prn=1)
+            self.reset = reset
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @piotro888, have you tested reboot command in the BIOS with this PR? If so and working, this could indeed be merged.

piotro888:
BIOS reset hangs up the SoC. I only tested external reset before, thanks!
I will check if workarounds from Xilinx clocking to delay the reset or anding `locked` with `~reset` would solve this issue

enjoy-digital:
@piotro888: Thanks for the feedback, that's possible the reset was not present for this reason, but spending enough time lookng at it should allow finding a proper solution.

piotro888:
`add_reset_delay` adapted from Xilinx clocking fixed the issue

enjoy-digital:
Great, thanks a lot @piotro888! Let's merge then.

piotro888:
Thanks!

PR #2140, opened 2024-12-13
soc/ethernet: enable full_memory_we by default for Quartus toolchain

piotro888:
There is an reported issue with Quartus toolchain, that fails to infer memories with multiple write enable signals.
It is a problem for liteeth module with large buffers (https://github.com/enjoy-digital/liteeth/issues/63) which was fixed and later changed  to configurable `full_memory_we`  option in https://github.com/enjoy-digital/liteeth/pull/72, because of regression on other FPGAs.

However, this option/problem is not documented and not used in existing codebase, and leads to unexpected, difficult to trace problems with high device utilization.
Because of that, I think it would be a good idea to enable this option by default if Quartus toolchain is detected.

The second place where `full_memory_we` option is defined is L2 Cache, where it is enabled by default for all targets.

### DIFF
--- a/litex/build/altera/quartus.py
+++ b/litex/build/altera/quartus.py
@@ -14,6 +14,7 @@ import math
 from shutil import which
 
 from migen.fhdl.structure import _Fragment
+from migen.fhdl.simplify import FullMemoryWE
 
 from litex.build.generic_platform import Pins, IOStandard, Misc
 from litex.build.generic_toolchain import GenericToolchain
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @piotro888 and sorry for the delay, otherwise maybe we could just use the same approach than on Efinix design and always transform the memories: https://github.com/enjoy-digital/litex/blob/master/litex/build/efinix/efinity.py#L73-L74 since it always seems to be requred on Altera FPGA?

piotro888:
Yes, that makes more sense. Updated and tested. Thanks!

enjoy-digital:
Great, thanks. This is merged.

PR #2143, opened 2024-12-17
soc/integration/soc: Fix CSRBridge Bus Width conversion

FlyGoat:
Wishbone2CSR and AXILite2CSR bridges are incapable for performing bus width conversion, which means it's Bus slave port must have same width as CSRs.

Use CSR width to create slave bus to allow width adaptar to be inserted by add_slave. Also add relevant assertion.

This fixes `--bus-data-width 64` build.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1138,13 +1138,14 @@ class SoC(LiteXModule, SoCCoreCompat):
         }[self.bus.standard]
         csr_bridge_name = f"{name}_bridge"
         self.check_if_exists(csr_bridge_name)
+        data_width = self.csr.data_width
         csr_bridge = csr_bridge_cls(
             bus_bridge_cls(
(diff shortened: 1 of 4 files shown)


FlyGoat:
Maybe `--bus-data-width` should be covered by CI? I'm not sure if it's worthy :-)

enjoy-digital:
Thanks @FlyGoat! Testing this in CI could indeed be interesting and should not be too complicated with the current CI in place, we could add some` --bus-standard` / `--bus-data-width` tests.

I just need to look at the CI failing test (related to the new assert) and we'll be able to merge.

FlyGoat:
@enjoy-digital Sorry for the oversight, CI is now fixed. It was due to difference on default bus parameters. I'm working on a soc_bus test case, will PR when it's ready.

Thanks

enjoy-digital:

PR #2117, opened 2024-11-06
Use MMCME4_ADV in USPMMCM to enable finer-grained clock output ctrl

long-pham:
Split PR #2111 into 3 parts as requested in[ #2111#issuecomment-2454519454](https://github.com/enjoy-digital/litex/pull/2111#issuecomment-2454519454)

Part 1/3: Update USPMMCM to use MMCME4_ADV for finer clock output control
- Enables 0.125 step size for clock division/multiplication
- Main change: replace MMCME2_ADV with MMCME4_ADV

PR #2148, opened 2024-12-21
backwards compatibility with python3.7

dayjaby:
https://github.com/enjoy-digital/litex/pull/2117 broke my linux-on-litex-vexriscv build as the f4pga conda environment still uses python3.7 and the new (python3.8) [assignment expression](https://docs.python.org/3/whatsnew/3.8.html#assignment-expressions)  got used in `if specific_div_range := getattr(self, f"clkout{n}_divide_range", None):`.

For the sake of backwards compatibility, and not unnecessarily breaking the build for other people, this PR gets rid of the assignment expression.

Edit: Same issue with building versa_ecp5 in linux-on-litex-vexriscv as the nextpnr-ecp5 install (`conda install litex-hub::yosys litex-hub::nextpnr-ecp5`) requires python>=3.7,<3.8 but some script is using the python3.9's str removesuffix method.

### DIFF
--- a/litex/soc/cores/clock/xilinx_usp.py
+++ b/litex/soc/cores/clock/xilinx_usp.py
@@ -143,7 +143,8 @@ class USPMMCM(XilinxClocking):
 
                     dividers = list(clkdiv_range(*self.clkout_divide_range))
                     # Add specific range dividers if they exist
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @dayjaby, this is merged.

PR #2153, opened 2024-12-30
[enh] Added differential input for ICE40

chmousset:
Everything is in the title.
Validated on ice40up5k pin 32


enjoy-digital:
Thanks @chmousset!

PR #2158, opened 2025-01-13
soc/interconnect/wishbone: Add Wishbone CDC

david-sawatzke:
Adds a wishbone CDC implementation. Inspired by BusSynchronizer, uses the PulseSynchronizer to begin/end transaction and ensure all bus bits are stable. 

Most of it is handling the WB interface definitions to generate interfaces and signals for both directions.

### DIFF
--- a/litex/soc/interconnect/wishbone.py
+++ b/litex/soc/interconnect/wishbone.py
@@ -14,6 +14,7 @@ from math import log2
 from migen import *
 from migen.genlib import roundrobin
 from migen.genlib.record import *
+from migen.genlib.cdc import PulseSynchronizer, MultiReg
 
 from litex.gen import *
 from litex.gen.genlib.misc import split, displacer, chooser, WaitTimer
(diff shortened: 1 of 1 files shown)


david-sawatzke:
Taking a quick look:
- MultiReg with manual synchronization with PulseSynchronizer vs. ClockDomainCrossing: The former is more complex to implement but might take less resources, as the stream AsyncFIFO is at least 4 large
- My implementation extracts the signals from the Interface which is less readable, but matches all variations of the Interface.
- I *think* you're incorrectly handling `err`. According to Rule 3.45 either `err` or `ack` can be asserted (to end the cycle), but not both.
- An additional timeout on the access FSM, which I don't think is required
- A timeout happening on the master side is not propagated to the other side by me and can result in weird accesses. It's also not properly handled by your implementation (might generate errant ACKs on the master side), but much less likely due to the short slave timeout.

enjoy-digital:
Thanks again for the PR and the detailed comments.

I reviewed this against the current LiteX CDC implementation. We now have Wishbone CDC support in LiteX, integrated with `SoCBusHandler.add_clock_domain_crossing`, and using the same `stream.ClockDomainCrossing` style as AXI/AXI-Lite. This is probably a bit larger than a PulseSynchronizer/MultiReg implementation, but it keeps the code simpler, consistent across bus types, and easier to maintain.

 One point from your review was valid and has now been improved on master with enjoy-digital/litex@08620b572: if a Wishbone master drops `cyc/stb` before the CDC response comes back, the CDC now drains the late response without pulsing `ack/err` back into an inactive master cycle. I also made the `cti`/`bte` CDC payload widths derive from the actual interface signals.

With this, I think the useful parts of this PR are now covered in the current implementation, so we can close it as superseded.

PR #2159, opened 2025-01-13
cores/cpu/openc906: Align with pythondata RTL

FlyGoat:
Use RTL from pythondata-openc906, remove unused file lists, remove outdated bus conversion logic, remove "riscv-dm" and "ethmac" from mem_map as they can be allocated automatically, use DTM to implement add_jtag.

Get the core working.


enjoy-digital:

PR #2160, opened 2025-01-14
cores/cpu/ibex: Align with latest RTL

FlyGoat:
Fix up file path and workaround verilator limitation on parameter type.

Also fix crt0 and system.h to reflect RTL changes.


enjoy-digital:
Thanks a lot @FlyGoat!

PR #2161, opened 2025-01-14
gen: fhdl: expression: resolve slice completly

maass-hamburg:
resolve slice completly, to reduce complexity
in the verilog files.


enjoy-digital:
Nice! Thanks @maass-hamburg, this makes sense and indeed reduce complexity.

maass-hamburg:
@enjoy-digital lowering slices is still needed, if it is used together with Cat(). I moved the logic of this inside the lowerer: https://github.com/m-labs/migen/pull/301

PR #2163, opened 2025-01-16
soc.py: ethernet: use phy_cd name from phy

maass-hamburg:
use phy_cd name directly from phy.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -1838,7 +1838,7 @@ class LiteXSoC(SoC):
             )
 
     # Add Ethernet ---------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @maass-hamburg, would you mind describing the context/use-case to think about it and review it with this in mind?

maass-hamburg:
It uses the same logic, as the part a little bit under it at ``with_timing_constraints``, to get the phy names. This can be very useful, when you want to use than one phy (multiple instances), specificly when each phy has its own ClockDomains, like here: https://github.com/enjoy-digital/liteeth/pull/181. There I numbered the ClockDomain Names, so they would not replace each other.

PR #2166, opened 2025-01-21
soc: interconnect: add logic to set a offset to bus slaves

maass-hamburg:
add logic to set a offset to bus slaves.

### DIFF
--- a/litex/soc/integration/soc.py
+++ b/litex/soc/integration/soc.py
@@ -2125,8 +2125,7 @@ class LiteXSoC(SoC):
         spiflash_core = LiteSPI(spiflash_phy, mmap_endianness=self.cpu.endianness, **kwargs)
         self.add_module(name=f"{name}_core", module=spiflash_core)
         spiflash_region = SoCRegion(origin=self.mem_map.get(name, None), size=module.total_size)
-        self.bus.add_slave(name=name, slave=spiflash_core.bus, region=spiflash_region)
-        self.comb += spiflash_core.mmap.offset.eq(self.bus.regions.get(name, None).origin)
+        self.bus.add_slave(name=name, slave=spiflash_core.bus, region=spiflash_region, offset_base=True)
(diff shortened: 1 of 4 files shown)


maass-hamburg:
@enjoy-digital please take a look. It might be very useful for other bus slaves, that also expect a start at ``0``

enjoy-digital:
Thanks @maass-hamburg, would you mind describing the context/use-case to think about it and review it with this in mind?

maass-hamburg:
@enjoy-digital Most modules with a axi or wishbone bus don't have a offset logic(they don't know the offset in the ``SoCRegion``), which means that they will send the axi or whishbone address directly to the device. If you want to access the address ``0x0`` an a mmaped flash(the first byte on that device), and the SoCRegion has the origin of that device at ``0x400``, you would start a read via the axi or wishbone bus at ```0x4000``, but that would lead to you reading at 0x4000 on that device, but you wanted 0x0. that why we have to substract the origin/start from the address. 

The reason, why if wasn't noticed before, is that
- a lot of devices have a rollover ( you want 0x405, but the device is only 0x100 big, it will return the value at 0x5)
- some modules ignore the last bits, because they will only use the first 16 bit, so if our offset is 0x20000, than the 0x2 is ignored.

enjoy-digital:
Thanks @maass-hamburg, this is merged. I'll just change the parameter name.

enjoy-digital:
`offset_base` has been renamed to `strip_origin` with https://github.com/enjoy-digital/litex/commit/0f620d01e2f680350a2c9dfcc4274344534fbcf4. Hoping this is good for you @maass-hamburg.

enjoy-digital:
Also fixed an issue with https://github.com/enjoy-digital/litex/commit/86b052e41fa894c323ed8177b469fdb6846032c4 since region can be None when automatically allocated by LiteX.

PR #2167, opened 2025-01-22
build: common: make sure Signal lenght is right

maass-hamburg:
make sure the the lenght of the o2 Signal is
the same lenght of o1, when a DDRImput implementation is used
for a SDRInput. This is needed for multibit io.

This was forgotten in https://github.com/enjoy-digital/litex/pull/2105


enjoy-digital:
Thanks @maass-hamburg!

PR #2168, opened 2025-01-23
build: common: TristateImpl: support longer oe

maass-hamburg:
support setting a different oe for every line in
the Tristates.

this is also needed for the multibit io, where oe has the same lenght
as the io., because some platforms use the generic implementation
of the SDR/DDRTristate, that uses a Tristate internally.


enjoy-digital:

Issue #2169, opened 2025-01-26
NaxRISCV Bare Metal Demo not working


BMorgan1296:
Hi,

I am experimenting with different cores and found that the bare metal demo does not work for NaxRISCV under simulation or FPGA.

I am able to boot, but only get the "Liftoff!" printout and nothing else once it jumps into the demo.bin. This is due to the processor stalling on the interrupt handling routines specified in demo's main.c:

```C
int main(void)
{
// #ifdef CONFIG_CPU_HAS_INTERRUPT      //       <-- If I comment these out I get the donut :)
//	 irq_setmask(0);
//	 irq_setie(1);
// #endif
	uart_init();

	donut_cmd();

	help();
	prompt();

	while(1) {
		console_service();
	}

	return 0;
}

```

Having spent a bit of time figuring out why, this seems related to #1754, #1584, #2023, where the processor did not have its interrupt handling defined.

I checked `/soc/cores/cpu/naxriscv/irq.h` and the interrupt handling is all there, which I assume is used to tell the core when I type into the terminal and press enter etc. 

I have found that the set global interrupts function, `irq_setie(1);` is causing the CPU to halt execution when it tries to enable the interrupst with the following: `asm volatile ("csrrs x0, " "mstatus" ", %0" :: "i"(CSR_MSTATUS_MIE));`

Which translates to `asm volatile ("csrrs x0, " "mstatus" ", %0" :: "i"(0x8));`

I have tested and setting this `mstatus` register to `0x0` doesn't halt the processor, as based off the instruction set, a value of 0x0 will not write to the register.

Cheers.

BMorgan1296:
Ok I have figured it out.

The NaxRISCV is booting into user mode, and for comparison, the VexRISCV is in machine mode.

Mstatus for the NaxRISCV is 0x2000, i.e. MPP field which is bits 11,12 = 0x00 == User mode.

Mstatus for the NaxRISCV is 0x1800, i.e. MPP field which is bits 11,12 = 0x11 == Machine mode.

Just figuring out how to change this now :)

Dolu1990:
Hi ^^

> The NaxRISCV is booting into user mode

NaxRiscv boot in machine mode.
The mstatus.mpp field doesn't indicate the current privilege mode, the the privilege mode which will be taken on the next mret instruction.

If do baremetal stuff only, you never need to change the mpp value, because once the CPU enter an interrupt, the MPP field is changed to the privilege the CPU was in (machine).

[...]

PR #2170, opened 2025-01-29
Fix NaxRISCV Bare Metal Demo and small fix for demo.py make routine

BMorgan1296:
Two small contributions:
* Added two small fixes to the NaxRISCV bare metal demo (which was not working due to lack of interrupts in #2169) 
* A small QoL fix in the demo.py routine which needed to be cleaned before running `make`.


BMorgan1296:
#2169 was resolved independently (fresh installs work!) so can close this.

PR #2171, opened 2025-02-02
soc/cores/clock/gowin_gw5a.py: GW5AT uses PLLA

trabucayre:
As for *GW5A-*, *GW5AT* uses PLLA. Without this fix synthesis fails because no PLL is found.


enjoy-digital:
Thanks @trabucayre.

PR #2181, opened 2025-02-14
build/efinix/platform.py: get_pins_name: avoids to create a new signal with a name already existing

trabucayre:
As mentioned in  #2180 the uses of ifacewriter and out-of-verilog primitives has for effect to creates new signals.
#2180 has fixed issues with 1 bit signals but when a multi bit signal is used directly a new signal with the same name is created.

This PR adds code to appends new name with a extension to avoid conflicts.


enjoy-digital:

PR #2183, opened 2025-02-16
Added fujprog in Lattice progammer for ULX3S

tarik-hamedovic:
While trying to program the ULX3S 85F variant board I got an error when trying to use `ujprog`.

I came across[ this issue](https://github.com/emard/ulx3s-bin/issues/7) where it says that using `ujprog` is outdated and that `fujprog` should be used instead. So I added the fujprog command in the lattice `programmer.py` file. 
Now the ULX3S board programs nicely with the `--load` command.


enjoy-digital:
Thanks @tarik-hamedovic, this is merged.

PR #2185, opened 2025-02-16
Optimize memspeed test for cores with slow modulo.

cr1901:
The random memspeed test uses modulo to bounds-check reads. The modulo operator is [rather heavy](https://mastodon.social/@cr1901/114015524022352658) on cores without division, and the time spent calculating it can dwarf the amount of time actually reading from memory.

By default, the memtests seem to use power-of-two sizes. If BIOS LTO is enabled (which is _not_ default), then the compiler should optimize modulo to an AND mask. However, since memtest is by default part of a static lib without LTO, the compiler can't do this optimization. This patch adds a dynamic check for power-of-two size and manually does the optimization if it applies.

After this patch, I've seen LiteX go from reporting a 4KiB/s random access read on QSPI flash to ~150KiB/s!

### DIFF
--- a/litex/soc/software/libbase/memtest.c
+++ b/litex/soc/software/libbase/memtest.c
@@ -318,11 +318,24 @@ void memspeed(unsigned int *addr, unsigned long size, bool read_only, bool rando
 	start = timer0_value_read();
 
 	int num = size/sz;
+	int ones_cnt = 0;
+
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @cr1901! Looks interesting and will indeed report more realistic speeds for these configurations.

PR #2186, opened 2025-02-17
Use Python's raw string in regex pattern.

higuoxing:
Otherwise, there will warnings like:

```
./litex/soc/cores/cpu/cv32e40p/core.py:68: SyntaxWarning: invalid escape sequence '\$'
  res = re.search('\$\{DESIGN_RTL_DIR\}/(.+)', l)
```


enjoy-digital:
Thanks @higuoxing.

Issue #2178, opened 2025-02-07
Converting Wishbone to riscv-steel custom bus


BMorgan1296:
Hi,

I am trying to incorporate the riscv-steel (https://github.com/riscv-steel/riscv-steel) processor into litex.
I have been able to get this working using a custom bus to Wishbone converter in simulation.

However, when I try to implement on FPGA, the processor does not boot properly. 
I have tried several various debugging methods for the past week to no avail.
I have attempted to use LiteScope to read in the values of the bus at boot, however as this is locked to the "sys" clock domain, I am unable to read these values at boot, as the scope it self also resets... I can read them in the middle of broken execution however I am finding it difficult to debug this.

Unfortunately, the riscv-steel repo does not document how this bus interface is truly meant to work. I have attempted to use the OBI2Wishbone used in the CV32E40P as a guide. Does anyone have any tips as to what I can do to debug this?

My implementation of the converter is below. 

```Python
steel_layout = [
    ("rw_address", 32),
    ("read_data", 32),
    ("read_request", 1),
    ("read_response", 1),
    ("write_data", 32),
    ("write_strobe", 4),
    ("write_request", 1),
    ("write_response", 1),
]

class SteelCore2Wishbone(LiteXModule):
    def __init__(self, steel_idbus, idbus):
        self.submodules.fsm = fsm = FSM(reset_state="IDLE")

        fsm.act("IDLE",
            If(steel_idbus.read_request,
                idbus.adr.eq(steel_idbus.rw_address),
                idbus.stb.eq(1),
                idbus.cyc.eq(1),
                idbus.we.eq(0),
                NextState("READ")
            ).Elif(steel_idbus.write_request,
                idbus.adr.eq(steel_idbus.rw_address),
                idbus.dat_w.eq(steel_idbus.write_data),
                idbus.stb.eq(1),
                idbus.cyc.eq(1),
                idbus.sel.eq(steel_idbus.write_strobe),
                idbus.we.eq(1),
                NextState("WRITE")
            ),
        )

        fsm.act("READ",
            If(idbus.ack,
                steel_idbus.read_data.eq(idbus.dat_r),
                steel_idbus.read_response.eq(1),
                NextState("IDLE")
            )
        )

        fsm.act("WRITE",
            If(idbus.ack,
                steel_idbus.write_response.eq(1),
                NextState("IDLE")
            )
        )
```

[...]

PR #2190, opened 2025-02-20
Added preliminary support for riscv-steel processor. 

BMorgan1296:
RTL held in `riscvsteel/standard/` subdirectory for now.

A few things to note:

I fixed (from issue #2178) the steel <-> wishbone converter, and this design mostly works on FPGA.

However, memory errors occur on the Digilent Nexys Video, yet everything works fine on Arty A7. I also have a KCU105 to test with but haven't yet.

I copy pasted the bootloader code from the VexRISCV and it seemed to work fine, I think it is all alright but this is my first rodeo with adding a CPU so might have missed something there.

`litex_sim --cpu-type=riscvsteel` should work fine.

### DIFF
--- a/litex/soc/cores/cpu/riscvsteel/__init__.py
+++ b/litex/soc/cores/cpu/riscvsteel/__init__.py
@@ -1,0 +1,1 @@
+from litex.soc.cores.cpu.riscvsteel.core import riscvsteel
--- a/litex/soc/cores/cpu/riscvsteel/boot-helper.S
+++ b/litex/soc/cores/cpu/riscvsteel/boot-helper.S
@@ -1,0 +1,4 @@
+.section    .text, "ax", @progbits
+.global     boot_helper
+boot_helper:
+	jr x13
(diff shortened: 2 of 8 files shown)


BMorgan1296:
Ok so I tried implementation again on the Arty A7 FPGA and it is also having memory errors....

```
Memtest at 0x40000000 (2.0MiB)...
  Write: 0x40000000-0x40200000 2.0MiB
   Read: 0x40000000-0x40200000 2.0MiB
  bus errors:  0/256
  addr errors: 0/8192
  data errors: 17/524288
Memtest KO
Memory initialization failed
```

Same thing on the Digilent Nexy Video. So something is up here.

BMorgan1296:
Hi,

Any updates on this? I would really like to test this on an FPGA but am unsure how to debug this specifically. Any hints as to what I can try?

BMorgan1296:
I have found that the SDRAM is causing the issue.
Using `--integrated-main-ram-size=0x20000` helps to lessen the number of memory errors.
On the DRAM side, I attempted changing the timing ratio but this did not seem to help.

I have also adjusted the sys_clk but it does not seem to have much of an effect on memory issues.

The next issue I am facing is the UART upload, which is very hit-and-miss. I get this error a fair amount and have to retry until it works, does not matter if I use `--safe`:

`[LITEX-TERM] Got unexpected response from device 'b'E''`

PR #2191, opened 2025-02-20
Fix Efinity 2024.2 missing device_*.xml files

m-byte:
Since Efinity 2024.2.294, the `/pt/db/device_*.xml` files are no longer shipped with the IDE. While it was possible to copy the files over from an older version, this likely will limit support for newer FPGAs.

With this change, those files are not required anymore.

This fixes the other (remaining) half of #2126


enjoy-digital:
Thanks a lot @m-byte!

PR #2193, opened 2025-03-03
build: efinix: programmer: add flash function

maass-hamburg:
add flash function for the efinix
programmer via jtag_bridge_new.


enjoy-digital:
Thanks @maass-hamburg!

PR #2199, opened 2025-03-05
Add a data_width parameter to the add_axi_hp_slave function, allowing…

itisme:
Since the Zynq 7000 AXI HP data width supports both 64-bit and 32-bit, using a 64-bit data width with a 32-bit CPU will cause errors. Therefore, a data_width parameter is added to the add_axi_hp_slave function to address this issue.


trabucayre:
Applied after:
- removing second commit related to `litex_setup.py`
- added `PCW_S_AXI_HP{n}_DATA_WIDTH` configuration

Thanks @itisme

PR #2204, opened 2025-03-11
update neorv32

pepijndevos:
WIP counterpart to https://github.com/stnolting/neorv32/pull/1204 that updates neorv32 and adds the ability to use multiple cores

Currently a bit WIP and broken.

I was able to run multicore on 1.9.7 but after the update the simulation hangs.

I've also temporarily update the git url.

And I haven't hooked up the debug jtag correctly.

### DIFF
--- a/litex/soc/cores/cpu/neorv32/boot-helper.S
+++ b/litex/soc/cores/cpu/neorv32/boot-helper.S
@@ -1,4 +1,17 @@
 	.section .text, "ax", @progbits
 	.global boot_helper
+.global smp_lottery_target
+.global smp_lottery_lock
+.global smp_lottery_args
+
 boot_helper:
+	sw x10, smp_lottery_args  , x14
+	sw x11, smp_lottery_args+4, x14
+	sw x12, smp_lottery_args+8, x14
+	sw x13, smp_lottery_target, x14
+	fence w, w
+	fence.i
+	li x15, 1
+	sw x15, smp_lottery_lock, x14
 	jr x13
+
--- a/litex/soc/cores/cpu/neorv32/core.py
+++ b/litex/soc/cores/cpu/neorv32/core.py
@@ -100,6 +100,7 @@ class NEORV32(CPU):
         # # #
 
         # CPU Instance.
+        self.cores = []
         for i in range(self.cpu_count):
             cpu_params = dict(
                 # Clk/Rst.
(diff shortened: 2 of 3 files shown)


pepijndevos:
One thing I don't particularly like is that this seems to synthesize the whole cpu to verilog for every core, and I can't figure out how to do parameterized VHDL modules otherwise. Pretty much only to pass the HART_ID.

pepijndevos:
So @stnolting mentioned the IO base address changed since then, but changing that I did not get it to boot yet. I don't have a very good way to see what the CPU is doing though. Like, how do I see what instructions it's trying to execute from where...

https://github.com/stnolting/neorv32/pull/1204#issuecomment-2716606976

pepijndevos:
I tried to bisect what broke it:
```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
56e39b5a099e1e729f86f124662e39c0d53b40da
7c6abcdf1384e39bbe98f3237728400865faf06d
2835087cedb975f23e660cc512df24f340426b5f
c8ae94c2878fa3ae6031af0550d530ec7ade7652
70f0c4bccb7144b26a8fab96b32311647703d295
We cannot bisect more!
```

All of those commits fail to compile with `neorv32_top.vhd:772:9:error: no declaration for "imem_as_irom"`

None of those are related to the mentioned IO base address change, but rather changing the boot process it seems.

pepijndevos:
These are all part of https://github.com/stnolting/neorv32/pull/1086 and adding `BOOT_MODE_SELECT      => 1,` seems to have fixed it.

The rest of the todo items for this PR still remain.

pepijndevos:
The SoC now starts with https://github.com/stnolting/neorv32/pull/1211 so

TODO:

- [ ] update git hash after merged
- [ ] hook up JTAG

Two other things I'm not happy about but unsure how to fix:

- VHDL is synthesized for every core
- Atomics completely hang the system

stnolting:
Sorry for closing...
But how did I close this by merging a PR in neorv32?! 🤔

pepijndevos:
I think because I said "fixes" it marked that PR as "fixing" this PR.

Anyway, this should work now with the caveat that not all variants work well with more than one core due to cache coherency.

There also appears to be a feature/bug with LiteX wishbone where atomic operations can hang the system. Setting the SoC bus to AXI kind of resolves this, but with more than two cores atomic operations are no longer atomic. See https://github.com/stnolting/neorv32/pull/1211#issuecomment-2741642720 for details.

enjoy-digital:
@pepijndevos, @stnolting: Happy to merge if things are working. From what I understand, if there is still a specific issue, if you provide the steps to reproduce ideally in with litex_sim, I could have a look.

enjoy-digital:
@pepijndevos: Thanks for the CPU update, I'm merging this to avoid blocking things and created https://github.com/enjoy-digital/litex/issues/2229 to have a look at the remaining issue.

PR #2209, opened 2025-03-16
copy picolibc to build folder, as it gets modified while building

TuxCoder:
this allows sharing or read only access to the python package

I face a build error while I tried to build it in nixos, as the system is there read-only by design.
The build script in `pythondata-software-picolibc` modifies the source code included.
This could also lead to conflicts when building multiple targets at once.

This patch copy the source code to the build folder to have a local copy.

Here is a more or less working example for an cyclone10lp - cyc1000 board with this patch:

https://git.o-g.at/FPGA/cyc1000_tests

A bit hacked together, as I'm still learning / testing.

### DIFF
--- a/litex/soc/integration/builder.py
+++ b/litex/soc/integration/builder.py
@@ -189,8 +189,13 @@ class Builder:
         picolibc_directory    = get_data_mod("software", "picolibc").data_location
         compiler_rt_directory = get_data_mod("software", "compiler_rt").data_location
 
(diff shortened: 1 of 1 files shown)


TuxCoder:
one ref to were the source gets modified:
https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libc/Makefile#L42

not sure if there are others

Issue #2107, opened 2024-10-26
Cant have multiple Instances of Custom VHDL core


Haron123:
The **Error Message** i get is

> own_periphs/uart/UART_1.v:1: ERROR: Re-definition of module `$abstract\uart_clk_div_17_1'!

When i try to instantiate multiple of the same cores, in the **Target file** i defined them like this:
```
self.uart_logging = CustomUart(self.platform, platform.request("uart_logging"), sys_clk_freq, 115200)
self.irq.add("uart_logging", use_loc_if_exists=True)

self.uart_ice40 = CustomUart(self.platform, platform.request("uart_ice40"), sys_clk_freq, 115200)
self.irq.add("uart_ice40", use_loc_if_exists=True)
```

The **VHDL To Verilog** conversion looks like this
```python
# Convert to Verilog
self.vhd2v_converter = VHD2VConverter(platform,
	top_entity = "UART",
	build_dir = os.path.abspath(os.path.dirname(__file__)),
	params = self.parameters,
	force_convert = True,
	add_instance = True,
	work_package = "uart",
)

self.sources_path = os.path.abspath(os.path.dirname(__file__)) + "/comp/";

self.sources = [
	# uart.vhd
	"uart.vhd",

	# comp
	"uart_debouncer.vhd",
	"uart_parity.vhd",
	"uart_rx.vhd",
	"uart_tx.vhd",
	"uart_clk_div.vhd",
]

self.vhd2v_converter.add_sources(self.sources_path, *self.sources)
```

Does anyone know what could cause it? 

### Debug Attempts:
I know that it works when i only use one VHDL File, so it seems to be an issue with converting multiple VHDL Files. I looked into the VHD2V Code and apparently it only renames the TOP Entity, i didnt manage to change the code to rename it when theres multiple subentities.

Thanks in advance for any attempts to help.

Haron123:
Im still stuck on this, does anybody know what could cause it ?

pepijndevos:
I'm facing the same problem.

The cause is that the VHDL converter only renames the top instance when there are multiple instances. Any other modules will have collisions.

https://github.com/enjoy-digital/litex/blob/b1eb23bec2ae2c9730b71a740d62d015579fa30c/litex/build/vhd2v_converter.py#L204

PR #2213, opened 2025-03-24
flatten ghdl output with yosys to avoid conflicts

pepijndevos:
fixes https://github.com/enjoy-digital/litex/issues/2107 by feeding the ghdl output to yosys to flatten it to avoid naming conflicts

This is still a kinda dumb way to have more than one VHDL instance but it works...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,12 @@ jobs:
           sudo python3 litex_setup.py --gcc=openrisc
           sudo python3 litex_setup.py --gcc=powerpc
 
+      # Install OSS CAD Suite (Fixed version to avoid issue with GHDL)
+      - name: Install OSS CAD Suite
+        uses: YosysHQ/setup-oss-cad-suite@v3
+        with:
+          version: '2025-04-17'
+
       # Build / Install GHDL
       - name: Build GHDL
         run: |
--- a/litex/build/vhd2v_converter.py
+++ b/litex/build/vhd2v_converter.py
@@ -7,6 +7,8 @@
 
 import os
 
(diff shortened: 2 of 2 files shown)


trabucayre:
Hi @pepijndevos.
Sorry for the delay.
I have found why the CI fails: yosys must be installed, not using package manager (the command you uses is not supported) but from source, see [this test](https://github.com/trabucayre/litex/commit/c866bfc079f4de17f5e3d44554417ec677775278)

What do you think about a test for yosys to bypass this step if not present? It's avoid to adds a new dependency (with a message to explain benefits to install yosys).

Another thing to test is:
```python
tools.replace_in_file(verilog_out, f"\\", f"ghdl_") # FIXME: GHDL synth workaround, improve.
```
Removed here: this line was added because with CAN IP ghdl generates a code where some signals have a conflict (signal name vs Verilog's reserved name).

Thanks

pepijndevos:
I don't really have the bandwidth to take this forward right now because my project got binned.

It seems to me that yosys is the only toolchain that doesn't support vhdl. I'm not sure why this is using force=true but ideally I think this whole vhdl crutch only makes sense when using yosys.

trabucayre:
Indeed, with most of the toolchains vhdl files are directly used and ghdl bypassed.
Using `force_convert` is usually set to `True` to validates the conversion (ie be sure a VHDL code will works with all toolchains (including verilator)).

trabucayre:
Applied with two modifications:
- added `oss-cad-suite` to CI: required to have yosys
- `yosys` is optional: if not found or if version is too old, flatten is bypassed and original code is used

Thanks @pepijndevos

PR #2214, opened 2025-03-24
Fix TFTP PUT: Properly Handle OACK for Option Negotiation

gvillanovanm:
## Brief

This change fixes an issue in the TFTP client where the OACK (Option Acknowledgment) from the server was not fully handled during a file upload (tftp_put). When the server sends an OACK in response to a WRQ with options (like a non-standard block size), the client only sent an ACK(0) without updating its state (e.g., last_ack and data_port). This prevented the client from proceeding with sending data blocks, resulting in a 0-byte file on the server. The fix ensures that upon receiving an OACK, the client sets last_ack to 0 and captures the server's data port, allowing the data transfer to continue as expected.


enjoy-digital:
Thanks @gvillanovanm!

PR #2218, opened 2025-03-28
axi: Wishbone2AXILite: increase speed by one cycle

maass-hamburg:
use axi bus one cycle earlier.


maass-hamburg:
@enjoy-digital please have a look

enjoy-digital:
Thanks @maass-hamburg, the extra cycle was a compromise for code simplicity. We can merge the changes but it would be good in the long term to try to remove the logic duplication (I'll also try to look at this).

PR #2225, opened 2025-04-12
Fix add_controller() call

grant-olson:
The `add_controller` alias for `add_master` incorrectly includes self in the method call, which causes a runtime error complaining if a user runs `self.bus.add_controller(name="x", controller=controller)`It just needs a simple fix to work correctly.


enjoy-digital:
Thanks @grant-olson.

PR #2228, opened 2025-04-24
.github/workflows/ci.yml: integrates oss-cad-suite and remove GHDL Build/Install

trabucayre:
_oss-cad-suite_ provides a complete tool set including `yosys` and `GHDL`. It also provides an action to simplify integration in _CI_.
This approach:
- avoids to build `GHDL` all the time: `GHDL` shipped with _oss-cad-suite_ is daily built with master branch (similar to the current approach in `LiteX` CI). This allows to reducing CI delays (around 4m/10% for my fork's CI)
- by using `yosys` (based on master branch too) from this package fixes #2213 because `yosys` is by default required and ubuntu's version is too old and fails. This also avoid a big overhead due to `yosys` compile time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,15 +43,9 @@ jobs:
           sudo python3 litex_setup.py --gcc=openrisc
           sudo python3 litex_setup.py --gcc=powerpc
 
-      # Build / Install GHDL
-      - name: Build GHDL
-        run: |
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied differently with https://github.com/enjoy-digital/litex/commit/e614d723232c7f7f973212b15f408a691a791d06

PR #2232, opened 2025-04-28
bios: make libliteeth/udp more portable

maass-hamburg:
By moving from platform dependent types like `unsigned int` to cross-platform types like `uint32_t`, the behavior will remain as expected even on different platforms/compilers.

I got some warnings when compiling for the simulator that hinted at a potential problem with the previous code.

cherry-pick from #2070


trabucayre:
Applied. Thanks @maass-hamburg

PR #2233, opened 2025-04-30
soc/cores/cpu/zynq7000/core.py: add_axi_hp_slave: added clock_freq parameters (optional), updated ps7_config to enable bus, to configure ID Width and set clock frequency

trabucayre:
when calling `dd_axi_hp_slave` method a new `AXIInterface` is created and connected to the PS7 `S_AXI_HPx`.
But the bus isn't enabled at tcl level, and only `PCW_S_AXI_HPx_DATA_WIDTH` is configured.

This PR adds:
- a new argument to specify clock_frequency (optional). When present `PCW_S_AXI_HP{n}_FREQMHZ` is added to the tcl list
- `PCW_USE_S_AXI_HPx` is added to tcl list to enable the bus
- `PCW_S_AXI_HPx_ID_WIDTH` is also added with a value of 6 to be coherence with the AXIInterface.

Tested with digilent_arty_z7 platform

### DIFF
--- a/litex/soc/cores/cpu/zynq7000/core.py
+++ b/litex/soc/cores/cpu/zynq7000/core.py
@@ -368,7 +368,7 @@ class Zynq7000(CPU):
 
     # AXI High Performance Slave -------------------------------------------------------------------
 
-    def add_axi_hp_slave(self, clock_domain="ps7", data_width=64):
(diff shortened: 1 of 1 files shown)


enjoy-digital:

PR #2234, opened 2025-04-30
Fix uptime command in bios

Ka-zam:
Simple fix to display correct uptime


enjoy-digital:
Thanks @Ka-zam, this is merged.

PR #2235, opened 2025-05-01
build/nextpnr_wrapper.py: directly handles himbaechel architecture

trabucayre:
Recent `nextPNR`'s evolutions have introduces *himbaechel* architecture. This one supports FPGAs like **gowin**, **gatemate**, **xilinx**, ...
For *apicula* toolchain current solution is to force nextPNR name in `finalize` method.

But since `nextpnr-gowin` is now deprecated it make more sense to moves this logic in `nextpnr_wrapper.py`: `apicula.py` know informations about gowin devices, `nextpnr_wrapper.py` know for gowin architecture the executable is called `nextpnr-himbaechel`.

A second `nextPNR` evolution is the introduction of `HIMBAECHEL_SPLIT` option:
- when unset, or set to OFF, all selected `uarch` are included into a single executable
- when set each `uarch` is supported by a specific executable (`nextpnr-himbaechel-XXX`) where XXX is the `uarch` name

This PR:
- adapts the code to move naming translation at `nextpnr_wrapper.py` level
- adds a test to select nextPNR executable name (`nextpnr-himbaechel` or `nextpnr-himbaechel-uarch`).

### DIFF
--- a/litex/build/gowin/apicula.py
+++ b/litex/build/gowin/apicula.py
@@ -66,9 +66,6 @@ class GowinApiculaToolchain(YosysNextPNRToolchain):
 
         YosysNextPNRToolchain.finalize(self)
 
-        # family is gowin but NextPNRWrapper needs to call 'nextpnr-himbaechel' not 'nextpnr-gowin'
-        self._nextpnr.name = "nextpnr-himbaechel"
-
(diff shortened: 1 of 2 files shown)


enjoy-digital:

PR #2243, opened 2025-05-13
soc/cores/cpu/zynq7000/core.py: added support for openXC7/yosys/nextPNR

trabucayre:
This is WIP to uses zynq7000 with opensource toolchain:
- CPU IOs are replaced by simple signals when toolchain is not Vivado
- IOs have different name pattern: before the `Instance` creation names are adapted
- PS name can't be changed with non Vivado toolchain


enjoy-digital:
Thanks @trabucayre. I've rebased the PR on current master and resolved the conflict while keeping the change focused on the non-Vivado/OpenXC7 PS7 pad handling. CI has restarted; if this updated version is convenient for you, we can merge it once the checks are green.

PR #2247, opened 2025-05-22
build: efinix: support efinity 2025.1

maass-hamburg:
support efinity 2025.1

### DIFF
--- a/litex/build/efinix/efinity.py
+++ b/litex/build/efinix/efinity.py
@@ -284,7 +284,9 @@ class EfinityToolchain(GenericToolchain):
         et.SubElement(device_info, "efx:timing_model", name=self.platform.timing_model)
 
         # Add Design Info.
-        design_info = et.SubElement(root, "efx:design_info")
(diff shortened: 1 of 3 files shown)


maass-hamburg:
@enjoy-digital this unfortunately will break support for older efinity versions, but due to the fact, that the current efinity version is available for free, it's not that big of a problem. This will also bring support for setting the build and compile parameters for the use side, just change the `kwargs` in the build function of your platform class

enjoy-digital:
Thanks @maass-hamburg for the additional info. I'll have a log to see how to get the logs in the console without touching `efx_run.py` (that would evolves at each release).

enjoy-digital:
@maass-hamburg: The PR is merged, thanks! With https://github.com/enjoy-digital/litex/commit/f5e5514b32466f947e9968b8a93e18c3391ef7f3 and https://github.com/enjoy-digital/litex/commit/83a14dd643430466d14a27584580832ab6077440 we now also get the build log redirected to stdout and colored during the build as before.

PR #2248, opened 2025-05-23
Gatemate: handle SDRInput, SDROutput and SDRTristate

mmicko:
These are required for SD card interface and also Ethernet to work.

Note: Will create PR in litex-boards that is requiring these changes


trabucayre:
Applied! Thanks @mmicko

PR #2251, opened 2025-05-31
Various fixes to DTS generation script

spersvold:
- Add support for L1/L2 cache topology for NaxRiscv
 - Fix DT naming for USB OHCI
 - Fix L1$ size reporting for all configs
 - Add L2 cache to DTS if present in config
 - Add local MAC address to ethernet if given

### DIFF
--- a/litex/soc/cores/cpu/naxriscv/core.py
+++ b/litex/soc/cores/cpu/naxriscv/core.py
@@ -141,11 +141,11 @@ class NaxRiscv(CPU):
             NaxRiscv.gcc_triple           = CPU_GCC_TRIPLE_RISCV64 if xlen == 64 else CPU_GCC_TRIPLE_RISCV32
             NaxRiscv.linker_output_format = f"elf{xlen}-littleriscv"
         if args.cpu_count:
-            NaxRiscv.cpu_count = args.cpu_count
+            NaxRiscv.cpu_count = int(args.cpu_count)
         if args.l2_bytes:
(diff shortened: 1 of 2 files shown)


litex/soc/cores/cpu/naxriscv/core.py
```diff
@@ -399,6 +399,18 @@ def add_soc_components(self, soc):
         soc.add_config("CPU_ISA", NaxRiscv.get_arch())
         soc.add_config("CPU_MMU", {32 : "sv32", 64 : "sv39"}[NaxRiscv.xlen])
 
+        # Constants for cache so we can add them in the DTS.
+        soc.add_config("CPU_DCACHE_SIZE", 16384)
+        soc.add_config("CPU_DCACHE_WAYS", 4)
+        soc.add_config("CPU_DCACHE_BLOCK_SIZE", 64) # hardwired?
```
Dolu1990, line 405: Right, let's keep block size at 64, this is the "golden standard" for memory coherent stuff :)

spersvold: For the NaxRiscv specifically there is no option to change the cache block size, and it is 64 :)

spersvold: Same with the sizes and ways for this CPU, the Scala code can be adjusted but its just not done yet

Dolu1990: Hooo right, VexiiRiscv is much cleaner in that regard XD

PR #2252, opened 2025-06-01
Add dummy interrupt class so we can use standard IRQ allocation

spersvold:


enjoy-digital:
Thanks @spersvold, this is merged.

PR #2254, opened 2025-06-02
colognechip: remove forcing ram_style=distributed

mmicko:
This change lets yosys `memory_libmap` choose proper memory on it's own when `"distributed"` is asked for. This makes CPUs like `vexriscv_smp` to be synthesized.


trabucayre:
Applied. Thanks @mmicko !

PR #2261, opened 2025-06-16
Add State Save and Load Functions for Verilator Simulation

mohammadshahidzade:
Add state save/load functions for Verilator simulation

- Implemented functions to save and restore simulation state.
- Added 'saveable' target in Makefile to support state saving/loading.
- Enables easier debugging of long simulations or late bugs (e.g., Linux boot).
- Note: This currently causes slower simulation performance.
- Please review and merge if you can identify the cause of slowdown.
- I will continue investigating and update with findings.

This feature allows checkpointing the simulation state to avoid rerunning from the start.

### DIFF
--- a/litex/build/sim/core/Makefile
+++ b/litex/build/sim/core/Makefile
@@ -17,7 +17,7 @@ else
 	LDFLAGS += -lpthread -Wl,--no-as-needed -ljson-c -lz -lm -lstdc++ -Wl,--no-as-needed -ldl -levent $(if $(VIDEO), -lSDL2)
 endif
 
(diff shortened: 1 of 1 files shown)


mohammadshahidzade:
I believe the changes are working now—you can go ahead and test them. I also modified the loading to occur after a user-specified time, since it makes more sense for it to happen once the CPU reset is complete.

enjoy-digital:
Thanks @mohammadshahidzade, I'll try to test/review this soon.

enjoy-digital:
@mohammadshahidzade:I haven't been able to test it yet, but I've just reviewed the code. It looks good overall, and the feature appears well-separated from other parts of the codebase, so it seems reasonable to merge it. If any issues arise with regular use cases, I'll report them—and I'll also share feedback once I perform more tests with the functionality. Thanks a lot for the PR!

PR #2263, opened 2025-06-18
bios: add broadcast support to libliteeth/udp

maass-hamburg:
This adds support for UDP broadcasts and a helper function to get the
currently set IP. These changes are required for BOOTP support.

now protected with a ifdef, so if its nat activated its not compiled in


maass-hamburg:
@enjoy-digital please take a look. It is originally from https://github.com/enjoy-digital/litex/pull/2070 and needed for bootp support, that we now put into a private module.

enjoy-digital:

PR #2264, opened 2025-06-18
Tests/test_integration: Fix log output on failure + Add vexiiriscv to unit tests

cklarhorst:
This PR adds vexiiriscv to the test_integration unit test.

While working on this, I noticed the following issues:
- The log output was broken when a CPU failed to boot. This was due to incorrect file mode settings when opening the log file.
- The previously recommended commit for vexiiriscv fails to compile (at least with default settings) due to the following error:
> Error:  Note: the super classes of <$anon: spinal.lib.bus.amba4.axi.sim.Axi4WriteOnlyMonitor> contain the following, non final members named onWriteStart:
> Error:  def onWriteStart(address: BigInt,id: Int,size: Int,len: Int,burst: Int,cache: Int): Unit
> Error:          override def onWriteStart(address: BigInt, id: Int, size: Int, len: Int, burst: Int): Unit = {
- To fix this, I updated the recommended commit to the latest dev version that successfully compiles and passes the test.

@Dolu1990  Is there a more suitable commit you would recommend instead?

### DIFF
--- a/litex/soc/cores/cpu/vexiiriscv/core.py
+++ b/litex/soc/cores/cpu/vexiiriscv/core.py
@@ -156,7 +156,7 @@ class VexiiRiscv(CPU):
         vdir = get_data_mod("cpu", "vexiiriscv").data_location
         ndir = os.path.join(vdir, "ext", "VexiiRiscv")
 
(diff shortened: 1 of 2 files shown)


trabucayre:
@Dolu1990 if it's fine to you and if this also fixes log file I could merge it?
Are you agree?
Thanks

Dolu1990:
@trabucayre Look good to me ^^

PR #2270, opened 2025-06-27
csr: fix and support ordering

maass-hamburg:
fix and support ordering

### DIFF
--- a/litex/soc/doc/__init__.py
+++ b/litex/soc/doc/__init__.py
@@ -101,7 +101,8 @@ def generate_docs(soc, base_dir,
             region         = region,
             module         = module,
             submodules     = submodules,
-            csr_data_width = soc.csr.data_width)
+            csr_data_width = soc.csr.data_width,
+            csr_ordering   = soc.csr.ordering)
         if documented_region.name in interrupts:
             documented_region.document_interrupt(
                 soc, submodules, interrupts[documented_region.name])
--- a/litex/soc/doc/csr.py
+++ b/litex/soc/doc/csr.py
@@ -63,7 +63,7 @@ class DocumentedCSR:
             f.description = self.trim(f.description)
 
 class DocumentedCSRRegion:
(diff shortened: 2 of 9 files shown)


maass-hamburg:
@enjoy-digital what do you think about it?

I struggled with the sdram.c part, inside the `sdram_write_read_check_test_pattern()` function
effective values of the old code:
```
SDRAM_PHY_XDR = 1:
pebo = (2 * DFII_PIX_DATA_BYTES) - 1 - (module * SDRAM_PHY_DQ_DQS_RATIO)/8;

SDRAM_PHY_XDR = 2:
nebo = (DFII_PIX_DATA_BYTES / 2) - 1 - (module * SDRAM_PHY_DQ_DQS_RATIO)/8;
pebo =  DFII_PIX_DATA_BYTES - 1 - (module * SDRAM_PHY_DQ_DQS_RATIO)/8;
```

and with module being iterated from 0 to `SDRAM_PHY_MODULES`, it seems like `pebo` could get bigger than `DFII_PIX_DATA_BYTES` and therefore outside of the `tst` array, which seems wrong in the old code.

enjoy-digital:
Thanks @maass-hamburg! This looks good overall, but the required sdram changes are indeed tricky and I would need to check this more carefully since when testing on a target with SDRAM, the changes break the calibration. What about keeping the previous ordering for now in `_csr_rd_buf` and `_csr_wr_buf` and then remove the software application changes in `sdram.`c/`sdcard.c` and merge this PR. And work together on another PR to change  `_csr_rd_buf`  and `_csr_wr_buf and have a correct change for sdram.c?

enjoy-digital:
Thanks @maass-hamburg, I'll have another look and will redo some tests.

enjoy-digital:
@maass-hamburg: I just did some tests and the regression is indeed fixed, thanks! So we can merge.

PR #2271, opened 2025-07-01
json2dts_zephyr: update interrupt name to use the correct identifier

maass-hamburg:
remove `_core`suffix.


enjoy-digital:
Thanks @maass-hamburg.

PR #2281, opened 2025-07-10
soc: put litesdcard modules in a parent class for add_sdcard()

maass-hamburg:
Put the LiteSDCard modules in one class so that the CSRs
are in the same region.

also rename `irq` to `ev` so it will still work.

### DIFF
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -549,14 +549,14 @@ def generate_dts(d, initrd_start=None, initrd_size=None, initrd=None, root_devic
 
     # SDCard ---------------------------------------------------------------------------------------
 
-    if "sdcard_core" in d["csr_bases"]:
+    if "sdcard" in d["csr_bases"]:
         dts += """
             mmc0: mmc@{mmc_csr_base:x} {{
                 compatible = "litex,mmc";
-                reg = <0x{sdcard_phy_csr_base:x} 0x100>,
-                      <0x{sdcard_core_csr_base:x} 0x100>,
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @maass-hamburg, this looks good and going in the right direction. In the future we'll probably want to harmonize between the different cores and have integration methods directly in the cores as discussed in https://github.com/enjoy-digital/litesdcard/pull/41, but let's do things step by step.

enjoy-digital:
The CI was failing due to coreblocks CPU that has been disabled for now, the other tests are passing, so we can merge.

enjoy-digital:
@gsomlo: This change will not impact Linux-on-LiteX-VexRiscv project since we are generating the .dts directly from the json2dts script. But it will probably impact the Linux-on-LiteX-Rocket project since (I think) .dts are still generated manually, so will probably need to be updated.

gsomlo:
On Wed, Jul 23, 2025 at 11:49:45PM -0700, enjoy-digital wrote:
> ●enjoy-digital left a comment (enjoy-digital/litex#2281)
> 
> @gsomlo: This change will not impact Linux-on-LiteX-VexRiscv project since
> we are generating the .dts directly from the json2dts script. But it will
> probably > impact the Linux-on-LiteX-Rocket project since (I think) .dts
> are still generated manually, so will probably need to be updated.

Thanks for the heads-up! I ran some tests, updated some manually
generated dts files, and everything seems to once again work fine! :)

Best,
--Gabriel

enjoy-digital:
Great, thanks for the feedback @gsomlo!

PR #2285, opened 2025-07-15
json2dts_zephyr: set default irq prio to 1

maass-hamburg:
plic irq controller has a threshold, that is default at 0 and the priority has to be greater than 0, so set that to 1.


trabucayre:
@maass-hamburg Indeed. `coreblocks` CPU core fails. Looks like a mix between 3.9 (builder) and 3.11 (Amaranth -> Verilog conversion part).

enjoy-digital:

PR #978, opened 2021-07-24
Make bios respond to pings

michael-betz:
A bit hacky but works ...

Tested on Marble-board and verified with wireshark.

PR #2287, opened 2025-07-19
Make litex bios respond to ping requests

michael-betz:
A useful feature to check if the ethernet connection is working.

It can be tested in the emulator with:

```bash
litex_sim --with-ethernet

# in another terminal
ping 192.168.1.50
```
This replaces the previous pull request #978 which is a bit out of date.


michael-betz:
Hi Olli 👋 

many thanks for tracking this down and sorry for breaking linux-on-litex (I'm eager to try this out btw.)

I've opened a new pull request for the fix #2302

PR #2288, opened 2025-07-21
stream: SyncFIFO: export writable signal

fallen:


fallen:
Oh you're right, I was using this like a "normal fifo" I should use it like a stream.
Thanks!

PR #2289, opened 2025-07-22
interconnect: axi: improve

maass-hamburg:


diorcety:
Since 6c1b659bee434825fb4dd88ffe3fb5b90aac79d5 I have issue with overlapped region with multiple ethernet interfaces.
Shouldn't we add check_linker=linker on check_regions_overlap call on line 288? https://github.com/VOGL-electronic/gw__litex/blob/6c1b659bee434825fb4dd88ffe3fb5b90aac79d5/litex/soc/integration/soc.py#L288

PR #2290, opened 2025-07-23
build: efinix: efinity: hide excluded ios

maass-hamburg:
hide excluded ios, so they are not part of
the top modules ios, so that we don't get a
warning by the efinity that they are unused.

excluded ios are added, when a SDR or DDR Input, Output or Tristate is used.


maass-hamburg:
won't work currently due to a bug in efinity https://forum.efinix.net/t/isf-file-not-used-when-compiling-via-efx-run-py/3314

enjoy-digital:
The core hide-excluded-I/O behavior appears to already be present on current master, so this branch is now largely superseded and conflicts with current `efinity.py`. I recommend closing it and opening a focused follow-up only if there are still missing pieces.

PR #2293, opened 2025-07-25
build: efinix: common: use add_iface_io

maass-hamburg:
make use of add_iface_io for efinix


enjoy-digital (approved):
Thanks @maass-hamburg.

PR #2295, opened 2025-07-27
Fix efinix hyperram

trabucayre:
This PR:
- updates `HyperRAM` module call with missing parameters
- Fix `EfinixTristateImpl`: this class fails when signal is `TSTriple`
- updates PLL uses according to current usage

Note: This PR fix build, but has not being tested since it requires a board based on ti60f100


enjoy-digital:

PR #2299, opened 2025-08-01
efinix: clock: pll: use margin of create_clkout for check

maass-hamburg:
use margin of create_clkout for checking
the calculated freq to the target freq.


enjoy-digital:

PR #2300, opened 2025-08-01
efinix: clock: pll: add argument to create_clkout for nclkout

maass-hamburg:
add argument to create_clkout for nclkout, so
a specific clkout can be used.
Usefull for hard ip, that requires, that a specific
clkout is used and we don't want to use the clkouts
before.


enjoy-digital:

PR #2309, opened 2025-08-19
build/lattice/trellis.py: added argument to overrides bitstream's IDCODE

trabucayre:
Some ECP5 differs by the name but the die is the same. For example the 12F and 25F have same die.
It's thus possible to build (`nextpnr-ecp5`) a bitstream targeted for a 25F and load/write it to a 12F.
The only limitation is to have the correct IDCODE in the bitstream, otherwise the ECP5 will reject it.
`ecppack` with `--idcode` is able to overrides IDCODE when `.json` is converted to `.bit` allowing the bitstream to be accepted by the FPGA.

This PR adds `--ecppack-idcode` argument to allows user to specify the targeted FPGA IDCODE.

### DIFF
--- a/litex/build/lattice/trellis.py
+++ b/litex/build/lattice/trellis.py
@@ -41,6 +41,7 @@ class LatticeTrellisToolchain(YosysNextPNRToolchain):
         bootaddr       = 0,
         spimode        = None,
         freq           = None,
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #2310, opened 2025-08-20
build: io: Tristate: let SDR and DDR Tristates support i being None

maass-hamburg:
let SDR and DDR Tristates support input being None, just like the unregistered Tristate from migen.
Usefull when only the oe is needed, like for the scl in i2c.


enjoy-digital:
Thanks @maass-hamburg, this looks good and is merged.

PR #2311, opened 2025-08-20
build: io: efinix: DDR: Input: correct timing

maass-hamburg:
The input of the DDRInput and the DDRTristate
of the Efinix implementations was not matching,
these of the other implementations. like for the sim in
litex/build/sim/verilog/iddr_verilog.v
This fixes it. For Titanium/Topaz, this can be done by
useing DDIO_RESYNC_PIPE for in_reg, the Trion doesn't
support that, so we have to delay i1 manualy by one cycle.

The old variants with the wrong timing are still there for
those, who need them.

Contains: #2310


enjoy-digital:
Thanks @maass-hamburg, this looks good and is merged.

PR #2313, opened 2025-08-22
soc.py: litespi: improve/fix add_spi_flash

maass-hamburg:
- only activate QPI mode in the flash, when we are going to use it
- expose number argument of platform.request to be able to reference instances of a flash with the same name
- expose **kwargs for phy module, so we can set it from add_spi_flash
- after enabling quad mode, wait until it finished, freq init might not work correctly (the first crc might be wrong) if we don't wait.

changes are also applied to add_spi_ram for consistency


enjoy-digital:

PR #2318, opened 2025-08-28
Agilex support update

trabucayre:
This PR aims to update agilex5 support with recent quartus pro version (25.1.1)
- First commit: removes `Agilex5AsyncResetSynchronizerImpl` to uses default implementation. `altera_std_synchronizer_nocut` is only available with `Platform designer` and corresponding code is more or less similar to uses of `DFF`
- Second commit updates `Agilex5SDRTristateImpl`. With new quartus version `tennm_ph2_io_ibuf` and `tennm_ph2_io_obuf` must have all parameters explicitly provided to avoid build failure
- Third commit adds `Agilex5TristateImpl` required when HyperRAM core is used. The code is quite similar to `Agilex5SDRTristateImpl`
- Last commit adds argument to select between `quartus_cpf` and `quartus_pfg`. Last one is required to convert sof to rbf for Agilex target (svf can't be produces at all).

Tested with Arrow AXE5000 board

### DIFF
--- a/litex/build/altera/common.py
+++ b/litex/build/altera/common.py
@@ -263,6 +263,50 @@ class Agilex5SDRTristate(Module):
     def lower(dr):
         return Agilex5SDRTristateImpl(dr.io, dr.o, dr.oe, dr.i, dr.clk)
 
+# Agilex5 Tristate ---------------------------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


enjoy-digital:

PR #2319, opened 2025-08-28
cores: uart: switch to EventSourceLevel irq

maass-hamburg:
switch to EventSourceLevel irq, this way we don't have to
reset pending.

Should still be compatible with the existing drivers in
linux and zephyr.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -222,8 +222,8 @@ class UART(LiteXModule, UARTInterface):
         self._rxempty = CSRStatus(description="RX FIFO Empty.")
 
         self.ev    = EventManager()
-        self.ev.tx = EventSourceProcess(edge="rising")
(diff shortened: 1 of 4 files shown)


enjoy-digital:
@maass-hamburg: Thanks, this looks good. I'll just do a test with Linux before merging.

Dolu1990:
Hi ^^

We got some weird issues related the UART interrupt in linux before this PR.
Let me know if on your side all is working well (with xor without this PR)

(https://github.com/litex-hub/linux-on-litex-vexriscv/issues/432#issuecomment-3302499099)

enjoy-digital:
@maass-hamburg: I just tested this PR with a Linux boot and the UART driver is indeed still working, so let's merge.
@Dolu1990: Could you redo a test with this merged? or tell me what you were testing previously that I do a test? Thanks.

PR #2320, opened 2025-08-29
interconnect:axi_full: Initialize AXIInterfaces with correct id_width

mkuhn99:
In the AXIInterconnectShared and AXICrossbar classes AXIInterfaces are always initialized with id_width = 1, which causes id errors.

Fixed by initializing them with the biggest id_width of all master Interfaces (like data_width and address_width).


enjoy-digital:
Thanks @mkuhn99, this is merged.

PR #2327, opened 2025-09-05
json2dts: Fix for sdcard rename

gatecat:
Otherwise this wasn't generating `vreg_mmc` and there would be an error about the reference to it not being found.


cheyao:
Thanks a lot for the patch - this bug has actually been puzzling me for the past few days and I thought I was doing something wrong :sweat_smile:

PR #2331, opened 2025-09-22
cores: uart: only use self._rxtx.we when rx_fifo_rx_we

maass-hamburg:
only use self._rxtx.we when rx_fifo_rx_we and not
both (not also the clear signal)


maass-hamburg:
@enjoy-digital something unrelated failed in ci

PR #2336, opened 2025-09-26
More quality of life improvements for Agilex family of devices

spersvold:
Added a new object for PLL SDC statements that gets applied earlier in the generated SDC file as a replacement for the "derive_pll_clocks" command which are no longer supported in the Agilex family.

Also some cleanups/fixes related to clock naming in the SDC file and false path statements.


enjoy-digital:
Thanks @spersvold, this looks good.

Issue #2337, opened 2025-09-27
Wrong ROM size for 64-bit CPUs


franos-cm:
Hi,

I have been using Litex in an academical project for some months now, and I initially had problems loading firmware into the ROM of my Rocket SoC. Specifically, the SoC simulation would crash before it even began, displaying this message on the terminal:

``` bash
[ethernet] loaded (0x617e6bf2c2f0)
[serial2tcp] loaded (0x617e6bf2c2f0)
[xgmii_ethernet] loaded (0x617e6bf2c2f0)
[gmii_ethernet] loaded (0x617e6bf2c2f0)
[clocker] loaded
[serial2console] loaded (0x617e6bf2c2f0)
[jtagremote] loaded (0x617e6bf2c2f0)
[spdeeprom] loaded (addr = 0x0)
MDEBUG: Save time: -1, load_time: 0
[clocker] sys_clk: freq_hz=1250000, phase_deg=0
Found port 4327
%Error: sim_rom.init:325: $readmem file address beyond bounds of array
Aborting...
run_sim.sh: line 1: 128024 Aborted                 (core dumped) obj_dir/Vsim
```

[Here is a repo with a minimal demo of the issue.](https://github.com/franos-cm/litex-rom-bug-demo)

Having worked around this issue for some time, I finally decided to debug it. The origin of the problem is that, **in the `SoCCore` initializer**, if `integrated_rom_init` is passed as a `str` (i.e. a path to a binary file), **Litex underestimates the total ROM size**:

https://github.com/enjoy-digital/litex/blob/ffb100c23d00a01d76a430783869a228d1f94340/litex/soc/integration/soc_core.py#L164-L171

`get_mem_data()` returns the instructions in the binary file as a `List[int]`, and `integrated_rom_size` should in turn correspond to the number of _bytes_ in the ROM file. **The multiplication by 4 implicitly assumes a 32-bit CPU.** This value is then used to instantiate the ROM block:

https://github.com/enjoy-digital/litex/blob/ffb100c23d00a01d76a430783869a228d1f94340/litex/soc/integration/soc_core.py#L231-L238

When using a 64-bit CPU (like Rocket), building the SoC generates the rom.init file with its correct size, but the resulting ROM register array in the Verilog file is half the expected size, since it assumed that each instruction is 32 bits long. This creates the `%Error: sim_rom.init:325: $readmem file address beyond bounds of array` error observed.

[...]

PR #2338, opened 2025-09-27
Fixed ROM size and endianness

franos-cm:
This is my first time opening a PR in an open-source project, so sorry in advance if I am doing something wrong.

This PR first fixes the issue in #2337, which is simply a one line change in the `SoCCore` initializer code:

https://github.com/enjoy-digital/litex/blob/ffb100c23d00a01d76a430783869a228d1f94340/litex/soc/integration/soc_core.py#L164-L171

where we change that last line to:

``` python
integrated_rom_size = (bus_data_width // 8) * len(integrated_rom_init)
```
and that's it!

&nbsp;

**But... since we are here,** we can also solve the `FIXME` shown by delaying the `integrated_rom_init` and `integrated_rom_size` assignments until after we add the CPU to the SoC, so we can reference its endianness easily. Something like:

``` python
self.add_cpu(
    name          = str(cpu_type),
    variant       = "standard" if cpu_variant is None else cpu_variant,
    reset_address = None if integrated_rom_size else cpu_reset_address,
    cfu           = cpu_cfu)

...

if integrated_rom_region_valid:
    integrated_rom_init = get_mem_data(integrated_rom_init,
        endianness = self.cpu.endianness, # fixed
        data_width = bus_data_width
    )
    integrated_rom_size = (bus_data_width // 8) * len(integrated_rom_init)
```

Evidently, the problem is that the `add_cpu()` function call is using `integrated_rom_size` to set the reset address. To deal with that, instead of using it directly, we create a flag _before_ calculating `integrated_rom_size`, but that serves the same purpose in the context of the `add_cpu()` function; that is, identifying if ROM exists.

Looking at the original `SoCCore` initializer code, ROM exists only if:

1.  we provide a binary file that is not empty (`get_mem_data()` will check for that), as shown in the first snippet;

2. **or** we provide a non-zero value for `integrated_rom_size` directly on the `SoCCore` initializer:
https://github.com/enjoy-digital/litex/blob/ffb100c23d00a01d76a430783869a228d1f94340/litex/soc/integration/soc_core.py#L54-L75

3.  **and** we are using a CPU that is not hard:
https://github.com/enjoy-digital/litex/blob/ffb100c23d00a01d76a430783869a228d1f94340/litex/soc/integration/soc_core.py#L173-L176

We can then encode these checks into our new `integrated_rom_exists` flag:

``` python
# ROM
# Check if CPU can support ROM.
hard_cpus_list = ["zynq7000", "zynqmp", "eos_s3"]
integrated_rom_disabled = (cpu_type is None) or (cpu_type in hard_cpus_list)
# Check if a ROM binary file was provided and that it is not empty.
integrated_rom_region = get_mem_regions(integrated_rom_init, offset=0) if isinstance(integrated_rom_init, str) else None
integrated_rom_region_valid = integrated_rom_region and get_mem_region_size(integrated_rom_region)
# Check if ROM exists.
integrated_rom_exists = (integrated_rom_region_valid or integrated_rom_size) and not integrated_rom_disabled
```

And then finally we change the line in `add_cpu()` to:
``` python
reset_address = None if integrated_rom_exists else cpu_reset_address,
```

This PR makes these changes, along with the creation of the `get_mem_region_size()` function used above, which is really just a block from `get_mem_data()` put into a separate function, so that we can assert that ROM size is non-zero before computing `integrated_rom_init` from the binary file path.

I was able to correctly build and simulate an SoC (from the [repo where I demoed the original issue](https://github.com/franos-cm/litex-rom-bug-demo)) after making these changes to my Litex fork.

Hope this helps! I honestly don't know how opening and closing PRs work for this (or, for that matter, any other open-source) project, so tell me if I need to do anything else :blush:

### DIFF
--- a/litex/soc/integration/common.py
+++ b/litex/soc/integration/common.py
@@ -39,6 +39,17 @@ def get_mem_regions(filename_or_regions, offset):
             regions = {filename: f"{offset:08x}"}
     return regions
 
+def get_mem_region_size(regions, offset=0):
+    # Determine memory region data size.
+    data_size = 0
+    for filename, base in regions.items():
+        if not os.path.isfile(filename):
+            raise OSError(f"Unable to find {filename} memory content file.")
+        data_size = max(int(base, 16) + os.path.getsize(filename) - offset, data_size)
+
+    assert data_size > 0
+    return data_size
+
 def get_mem_data(filename_or_regions, data_width=32, endianness="big", mem_size=None, offset=0):
     assert data_width % 32 == 0
     assert endianness in ["big", "little"]
@@ -50,13 +61,9 @@ def get_mem_data(filename_or_regions, data_width=32, endianness="big", mem_size=
     # Create memory regions.
     regions = get_mem_regions(filename_or_regions, offset)
 
-    # Determine data_size.
-    data_size = 0
-    for filename, base in regions.items():
-        if not os.path.isfile(filename):
-            raise OSError(f"Unable to find {filename} memory content file.")
-        data_size = max(int(base, 16) + os.path.getsize(filename) - offset, data_size)
-    assert data_size > 0
+    # Get data_size.
+    data_size = get_mem_region_size(regions, offset)
+
     if mem_size is not None:
         assert data_size < mem_size, (
             "file is too big: {}/{} bytes".format(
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Closing since solved as discussed. But thanks a lot for your PR!

PR #2340, opened 2025-10-02
Zynq7000 improve

trabucayre:
**Xilinx Zynq7000** device contains a **ARM CortexA9 CPU** (**PS**).
This **PS** has its set of hard controlers as usual with a _SoC_, these may be enabled and connected via **PS** pads or via **PL**(_FPGA_) _EMIO_ pads.

- Commits 1 improves EMIO GPIOs support to allows user providing pads, `Signal` or `TSTriple`
- Commit 2-4 adds _UART_, _SPI_ and _I2C_ controlers supports with _EMIO_ vs **PS** pads configuration. _constants_ are also updated to reflect controlers status and if it configured for _EMIO_
- Commit 5: updates `vivado.py` to allows .bit.bin creation: when the bitstream is loaded via the Linux executed on PS, this file must be signed with `bootgen`. It is, of course, possible to do this step manually but it's a big boring:
  - a bif file must be filled with .bit path
  - `bootgen` must be called with correct arguments.


enjoy-digital:
Thanks @trabucayre, this looks good. Let's merge!

PR #2341, opened 2025-10-06
soc/cores/video.py: VideoGowinHDMIPHY: allows to specify if _p/_n are true LVDS or if LVDS must be emulated

trabucayre:
With Gowin IDE using `ELVDS_OBUF` for true differential pins is fine and build end successfully.
But with `apicula` a `TLVDS_OBUF` must be used otherwise the build end with:
```bash
ValueError: Cannot place hdmi_data2_p_OBUF_O at IOB40A - it is a true lvds pin
```

This failure may be observed for the `sipeed_tang_nano_20k.py` target when `VideoGowinHDMIPHY` is enabled.

This PR adds an additional (and optional) CTOR's parameter to specify which type of primitive must be used:
- `ELVDS_OBUF` when `true_lvds == False` (default)
- `TLVDS_OBUF` when `true_lvds == True`

### DIFF
--- a/litex/soc/cores/video.py
+++ b/litex/soc/cores/video.py
@@ -881,17 +881,22 @@ class VideoHDMIPHY(LiteXModule):
 # HDMI (Gowin).
 
 class VideoGowinHDMIPHY(LiteXModule):
-    def __init__(self, pads, clock_domain="sys", pn_swap=[]):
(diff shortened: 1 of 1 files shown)


trabucayre:
Indeed. Fixed.
Thanks @enjoy-digital

PR #917, opened 2021-05-17
Add utilities for debugging DRAM problems

jedrzejboczar:
This PR mainly adds the `sdram_debug` BIOS command that can be used to better analyze DRAM problems. I am using it when analyzing DRAM problems and thought that maybe it could be useful to merge with Litex (possibly only some parts, if you think something is not useful/worth including).

This PR requires the following memtest extensions: https://github.com/enjoy-digital/litex/pull/916 and is built on top of it. To see only the additional changes please use the following page:
https://github.com/antmicro/litex/compare/jboc/memtest-config...antmicro:jboc/sdram-dbg
Additionally, the following PR is required: https://github.com/enjoy-digital/litedram/pull/252. It allows to `#include <generated/sdram_phy.h>` in more than one .c file.

**Example**

The `sdram_debug` command provides the following output:
```
Memtest at 0x40000000 (2MiB)...
  Write: 0x40000000-0x40200000 2MiB
   Read: 0x40000000-0x40200000 2MiB
  bus errors:  0/256
  addr errors: 0/8192
  data errors: 7378/524288
Memtest KO
Memory initialization failed

--============= Console ================--

litex> sdram_debug

Error stats:
Running initial memtest to fill memory ...
  Write: 0x40000000-0x40200000 2MiB
   Read: 0x40000000-0x40200000 2MiB
Running read-only memtests ...
Running read-only memtest  10/ 10 ...
        DQ:     0     1     2     3     4     5     6     7     8     9    10    11    12    13    14    15
  beat[ 0]:     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0
  beat[ 1]:  2240     0     0     0     0     0     0     0     0     0     0     0     0     0     0     0
  beat[ 2]:     0     0     0     0    22     0     0     0    62     0     0     0     1     0    19     0
  beat[ 3]:  3514     0     0     0     3     0     0     0     0     0     0     0     0     1     7     0
  beat[ 4]:     0     0     0     0    13     0     0     0     6     0     0     0     0     0    18     0
  beat[ 5]:  7673     0     0     0     9     0     0    40     0     0     0     0     0     0    13     0
  beat[ 6]:     0     0     0     0     3     0     0     0    17     0     0     0     0     0    10     0
  beat[ 7]: 28335    10    10    80    24    30    20  1310     0     0     0     0     0     0    11     0
  beat[ 8]:     0     0     0     0    26     0     0     0    42     0     0     0    12     0    41     0
[...]

PR #2347, opened 2025-10-08
bios: fix sdram_debug error report

ozel:
The error report of `sdram_debug` was hard coded for testing 16-bit DDR RAM, now it's generic as the data collection already was. It reports now one beat line for SDR RAM run at sys_clck and two beat lines when running SDR RAM at sys_clk * 2 (`--sdram-rate 1:2`).

Some hopefully useful context for anyone finding this later via github keyword search:

This bios command was great for understanding that certain colorlight i5 modules come only with half of the SDRAM size accessible (one bank address pin out of two can be unconnected, full story: https://github.com/enjoy-digital/litex/pull/917#issuecomment-3220207378 and https://github.com/wuxx/Colorlight-FPGA-Projects/issues/26#issuecomment-3225063362). 

Since litex uses `ROW_BANK_COL` address mapping, it's not just the upper half of RAM that was missing in my case.

It was also helpful to reduce `MEMTEST_DATA_SIZE`, e.g. `self.add_constant("MEMTEST_DATA_SIZE", 0x40)` in the SoCCore-derived system design. Because with the address mapping that LiteX uses, a bank access error depends highly at which base address the SDRAM is being read/written: large `sdram_debug` error stats per data line may hide the actual issue and can make it seem like general signal integrity/delay errors - which was luckily not happening in my case. `sdram_debug` should eventually get parameters to specify an offset and test range. Or be interruptible by key press like `srdam_bist`. Something for next time...

### DIFF
--- a/litex/soc/software/liblitedram/sdram_dbg.c
+++ b/litex/soc/software/liblitedram/sdram_dbg.c
@@ -28,15 +28,15 @@ void error_stats_update(struct error_stats *stats, struct memory_error err) {
 
 void error_stats_print(struct error_stats *stats) {
 	printf("        DQ:");
-	for (int bit = 0; bit < 16; ++bit) {
+	for (int bit = 0; bit < SDRAM_PHY_DATABITS; ++bit) {
 		printf(" %5d", bit);
 	}
 	printf("\n");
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @ozel!

PR #2348, opened 2025-10-09
cpu/coreblocks: remove unnecessary mem_map constraints

piotro888:
Remove unnecessary `mem_map` constraints from corebloks cpu. I misunderstood this parameter to be overridable from SoC class.
Only `clint` needs to be fixed, and `csr` base is required for LiteX.

Without it, for example `main_ram` constraint requires patch for no-mmu Linux to run.


enjoy-digital:
Thanks @piotro888.

PR #2350, opened 2025-10-15
soc: make more use of the mode parameter of the interface

maass-hamburg:
make more use of the mode parameter of the interface (beneficial for axi)  + fix in SoCBusHandler where mode was applied to the adapted interfaces.

more info in commit msg.


enjoy-digital:
Thanks @maass-hamburg, this looks good.

Issue #2361, opened 2025-11-11
SD1 PS peripheral not supported in Zynq-7000


gkirilow-motec:
We are currently working on a project where we would like to use SD1 as an EMIO but noticed no references to SD1 in the generation scripts. SD0 is handled as below in 

```
        # SDIO0 ------------------------------------------------------------------------------------
        ps7_sdio0_pads = platform.request("ps7_sdio0", loose=True, reserve=self.reserve_pads)
        if ps7_sdio0_pads is not None:
            self.cpu_params.update(
                o_SDIO0_CLK     = ps7_sdio0_pads.clk,
                i_SDIO0_CLK_FB  = ps7_sdio0_pads.clk_fb,
                o_SDIO0_CMD_O   = ps7_sdio0_pads.cmd_o,
                i_SDIO0_CMD_I   = ps7_sdio0_pads.cmd_i,
                o_SDIO0_CMD_T   = ps7_sdio0_pads.cmd_t,
                o_SDIO0_DATA_O  = ps7_sdio0_pads.data_o,
                i_SDIO0_DATA_I  = ps7_sdio0_pads.data_i,
                o_SDIO0_DATA_T  = ps7_sdio0_pads.data_t,
                o_SDIO0_LED     = ps7_sdio0_pads.led,
                o_SDIO0_BUSPOW  = ps7_sdio0_pads.buspow,
                o_SDIO0_BUSVOLT = ps7_sdio0_pads.busvolt,
            )

        # SDIO0_CD ---------------------------------------------------------------------------------
        ps7_sdio0_cd_pads = platform.request("ps7_sdio0_cd", loose=True, reserve=self.reserve_pads)
        if ps7_sdio0_cd_pads is not None:
            self.cpu_params.update(i_SDIO0_CDN = ps7_sdio0_cd_pads.cdn)

        # SDIO0_WP ---------------------------------------------------------------------------------
        ps7_sdio0_wp_pads = platform.request("ps7_sdio0_wp", loose=True, reserve=self.reserve_pads)
        if ps7_sdio0_wp_pads is not None:
            self.cpu_params.update(i_SDIO0_WP = ps7_sdio0_wp_pads.wp)
```

Which we use perfectly fine for SD0 booting.

Is this a known gap that is planned to be supported?

Above snippet is from litex/litex/soc/cores/cpu/zynq7000/core.py

We were attempting to integrate SD1 using:

```
"PCW_SD1_PERIPHERAL_ENABLE"             : "1",
[...]

PR #2365, opened 2025-11-11
Zynq7000 sdio

trabucayre:
The *zynq7000* contains 2 SDIO controlers. These may be used in *MIO mode* (PS side) or in *EMIO mode* (via PL IOs).
- In *MIO mode* no IOs are exposed to the logic, and `card detect` and `write protect` may be enabled/disabled and used via *MIO* or *EMIO*
- in *EMIO mode* a single signals set of IOs are exposed in logic. `card_detect` and `write protect` are not configurables and always present in IOs set
- in both mode: `power control` may be enabled/disabled but must be always connected to PS MIO.

The second commit removes in CTOR the SDIO part unused and not working (in *MIO mode* signals are not accessibles by the logic).

Tested with an `digilent_arty_z7` and a digilent micro SD PMOD.

This solution address issue #2361

### DIFF
--- a/litex/soc/cores/cpu/zynq7000/core.py
+++ b/litex/soc/cores/cpu/zynq7000/core.py
@@ -143,33 +143,6 @@ class Zynq7000(CPU):
         )
         self.specials += AsyncResetSynchronizer(self.cd_ps7, ~ps7_rst_n)
 
-        # SDIO0 ------------------------------------------------------------------------------------
-        ps7_sdio0_pads = platform.request("ps7_sdio0", loose=True, reserve=self.reserve_pads)
-        if ps7_sdio0_pads is not None:
-            self.cpu_params.update(
-                o_SDIO0_CLK     = ps7_sdio0_pads.clk,
-                i_SDIO0_CLK_FB  = ps7_sdio0_pads.clk_fb,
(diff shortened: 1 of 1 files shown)


litex/soc/cores/cpu/zynq7000/core.py
```diff
@@ -531,6 +505,137 @@ def add_uart(self, n, pads_or_mio_group):
                 f"i_UART{n}_RX" : pads_or_mio_group.rx,
             })
 
+    """
+    Connect and Enables SDn controler (may be via PS7 MIO or PL EMIO).
```
gkirilow-motec: Minor: controler -> controller. Repeat for all instances below.

litex/soc/cores/cpu/zynq7000/core.py
```diff
@@ -531,6 +505,137 @@ def add_uart(self, n, pads_or_mio_group):
...
+    Attributes
+    ==========
+    n: int
+        controler ID 0/1
+    pads_or_mio_group: Record or str:
+        When pads_or_mio_group is:
+        - a Record, SDn controler is configured to uses EMIO
```
gkirilow-motec: Minor: to uses -> to use

PR #2375, opened 2025-11-28
build/microsemi: Reworks Plaform and Toolchain to accept devices other than PolarFire

trabucayre:
Currently, *Microsemi*/*Microchip* toolchain can only be used for _PolarFire_ devices.

This PR updates the code to be less dependant:
- the Toolchain is renamed to remove _PolarFire_ aspect
- `finalize` method is added to parse `device` as soon as possible, knowing device family is mandatory for most of the subsequent operations
- `pdc` methods are also reworked: depending to the family, options and keywords may differs
- `part_range`, `speed` extract is also improved by analyzing package
- `tcl` operations are modified
  -  to specify `family`, `package`, `part_range`, `speed` and `voltage` when the project is created
  - `export_prog_job` is also modified, some options are only compatibles with _PolarFire_.

### DIFF
--- a/litex/build/microsemi/libero_soc.py
+++ b/litex/build/microsemi/libero_soc.py
@@ -18,18 +18,37 @@ from litex.build.generic_toolchain import GenericToolchain
 from litex.build.microsemi import common
 
 
-# MicrosemiLiberoSoCPolarfireToolchain -------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #2379, opened 2025-12-04
Libero proasic3 support

trabucayre:
This PR contains 2 commits:
- The first commit adds Libero SoC version detect and some code adaptations for v11.9 release
- The second commit adds another set of modifications to enables support for proASIC3 family (only supported by Libero SoC <= v11.9)

I'm able to build a basic `gateware` without CPU and with `uartbone`. No physical boards to perform tests.


enjoy-digital:
Thanks @trabucayre!

PR #2381, opened 2025-12-09
Import get_sdram_phy_settings from litedram.phy.model

0d41146:
Fixed the following import error that occurred when running `litedram_gen example.yml --sim`: `ImportError: cannot import name 'get_sdram_phy_settings' from 'litex.tools.litex_sim'`

### DIFF
--- a/litex/tools/litex_sim.py
+++ b/litex/tools/litex_sim.py
@@ -29,6 +29,8 @@ from litex.soc.cores.gpio    import GPIOTristate
 from litex.soc.cores.cpu     import CPUS
 from litex.soc.cores.video   import VideoGenericPHY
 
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @0d41146!

0d41146:
Thanks for the feedback. I understand that my previous change introduced an unwanted dependency on `litedram`.

I will follow your suggestion and open a new PR in the `litedram` repository to update `litedram/gen.py` (changing the import to `from litedram.phy.model import get_sdram_phy_settings`).

After that, I will submit a new PR here to revert the changes in `litex_sim.py` to clean up the dependency.

0d41146:
Ah, I noticed there is already a PR for the litedram fix.
So I will proceed with reverting the changes here.

PR #2383, opened 2025-12-15
zynq7000: Fix invalid escape sequence warning

cklarhorst:
This avoids:
`DeprecationWarning: invalid escape sequence \d assert not (io_type != "EMIO" and re.match("MIO \d\d .. \d\d", io_type) is None)`

In normal Python strings, `\d` must be written as `\\d`. Changed it to a raw string for the regex so no escaping is needed.


trabucayre:
Good catch. Applied.
Thanks @cklarhorst !

Issue #2321, opened 2025-09-01
neorv32 fails to build (multiple assignments for "addr_mask_napot" , "addr_mask" )


Ruinland:
Hi Florent,

I just bumped into this error while trying to build neorv32 for orange crab.

It gives out following error : 
```
/home/ruinland/machdyne_lakritz/lol_root/litex/litex/soc/cores/cpu/neorv32/neorv32_cpu_pmp.vhd:90:10:error: multiple assignments for "addr_mask_napot" offsets 0:239
  signal addr_mask_napot, addr_mask : addr_mask_t;
         ^
/home/ruinland/machdyne_lakritz/lol_root/litex/litex/soc/cores/cpu/neorv32/neorv32_cpu_pmp.vhd:90:27:error: multiple assignments for "addr_mask" offsets 0:239
  signal addr_mask_napot, addr_mask : addr_mask_t;


...

Traceback (most recent call last):
  File "/home/ruinland/machdyne_lakritz/lol_root/neorv32_test/../litex-boards/litex_boards/targets/gsd_orangecrab.py", line 231, in <module>
    main()
    ~~~~^^
  File "/home/ruinland/machdyne_lakritz/lol_root/neorv32_test/../litex-boards/litex_boards/targets/gsd_orangecrab.py", line 224, in main
    builder.build(**parser.toolchain_argdict)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ruinland/machdyne_lakritz/lol_root/litex/litex/soc/integration/builder.py", line 373, in build
    self.soc.finalize()
    ~~~~~~~~~~~~~~~~~^^
  File "/home/ruinland/machdyne_lakritz/lol_root/litex/litex/soc/integration/soc.py", line 1505, in finalize
    Module.finalize(self)
    ~~~~~~~~~~~~~~~^^^^^^
  File "/home/ruinland/machdyne_lakritz/lol_root/lol_venv/lib/python3.13/site-packages/migen/fhdl/module.py", line 156, in finalize
    subfragments = self._collect_submodules()
  File "/home/ruinland/machdyne_lakritz/lol_root/lol_venv/lib/python3.13/site-packages/migen/fhdl/module.py", line 149, in _collect_submodules
    r.append((name, submodule.get_fragment()))
                    ~~~~~~~~~~~~~~~~~~~~~~^^
[...]

PR #2387, opened 2025-12-29
Update neorv32 to v1.12.6.2

danielgusvt:
With this I was able to synthesize the SoC with the NEORV32 CPU, fixing https://github.com/enjoy-digital/litex/issues/2321
I have tested `standard` variant on both FPGA and simulator up to litex bios prompt successfully,
but `full` variant will hang and I haven't had the time to debug this yet.


enjoy-digital:
Thanks @danielgusvt for the CPU update. Hoping we'll also manage to get the `full` variant working later.

PR #2391, opened 2026-01-07
add fractional multiplier to s7 MMCMs

Mimil25:
Hello,

This enables the config finder to use the fractional part of the multiplier on xilinx 7 series MMCMs, allowing more frequencies ratios to be produced (e.g. 10 MHz -> 62.5 MHz).

Additionally, the fractional mode of operation can be forbidden, this is useful for a safe instantiation MMCMs with fine phase shift mode, which is incompatible with the fractional mode.

Originally, the iteration over the feedback multiplier (l. 82) was reversed, I am not sure why, if needed I can fix that back.

Thank you


enjoy-digital:
Thanks @Mimil25, this is merged.

PR #2395, opened 2026-01-09
soc/cores/cpu/zynq7000/core.py: fixed one constant in add_ethernet method

trabucayre:
Explicit uses 1/0 instead of expecting for conversion between boolean and int


enjoy-digital:

Issue #2394, opened 2026-01-09
Dynamic baud rate on UART no longer works


0xDRRB:
Since commit 1b5fc97, adding a UART with `add_uart()` while specifying `with_dynamic_baudrate=True` makes the device unable to send data (with or without touching the tuning word).

Everything works fine by switching the parameter to `False` and setting a fixed baud rate with `baudrate=38400`, for example. Dynamic baud rate and tuning word adjustment worked fine before this commit (I checked).

helium729:
@0xDRRB I have reproduced the error and submitted a PR to fix.

0xDRRB:
No. Still no state change on the pin connected to the TX of the serial port.

Would it help if I provided the contents of the build directory, with/without dynamic baudrate, and before/after the e5e6473 commit?

If so, I have created an archive available [here](https://rollmops.ninja/build_dynamicbaudrate_mix.tar.gz) ("OLD"=e5e6473, "true"=`with_dynamic_baudrate=True`).

PR #2406, opened 2026-01-28
Fix dynamic baudrate UART issue by explicitly naming CSRs. 

helium729:
#2394 reported that enabling dynamic baudrate on UART caused transmission failure.

`CSR` objects were instantiated without explicit names, relying on automatic name extraction from variable names. This extraction can be fragile, especially with private attributes (starting with `_`) or in certain contexts (like `EventManager`).
When `LiteXModule` introspection logic changed (as noted in the issue referencing commit `1b5fc97`), this implicit naming likely broke, causing the `tuning_word` CSR (required for dynamic baudrate) or `rxtx` CSR to be mishandled.

I reproduced the `ValueError: Cannot extract CSR name` in a standalone simulation script, confirming the fragility.
The fix involves explicitly passing `name="..."` to `CSR`, `CSRStatus`, and `CSRStorage` in `RS232PHY`, `UART`, and `EventManager`.
This ensures robust CSR generation and fixes the issue.

### DIFF
--- a/litex/soc/cores/uart.py
+++ b/litex/soc/cores/uart.py
@@ -153,7 +153,7 @@ class RS232PHY(LiteXModule):
     def __init__(self, pads, clk_freq, baudrate=115200, with_dynamic_baudrate=False):
         tuning_word = int((baudrate/clk_freq)*2**32)
         if with_dynamic_baudrate:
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @helium729! Let's merge. I'll still probably investigate further since there could be another underlying issue.

PR #2441, opened 2026-04-01
Allow for changing starting position of header

Skinbow:
In extending LiteEth with RDMA with the pull request ["Implementing RoCEv2 passive receiver#198"](https://github.com/enjoy-digital/liteeth/pull/198), there was a need for shifting Header encoding and decoding, as the InfiniBand protocol has variable opcode-dependent headers.

This change simply adds the possibility to shift the place the header is extracted from in a wider Signal.


enjoy-digital:
Thanks @Skinbow, this is merged.

PR #2445, opened 2026-04-03
Add Gowin support for MacOS

AllinLeeYL:
Dear repo maintainer,

I am using Gowin FPGA board with LiteX on MacOS recently. I added a small modifiaction to support Gowin toolchain on MacOS. If you feel this useful, could you merge it?

Best

### DIFF
--- a/litex/build/gowin/gowin.py
+++ b/litex/build/gowin/gowin.py
@@ -9,6 +9,7 @@ import os
 import sys
 import math
 import subprocess
+import platform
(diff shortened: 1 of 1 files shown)


trabucayre:
Applied with few modifications:
- a dict is uses to handle Darwin/Linux (more easy to read)
- `gowin_lib` is obtained via `gw_sh` path (more generic way: same approach for all OS/ specific install path)
Thanks @AllinLeeYL !

PR #2448, opened 2026-04-21
json2dts_zephyr: add uart_handler

maass-hamburg:
add uart handler with support for setting
"rx-fifo-rx-we".


enjoy-digital:
Thanks @maass-hamburg!

PR #2451, opened 2026-04-28
cores: cpu: vexiiriscv: use gcc macro for riscv extensions

maass-hamburg:
gcc already adds defines for each enabled riscv extension via `riscv_*`. Use that for the zicbom extension.


enjoy-digital:
Thanks @maass-hamburg,  I'll let @Dolu1990 look and merge if this seems OK.

Dolu1990:
Looks good to me ^^
Thanks

PR #2455, opened 2026-05-04
json2dts_zephyr: use dma-coherent dt prop

maass-hamburg:
zephyr now supports the dma-coherent
dt prop, therefore set it, so drivers, like the
litesdcard driver can use it.


enjoy-digital:
Thanks @maass-hamburg.

PR #2458, opened 2026-05-12
cores: gpio: add intermediate Signals

maass-hamburg:
If the gpio ports should be used by other modules too we need to have intermediate Signals, otherwise we get a error durring build.

for in self.i is used because self.in would use the word in which is reserved in python.


enjoy-digital:
Thanks @maass-hamburg, that's merged with some changes on top: https://github.com/enjoy-digital/litex/commit/04ebd7421d4a1bdd0f016f8adce81e19954c38b6

PR #2461, opened 2026-05-16
soc/cores: Update io_regions of gowin_emcu

david-sawatzke:
The gowin emcu only exposes very small slices of the address space. 

I've tested this with litescope, any address above 0xa000_1FFC isn't exposed anymore on the AHB master. The APB master also seems to be quite constrained, but I haven't verified it. It also doesn't matter, as it isn't even hooked up (so not accessible in general by user logic in the current code).

This is the document I'm referring to in the comments: https://cdn.gowinsemi.com.cn/IPUG517E.pdf

This also implies that the memory-mapped hyperram in the target file *can't* work: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/targets/sipeed_tang_nano_4k.py#L114

@enjoy-digital Is there any way to use the hyperram core without it needing to be memory mapped in the CPU?

### DIFF
--- a/litex/soc/cores/cpu/gowin_emcu/core.py
+++ b/litex/soc/cores/cpu/gowin_emcu/core.py
@@ -29,8 +29,15 @@ class GowinEMCU(CPU):
     nop                  = "nop"
     io_regions           = {
         # Origin, Length.
-        0x4000_0000 : 0x2000_0000,
-        0xa000_0000 : 0x6000_0000,
+        #
+        # This is part of the APB master, only exposed on the APBTARGEXP2 pins
+        #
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the PR/findings @david-sawatzke! I'll have a closer look. For they Hyperram core, you could wrap the Wishbone with CSRs and use them to do indirect accesses (slower but could allow you to access the Hyperram with the emcu).

enjoy-digital:
This is merged, I just updated the comments a bit to less a personal note: https://github.com/enjoy-digital/litex/commit/c4f606aa2b0c693bafa18818b34a7b1c18f21cb2

david-sawatzke:
Update: I'm pretty sure the emcu is the [`ARM CoreLink SSE-050 Subsystem`](https://developer.arm.com/documentation/100918/0000), with only some of the busses exposed

PR #2467, opened 2026-05-25
cpu: Coreblocks update

awariac:
Changes:
- add PLIC support. 
 Used by default in Linux-capable configs. Interrupts are wired directly to PLIC.
 In simpler variants without PLIC interrupts are wired directly to RISC-V hart local interrupt controller, as in original code (shifted). Both schemes are supported in `irq.h`
- Packaging changes
  Coreblocks is now distributed as standalone package.
  I made changes to [pythondata-cpu-coreblocksx](https://github.com/litex-hub/pythondata-cpu-coreblocks) to simplify packaging process. pythondata package now runs isolated pipx verilog generation from pinned coreblocks commit (without submodule and dependency duplication).
This is the preffered way of running elaboration, because we depend on very fragile packages and python version, yosys version.
For developer use, `coreblocks` package can be installed directly in venv, and `LITEX_COREBLOCKS_RUN_NATIVE` flag specified to run it directly from installed package instead.
Everything is documented in pythondatas' package readme.
- Native Amaranth conversion removal
  I'm very sorry to remove your native Amaranth conversion work, but we depend on some amaranth hacks that are applied in internal verilog generation (for now e.g. `--enable-vivado-hacks` to workaround Vivado synthesis bugs). Due to fragile nature of verilog generation with such a big project, I would strongly prefer generating verilog source with our generation code (preferably isolated), and using it in LiteX.
- Custom core configurations can be injected with `COREBLOCKS_CORECONFIG_FILE` env var (waiting for merge in Coreblocks), documented in pythondata.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -25,7 +25,7 @@ Copyright (c) 2018-2023 Antmicro <Antmicro.com>
 Copyright (c) 2019      Antony Pavlov <antonynpavlov@gmail.com>
 Copyright (c) 2019      Antti Lukats <Antti.Lukats@googlemail.com>
 Copyright (c) 2025-2026 Aoba Fujino <41146f@gmail.com>
-Copyright (c) 2025      Aria Węgrzyn <piotro@piotro.eu>
+Copyright (c) 2024-2026 Aria Węgrzyn <git@ariac.at>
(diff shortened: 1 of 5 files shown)


enjoy-digital:
Thanks @awariac, this is merged. I perfectly understand of for the verilog generation. I just wanted to avoid duplication in LiteX on the Amaranth to Verilog conversion, but if you are handling it directly, it's fine.

PR #2469, opened 2026-05-30
build: resolve symlinks for Gowin toolchain

flafflar:
In Arch Linux, there is a native package for Gowin's toolchain, which symlinks the `gw_sh` binary into `/usr/bin`. This patch resolves that symlink to locate the actual location of the toolchain.


trabucayre:
Applied.
Thanks @flafflar !

PR #2470, opened 2026-06-02
sim/ethernet: handle both GMII and bare frames on TX for tap0

theox33:
### Problem

The `ethernet_tick()` function in `litex/build/sim/core/modules/ethernet.c` accumulates every byte emitted on the `source` signals and writes them verbatim to the TAP interface via `tapcfg_write()`.

This causes an issue depending on the module connected to the PHY:
1. **LiteEthMAC** outputs a complete GMII byte stream on TX (7 bytes preamble `0x55`, 1 byte SFD `0xd5`, the Ethernet frame, and a 4 bytes FCS). Writing this full stream shifts the frame by 8 bytes on the TAP interface. This produces corrupted Ethernet headers that the Linux kernel silently discards.
2. **Etherbone** (when `with_preamble_crc = False` in `LiteEthPHYModel`) outputs bare frames without preamble or FCS, which works perfectly with the current code.

**Observed symptom on tap0 with LiteEthMAC (tcpdump):**
`55:d5:ff:ff:ff:ff > 55:55:55:55:55:55, ethertype Unknown (0xffff), length 72`
The ARP/ICMP payload was present and correct at a +8 byte offset, but the malformed header prevented any host↔SoC communication.

### Fix

This PR introduces a lightweight state machine in the TX path of `ethernet_tick()` to dynamically support both use cases:

1. **For GMII streams (LiteEthMAC):** It detects the preamble and the SFD byte (`0xd5`). It starts accumulating only from the following byte (first byte of dst MAC) and flags the frame to strip the last 4 bytes (FCS) before calling `tapcfg_write()`.
2. **For bare frames (Etherbone):** If the first byte is not a preamble, it assumes a bare frame. It accumulates everything and does not strip any FCS at the end.

The RX path (tap0 → SoC) remains unchanged.

### Testing

Verified in Verilator simulation with `LiteEthPHYModel` on a TAP interface (`tap0`):
- Both `LiteEthMAC` (192.168.1.50) and `Etherbone` (192.168.1.51) can coexist and operate correctly.
- `ping` from host to both IP addresses returns 5/5 replies.
- `tcpdump -i tap0` shows correct ARP request/reply exchanges followed by valid ICMP echo requests/replies, with no `0xffff` ethertype corruption.

### DIFF
--- a/litex/build/sim/core/modules/ethernet/ethernet.c
+++ b/litex/build/sim/core/modules/ethernet/ethernet.c
@@ -101,7 +101,7 @@ static int ethernet_start(void *b)
 {
   base = (struct event_base *) b;
   printf("[ethernet] loaded (%p)\n", base);
-  return RC_OK;
+return RC_OK;
 }
 
 void event_handler(int fd, short event, void *arg)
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks for the report and the PR. The issue is valid, but I integrated a slightly reworked version to avoid ambiguity in auto-detection.

The final implementation only switches to GMII/preamble+FCS handling after seeing the complete Ethernet preamble sequence (`55` x7 + `d5`). Partial matches now fall back to bare-frame handling, so valid bare frames starting with `0x55` are preserved. I also kept compatibility with existing simulator configs and added an optional `tx_mode` override for cases where the mode should be forced explicitly.

This was integrated as commit 59f129ca2 (`sim/ethernet: Handle GMII TX frames on TAP`).

So the fix keeps your intended LiteEthMAC/TAP behavior while avoiding regressions for Etherbone/bare frames. 

Thanks again for identifying and testing the original issue!

PR #2471, opened 2026-06-03
Add strtoull to the picolibc build list

smunaut:
sdram_hw_test_handler makes use of strtoull and building fails without it.


enjoy-digital:
Thanks @smunaut!

PR #2473, opened 2026-06-07
Update gowin emcu to match hardware-connected signals

david-sawatzke:
Quite a few of the signals on the EMCU primitive are connected in hardware and deliberately [disconnected by nextpnr](https://github.com/YosysHQ/nextpnr/blob/32324500c4ae33670df429d2e8e2a83b51b062ab/himbaechel/uarch/gowin/pack.cc#L518) during the build. This includes the flash read data and most debug signals. 

EDIT: at least the flash changes don't really change anything and are more-so for accuracy. Feel free to not close/drop this commit if it doesn't seem useful.


enjoy-digital:
Thanks @david-sawatzke!

PR #2474, opened 2026-06-08
soc/cores/clock/intel_agilex.py: set self_reset_en to TRUE to allows auto reset when lock is lost

trabucayre:
Agilex pll (`tennm_ph2_iopll`) may performs an auto reset when the input signal/lock is lost. The current status is to disable this feature with some drawback for external clock coming from a device is disabled during a reset or something similar.

This commit enables unconditionally this feature.


enjoy-digital:

PR #2481, opened 2026-06-20
vexiiriscv: core: Use computed isa mapping from VexiiRiscv

inochisa:
Use ISA reported by the VexiiRiscv to avoid wrong ISA configuration

Depends SpinalHDL/VexiiRiscv#143


Dolu1990:
Merged on the VexiiRiscv side

inochisa:
Thanks, @Dolu1990 , I have updated the hash.

enjoy-digital:
Thanks @inochisa, this is merged.

PR #2484, opened 2026-06-24
Vexiiriscv: fix gcc flags and IMSIC parameters

inochisa:
Fix unsupported gcc flags for VexiiRiscv, only use at most rv64gc when building BIOS
Also, forward the `--imsic-interrupt` parameter to the VexiiRiscv core, otherwise the build with AIA support will failed.

### DIFF
--- a/litex/soc/cores/cpu/vexiiriscv/core.py
+++ b/litex/soc/cores/cpu/vexiiriscv/core.py
@@ -78,7 +78,7 @@ class VexiiRiscv(CPU):
     # Arch.
     @staticmethod
     def get_arch(ignores = []):
-        arch = f"rv{VexiiRiscv.xlen}i2p0_"
+        arch = f"rv{VexiiRiscv.xlen}i"
 
         # s/u is not accepted by the ISA string
         isas = "mafdcvh"
@@ -97,11 +97,15 @@ class VexiiRiscv(CPU):
 
     # Arch for GCC march.
     def get_gcc_arch():
-        # Workaround: add exception for extensions that are not supported
-        # by the GCC in default build.
-        ex = ["zicbom", "zicntr", "zicsr", "zifencei", "zihpm", "sscofpmf"]
+        # Workaround: Assume only allowed supported single letter extension for BIOS
+        arch = f"rv{VexiiRiscv.xlen}i2p0_"
 
-        return VexiiRiscv.get_arch(ex)
+        isas = "mafdcvh"
+        for isa in isas:
+            if isa in VexiiRiscv.isa_map:
+                arch += isa
+
+        return arch + "_zifencei_zicsr"
 
     # Memory Mapping.
     @property
--- a/litex/soc/cores/cpu/vexiiriscv/core.py
+++ b/litex/soc/cores/cpu/vexiiriscv/core.py
@@ -195,7 +195,7 @@ class VexiiRiscv(CPU):
         add_soc_arg("--l2-self-flush", default=None,        help="VexiiRiscv L2 ways will self flush on from,to,cycles")
         # Vexii: interrupts
         add_soc_arg("--with-aplic",                  action="store_true",   help="Enable APLIC.")
-        add_soc_arg("--imsic-interrupts", dest="imsic_interrupt_number", forward=False, default=0, type=int, help="VexiiRiscv IMSIC interrupts, 0 disables this feature.")
+        add_soc_arg("--imsic-interrupts", dest="imsic_interrupt_number", default=0, type=int, help="VexiiRiscv IMSIC interrupts, 0 disables this feature.")
         # Vexii: peripherals
         add_soc_arg("--vexii-video", dest="video",  action="append", default=[], help="Add the memory coherent video controller")
         add_soc_arg("--vexii-macsg", dest="mac_sg", action="append", default=[], help="Add the memory coherent ethernet mac")
--- a/litex/soc/cores/cpu/vexiiriscv/core.py
+++ b/litex/soc/cores/cpu/vexiiriscv/core.py
@@ -78,7 +78,7 @@ class VexiiRiscv(CPU):
     # Arch.
     @staticmethod
     def get_arch(ignores = []):
-        arch = f"rv{VexiiRiscv.xlen}i"
+        arch = f"rv{VexiiRiscv.xlen}i2p0_"
 
         # s/u is not accepted by the ISA string
         isas = "mafdcvh"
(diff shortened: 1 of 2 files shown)


inochisa:
@enjoy-digital, Currently I am not sure whether it is acceptable to keep the BIOS rv64gc only. (But I guess it should not use something special).

Also, I am not sure the `--imsic-interrupt` has some bug since you do not forward it, if you can mention it, I think I can fix it.

Test this with my branch https://github.com/project-inochi/VexiiRiscv/tree/feat/extension-manager, for a success build, requires SpinalHDL/rvls#10. I will send the PR for VexiiRiscv after the hypervisor support is merged.

enjoy-digital:
Thanks for the feedback. I reworked this in a follow-up commit here:

  enjoy-digital/litex@2f4de7338b6f0b0ada810a0f17b8651591822d26

The BIOS GCC flags are now kept conservative: LiteX derives the BIOS `-march` from the actual Vexii XLEN and only keeps the standard BIOS-relevant extensions: `m/a/f/d/c`.

So RV64 builds still use an RV64 `-march`, for example `rv64i2p0_mafdc`, and RV32 builds still use `rv32i2p0_...`, but newer/privileged Vexii extensions such as AIA/IMSIC/H/S* extensions are no longer passed to the BIOS compiler.

The full Vexii ISA is still kept for `CPU_ISA` / metadata reporting; this only limits what we ask the BIOS toolchain to accept.

I also fixed the IMSIC forwarding explicitly: LiteX’s `--imsic-interrupts` is now forwarded to Vexii SocGen as `--imsic-interrupt-number`, so the generated core and metadata receive the configured IMSIC interrupt count.

I added regression coverage for both points, including the future extension names from your extension-manager branch.

I could not push directly to your fork branch since maintainer writes are not enabled, so I pushed the reworked branch here instead:

https://github.com/enjoy-digital/litex/tree/pr-2484-vexiiriscv-rework

Could you please review this approach and, if it looks good to you, either enable maintainer edits on this PR or cherry-pick `2f4de7338b6f0b0ada810a0f17b8651591822d26` into your branch?

inochisa:
@enjoy-digital I guess the 

> Thanks for the feedback. I reworked this in a follow-up commit here:
> 
> [2f4de73](https://github.com/enjoy-digital/litex/commit/2f4de7338b6f0b0ada810a0f17b8651591822d26)
> 
> The BIOS GCC flags are now kept conservative: LiteX derives the BIOS `-march` from the actual Vexii XLEN and only keeps the standard BIOS-relevant extensions: `m/a/f/d/c`.
> 
> So RV64 builds still use an RV64 `-march`, for example `rv64i2p0_mafdc`, and RV32 builds still use `rv32i2p0_...`, but newer/privileged Vexii extensions such as AIA/IMSIC/H/S* extensions are no longer passed to the BIOS compiler.
> 
> The full Vexii ISA is still kept for `CPU_ISA` / metadata reporting; this only limits what we ask the BIOS toolchain to accept.
> 

Great, I have seen your patch and LGTM.

A small question I have found is, since GCC 11 require user to set `zifencei` and `zicsr` extension. It lead to a broken build when providing gcc flags in the currently litex build with march `rv64ima_zifencei_zicsr`. I hope a update of the generic toolchain to match this as RISC-V always required a latest generic toolchain (for bugfix, code optimization, etc...). Using something like `i2p0` is kind of not good and does not reflect the real ISA supported by the core.

> I also fixed the IMSIC forwarding explicitly: LiteX’s `--imsic-interrupts` is now forwarded to Vexii SocGen as `--imsic-interrupt-number`, so the generated core and metadata receive the configured IMSIC interrupt count.
> 
> I added regression coverage for both points, including the future extension names from your extension-manager branch.
> 

Thanks for this improvement.

> I could not push directly to your fork branch since maintainer writes are not enabled, so I pushed the reworked branch here instead:
> 
> https://github.com/enjoy-digital/litex/tree/pr-2484-vexiiriscv-rework
> 
> Could you please review this approach and, if it looks good to you, either enable maintainer edits on this PR or cherry-pick `2f4de7338b6f0b0ada810a0f17b8651591822d26` into your branch?

I wonder how to enabled maintainer edits on this PR? I haven't see the checkbox mentioned in 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

So at this time, I will cherry-pick your patch.

enjoy-digital:
About `zicsr` / `zifencei`: I agree the full Vexii ISA should be reported accurately, and this is still kept in `CPU_ISA` / metadata.

For the BIOS `-march`, I would prefer to keep it more conservative for now. LiteX still supports several RISC-V toolchain versions and many existing CPU integrations already use the `rv*i2p0_...` form for BIOS builds. The BIOS only needs a small base ISA subset, so passing the full generated Vexii ISA to GCC risks breaking older/default toolchains on extensions that are not actually needed to build the BIOS.

So I think this PR should stay focused on fixing the current Vexii build/IMSIC issue. Updating the generic LiteX RISC-V toolchain baseline and moving all RISC-V CPU flags to explicit  `zicsr` / `zifencei` would be better handled as a separate change, with CI coverage across the existing RISC-V cores.

For maintainer edits: no problem if the checkbox is not available on your fork/organization setup; cherry-picking is fine.

inochisa:
> About `zicsr` / `zifencei`: I agree the full Vexii ISA should be reported accurately, and this is still kept in `CPU_ISA` / metadata.
> 
> For the BIOS `-march`, I would prefer to keep it more conservative for now. LiteX still supports several RISC-V toolchain versions and many existing CPU integrations already use the `rv*i2p0_...` form for BIOS builds. The BIOS only needs a small base ISA subset, so passing the full generated Vexii ISA to GCC risks breaking older/default toolchains on extensions that are not actually needed to build the BIOS.
> 

OK, at least now we have the right ISA string, which is good enough for this PR.

> So I think this PR should stay focused on fixing the current Vexii build/IMSIC issue. Updating the generic LiteX RISC-V toolchain baseline and moving all RISC-V CPU flags to explicit `zicsr` / `zifencei` would be better handled as a separate change, with CI coverage across the existing RISC-V cores.
> 

Agreed, this is why I add nothing about bumping the toolchain. I mentioned this as it is problem related. Thanks for your kindly explanation.

> For maintainer edits: no problem if the checkbox is not available on your fork/organization setup; cherry-picking is fine.

Thanks.

enjoy-digital:
Great, thanks for the feedback. Then I think we can merge!

Issue #2486, opened 2026-06-28
cpu/ibex: crt0 reset entry point offset is incorrect


yusefkarim:
I recently tried to get the Ibex core running on a Tang Nano 9K, but ran into a handful of issues. This was one of the more major ones specific to the Ibex core and, I believe, generic to any FPGA board.

The current Ibex CPU [crt0.s](https://github.com/enjoy-digital/litex/blob/master/litex/soc/cores/cpu/ibex/crt0.S) file has `_start` (the reset entry point) right at `0x00`. However, the actual Ibex reset value is `0x80` [[ref](https://ibex-core.readthedocs.io/en/latest/03_reference/exception_interrupts.html)].

I believe an additional problem is that `.norvc` is missing for the `vector_table` instructions, so compressed instruction may be used and the table will not be aligned as expected. You can see this being done in the Ibex example code here: https://github.com/lowRISC/ibex/blob/master/examples/sw/simple_system/common/crt0.S#L88

enjoy-digital:
Thanks @yusefkarim, #2487 has been merged. Ibex integration has also been updated/cleaned up with https://github.com/enjoy-digital/litex/commit/4b97f01a946f422e68d6ecf138132a170e2a1fc2.

PR #2487, opened 2026-06-28
cpu/ibex: fix crt0 reset entry offset

yusefkarim:
This PR is for https://github.com/enjoy-digital/litex/issues/2486.

Adds `norvc`  to the vector table definition to ensure it is correctly aligned and moves the reset entry point (`_start`) down to the correct address (`0x80`).

I've tested this locally on my Tang Nano 9K FPGA board.


enjoy-digital:
Thanks @yusefkarim, this is merged.

PR #2491, opened 2026-06-30
vexiirsicv: system: Fix gcc warning on -Wint-to-pointer-caast

inochisa:
Use unsigned long is safe, as it is 32 bits in ilp32 and 64 bits in ilp64.


enjoy-digital:
Thanks @inochisa!

PR #2493, opened 2026-07-03
litex/build/openocd.py: Add CycloneIV to get_ir()

Tomserv-512:
Added a CycloneIV entry to the get_ir() if/else tree in openocd.py

Value tested as working on a CycloneIV device with jtagbone.


enjoy-digital:
Thanks @Tomserv-512!

PR #2503, opened 2026-07-07
cpu/coreblocks: fix missing PLIC init in crt0

awariac:
PLIC in coreblocks was not initialized, i missed one place when adding support for it.

Thought for other change - maybe it would be good to add a common `irq_init` common interface C function in `irq.h`? I'm not sure how useful that would be other CPUs.

### DIFF
--- a/litex/soc/cores/cpu/coreblocks/crt0.S
+++ b/litex/soc/cores/cpu/coreblocks/crt0.S
@@ -28,8 +28,15 @@ reset_vector:
 	j 1b
 3:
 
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @awariac, this is merged. I'll think about adding the common `irq_init `and see if it can simplify the code.

enjoy-digital:
@awariac: https://github.com/enjoy-digital/litex/pull/2504 -> would you mind providing feedback?

PR #2504, opened 2026-07-08
software: Add common IRQ initialization hook

enjoy-digital:
This adds a common `irq_init()` libbase entry point for CPU startup code that needs to initialize interrupt-controller state.

Changes:
- Declare `irq_init()` in the shared `irq.h` wrapper.
- Move the RISC-V PLIC/APLIC startup setup behind `irq_init()`.
- Keep global interrupt enable controlled by `irq_setie()`.
- Drop CVA5's duplicate IRQ table/attach/detach definitions so it uses the common libbase table.

Validation:
- `git diff --check`
- `riscv64-unknown-elf-gcc` compiled `isr.c` for the touched PLIC paths.
- `riscv64-unknown-elf-gcc` compiled `isr.c` for VexiiRiscv APLIC and CVA5 non-PLIC.
- `riscv64-unknown-elf-gcc` assembled the touched `crt0.S` files.

### DIFF
--- a/litex/soc/cores/cpu/coreblocks/crt0.S
+++ b/litex/soc/cores/cpu/coreblocks/crt0.S
@@ -29,14 +29,12 @@ reset_vector:
 3:
 
 #ifdef __riscv_plic__
-    call plic_init
-    li t0, 0x800
-    csrw mie,t0
+	call irq_init // initialize interrupt controller
 #endif
 
 	call litex_startup_init
 
-    call main
+	call main
 
 1:	j 1b
 
--- a/litex/soc/cores/cpu/coreblocks/crt0.S
+++ b/litex/soc/cores/cpu/coreblocks/crt0.S
@@ -28,10 +28,6 @@ reset_vector:
 	j 1b
 3:
 
-#ifdef __riscv_plic__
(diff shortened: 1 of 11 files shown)


litex/soc/cores/cpu/coreblocks/crt0.S
```diff
@@ -29,14 +29,12 @@ reset_vector:
 3:
 
 #ifdef __riscv_plic__
-    call plic_init
-    li t0, 0x800
-    csrw mie,t0
+	call irq_init // initialize interrupt controller
 #endif
```
awariac: `#ifdef` can be dropped here, it's now at C level

awariac (approved):
Looks good; That's what I thought about, thanks!

litex/soc/cores/cpu/coreblocks/crt0.S
enjoy-digital: Thanks @awariac, this has been simplified even further with: https://github.com/enjoy-digital/litex/pull/2504/commits/186a8be9fd306183f00d8087628cfd54e8ad8f5b.

PR #2511, opened 2026-07-09
setup: preserve leading zeroes in pinned SHAs

enjoy-digital:
## Summary
- keep existing integer SHA pins compatible with abbreviated Git hashes, such as the Microwatt pythondata pin
- make newly frozen full SHAs strings so leading zeroes are preserved exactly
- use the same SHA formatter for checkout and init/update error messages
- add regression tests for string full SHAs with leading zeroes and integer abbreviated SHAs

This fixes the intermittent CI failure where test-generated frozen Git commits sometimes start with 0, without breaking existing abbreviated integer pins.

## Tests
- pytest -q test/tools/test_litex_setup.py
- python3 -m py_compile litex_setup.py test/tools/test_litex_setup.py
- git diff --check

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -217,13 +217,18 @@ def print_indented(output, indent="    ", max_lines=None):
         lines = lines[:max_lines] + [f"... ({remaining} more line(s))"]
     print("\n".join(indent + line for line in lines))
 
+def git_format_sha1(sha1):
+    if isinstance(sha1, int):
+        return f"{sha1:040x}"
+    return str(sha1)
+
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Follow-up pushed as 32ffc9c42. The first version overcorrected by padding all integer SHAs to 40 hex digits, which breaks existing abbreviated integer pins such as pythondata-cpu-microwatt's c69953aff92.

The updated behavior is:
- integer sha1 values keep the previous abbreviated formatting;
- newly frozen full SHAs are emitted as strings, so leading zeroes are preserved exactly;
- tests cover both the leading-zero full SHA string case and the abbreviated integer case.

PR #2522, opened 2026-07-13
SoC: add HyperRAM framebuffer and AXI support with extended test coverage

CauaBF1:
## Summary

  This PR extends LiteX framebuffer and SoC integration, adds dedicated tests for
  existing command-line tools, covers the newly introduced functionality, and
  includes a small set of behavior-preserving cleanups identified with SonarQube
  Cloud.

  The main additions are:

  - Wishbone DMA support for video framebuffers.
  - HyperRAM-backed framebuffer support.
  - Explicit framebuffer port and memory selection.
  - A new external AXI/AXI-Lite master integration API.
  - Explicit CRG clock domain names.
  - Dedicated tests for `litex_periph_gen` and `litex_json2renode`.
  - Unit tests for the new framebuffer, HyperRAM, and AXI behavior.
  - Small readability and maintainability improvements reported by SonarQube.

  ## Motivation

  `VideoFrameBuffer` previously assumed that its DMA port was provided by
  LiteDRAM. This prevented the framebuffer from directly using memories exposed
  through Wishbone, including HyperRAM integrations.

  The SoC framebuffer helper also required SDRAM and always reserved its memory
  inside `main_ram`. This made it difficult to use the framebuffer in systems
  where HyperRAM is the primary external memory or where the caller already has a
  suitable memory port.

  In addition, `litex_periph_gen` and `litex_json2renode` had important existing
  behavior without dedicated unit test modules.

  ## SoC and video changes

  ### Wishbone framebuffer DMA

  `VideoFrameBuffer` now selects its DMA reader according to the provided port:

  - `WishboneDMAReader` is used for a Wishbone interface.
  - `LiteDRAMDMAReader` remains the default for native LiteDRAM ports.

  The existing LiteDRAM path and its buffered FIFO behavior are preserved.

  ### HyperRAM framebuffer support

  The framebuffer integration can now:

  - Accept an explicit DMA port.
  - Accept an explicit memory region name.
  - Prefer `main_ram` when available.
  - Fall back to HyperRAM when SDRAM and `main_ram` are unavailable.
  - Reserve the framebuffer region inside the selected memory.
  - Report an error when an explicitly selected memory region does not exist.
  - Report an error when neither SDRAM nor an explicit/HyperRAM port is
    available.

  This allows systems with Wishbone-connected HyperRAM to use the regular LiteX
  video framebuffer path.

  ### External AXI master integration

  A new `SoC.add_axi_master()` helper creates and exposes an AXI-Lite or full AXI
  interface.

[...]

PR #2524, opened 2026-07-15
test: add LiteX JSON-to-Renode coverage

CauaBF1:
### `litex_json2renode`

  A new `test/tools/test_litex_json2renode.py` module verifies:

  - Named normal and shadow system-bus registrations.
  - Preservation of original memory metadata after alignment.
  - Automatic memory-region alignment.
  - VexRiscv variant and time-provider generation for multiple cores.
  - Framebuffer association with its containing memory and the correct offset.

Following up on PR #2522

### DIFF
--- a/test/tools/test_litex_json2renode.py
+++ b/test/tools/test_litex_json2renode.py
@@ -1,0 +1,102 @@
+#
+# This file is part of LiteX.
+#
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @CauaBF1, that's merged.

PR #2526, opened 2026-07-15
build: explicitly name CRG clock domains

CauaBF1:
Explicitly name the system and power-on-reset clock domains to avoid implicit clock domain naming issues.

Following up PR #2522


enjoy-digital:
Thanks @CauaBF1, that's merged.

PR #2532, opened 2026-07-17
build/openfpgaloader: Add --device option

BZab:
Since [2 months](https://github.com/trabucayre/openFPGALoader/commit/9e548983f99f988778abfcc63b04ccfb8c2594e2) `--ftdi-serial` is marked as deprecated. Furthermore, the openFPGALoader yields an error when there's an attempt to access the device with given iSerial value, while other connected USB devices may have iSerial set to `none`.

### DIFF
--- a/litex/build/openfpgaloader.py
+++ b/litex/build/openfpgaloader.py
@@ -12,7 +12,7 @@ from litex.build.generic_programmer import GenericProgrammer
 class OpenFPGALoader(GenericProgrammer):
     needs_bitreverse = False
 
-    def __init__(self, board="", cable="", freq=0, fpga_part="", index_chain=None, ftdi_serial=None):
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @BZab. The `--device` support is useful for selecting serial-backed devices through `/dev/ttyUSBx`.

I recreated this as #2538 because maintainer edits are disabled on this branch and the new constructor argument needed to move after the existing arguments to preserve positional API compatibility. The replacement preserves your original implementation and authorship in commit 04b37ddd, with the compatibility tests and adjustment in follow-up commit 5c42d5ca.

The replacement PR also clarifies that `--device` selects a serial device; USB iSerial selection remains a separate openFPGALoader option.

Closing this one in favor of #2538.

PR #2534, opened 2026-07-20
setup: add -y/--yes for repository updates.

trabucayre:
## Summary

When repositories contain tracked local changes, `litex_setup.py --update` asks for confirmation before updating each one. Add the conventional `-y` / `--yes` option to accept these confirmations for the complete update operation:

```
./litex_setup.py --update --yes
```

Interactive confirmation remains the default. The option only skips the prompts; it does not merge, rebase, overwrite or discard local work, and Git failures are still reported normally.

## Validation

Added a dirty-repository regression confirming that `--yes` skips `input()` while preserving the local working-tree modification.

```
pytest -q test/tools
124 passed
```

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -287,7 +287,7 @@ def git_status_has_tracked_changes(status):
             return True
     return False
 
-def git_confirm_update_with_local_changes(name, repo_path):
+def git_confirm_update_with_local_changes(name, repo_path, always_update_repo):
     status = git_status(repo_path, short=True)
     if not git_status_has_tracked_changes(status) or not sys.stdin.isatty():
         return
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Added follow-up commit 62c47eb1 with the agreed naming and coverage:

- expose `-y` / `--yes` instead of `--always-update-repo`;
- use `assume_yes` internally while preserving the interactive default; and
- verify that automatic confirmation skips the prompt without discarding local changes.

Validation: `124` tool tests pass.

enjoy-digital:
Thanks @trabucayre, that's merged (with the initial -y/--yes you suggested, sorry for asking to change it :)).

PR #2535, opened 2026-07-20
Fix broken inlining in hierarchical verilog

hansfbaier:
I had the issue that a USB design worked as a flat build, but did not as a hierarchical build.
After this fix also the hierarchical USB design works. 
A minimal example to reproduce the bug has been added to the unit tests.

PR #2536, opened 2026-07-20
RfC full hierarchy generation

hansfbaier:
This PR currently is a request for comment.
Note: As this PR depends on https://github.com/enjoy-digital/litex/pull/2535
its commits are still included.
The current hierarchical verilog generation still flattens a child
module if it drives the internals of a parent.
With this patch there is the option of preserving the full hierarchy.
I am not sure about the naming yet.
prefer_ports is a bit too much focused on implementation details.
maybe full_hierarchy would be a better name.
Possible other choices would be:
- keep_hierarchy
- preserve_hierarchy
- full_hierarchy
- no_flatten
- deep_hierarchy
- strict_hierarchy

I verified this code on a USB controller design with full hierarchy.

### DIFF
--- a/litex/gen/fhdl/verilog.py
+++ b/litex/gen/fhdl/verilog.py
@@ -1140,10 +1140,20 @@ def _convert_hierarchical(f, ios, name, platform, special_overrides, attr_transl
 
     _collect_raw_statements(ctx.root)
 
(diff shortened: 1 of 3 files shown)


hansfbaier:
I think --keep-hierarchy reads the nicest. I used this for now.
--no_flatten sould be a nice alternative though.
What do you think?

PR #2537, opened 2026-07-21
liblitedram: Handle NEXUSDDRPHY defines

gatecat:
Adds the defines necessary for https://github.com/enjoy-digital/litedram/pull/396


enjoy-digital:
Thanks @gatecat!

PR #2539, opened 2026-07-22
Add trailing space to synth_opts to fix abc9 and flow3 flags

rblilja:
_synth_opts was "-dsp" with no trailing space making YosysWrapper append the abc9 or flow3 flag directly onto it. Generating an invalid "synth_ice40 -dsp-abc9" command. This made --yosys-abc9 and --yosys-flow3 broken on the iceStorm/iCE40 flow.

The fix is matching the other Lattice toolchains (e.g. oxide.py uses "-flatten ") by adding a trailing space making options concatenate correctly as e.g. "-dsp -abc9".


enjoy-digital:
Thanks @rblilja!

PR #2540, opened 2026-07-24
setup: support uv virtualenv installs

FlyGoat:
## Summary

- Detect uv-created virtual environments from pyvenv.cfg.
- Use `uv pip install --python sys.executable` inside those environments.
- Keep the existing `python -m pip install` path for regular virtual environments and system Python installs.

## Motivation

uv-created virtual environments can provide a uv-backed `pip` executable without installing the `pip` Python module. LiteX Setup previously always used `sys.executable -m pip install`, which fails in that environment with `No module named pip` even though package installation is available through `uv pip`.

## Validation

- `.venv/bin/python -m py_compile litex_setup.py litex/litex_setup.py`
- `.venv/bin/python litex_setup.py --install --config=minimal --dev`

### DIFF
--- a/litex_setup.py
+++ b/litex_setup.py
@@ -475,8 +475,37 @@ def litex_setup_update_repos(config="standard", tag=None, assume_yes=False):
 
 # Git repositories install -------------------------------------------------------------------------
 
+def pip_install_in_virtualenv():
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @FlyGoat!

PR #2541, opened 2026-07-25
Gowin jtag support

trabucayre:
The JTAG primitive is not documented by Gowin, but this is used for goConfig.
[fpgacapZero](https://github.com/lcapossio/fpgacapZero/) also provides a way to access JTAG in GW1N/GW2A.

There is no official primitive but with a simple verilog file (called `gw_jtag.v`) the Gowin toolchain is able to integrates this feature.

This PR adds native support to jtag in LiteX with 3 commits:
- the first commit integrates the logic (mainly based on Lattice ECP5)
- the second commit updates openocd.py to return USER1/USER2 ir code according to chain ID
- the last commit integrates and enables jtag support in `build/gowin` only when the toolchain is fixed to `gowin`.

Note:
- `reset` signal in `GOWINJtag` class is fixed to '0'. With the uses of an Logic Analyzer the `test_logic_reset_o` is always set to '1'
- the `gw_jtag.v` comes from Gowin toolchain. This file is added to the build script at finalize time. If `gw_sh` is not in `PATH` this step is skipped.

### DIFF
--- a/litex/build/gowin/gowin.py
+++ b/litex/build/gowin/gowin.py
@@ -48,6 +48,14 @@ def _gowin_uses_windows_paths():
 
     return os.path.basename(os.path.realpath(gw_sh_path)).lower().endswith(".exe")
 
+def _gowin_ide_path():
+    _, gw_sh_path = _find_gowin_shell()
+    if gw_sh_path is None:
+        return None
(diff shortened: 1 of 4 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #2548, opened 2026-07-28
Rework libxil (zynq7000 & zynqMP)

trabucayre:
Most of the logic required to integrates the _libxil_ used by **zynq** devices is performed in targets scripts:
- `main` method extends software package/library with this lib
- `BaseSoC.finalize`:
  -  clone `embeddedsw` repository
  - copy headers from the repository
  - creates required board specifics headers.

This imply around 50 lines per boards with a lot of dupplication:
- The `main` part is strictly the same for all zynq based boards.
- The `finalize` part only differs by `xparameters.h` content (and its content is more limited to 3 lines: UART controler id and memory area).

So instead of having these big code dupplicate for all boards:
- the libxil directory is extended to handle clone/copy/header fill
- `CPU` class is extended to adds two dummy helpers to adds and prepare specifics software packages

As the result boards have, only, to provides specifics `#define`, rest of the logic is automatically performed at software build time.

A companion PR will be added for litex-boards

### DIFF
--- a/litex/soc/cores/cpu/__init__.py
+++ b/litex/soc/cores/cpu/__init__.py
@@ -45,6 +45,18 @@ class CPU(LiteXModule):
     def bios_map(self, addr, cached):
         return addr
 
+    """
+    Add CPU-specific software packages/libraries to Builder.
+    """
+    def add_software_packages(self, builder):
+        pass
+
+    """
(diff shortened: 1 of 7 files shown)


enjoy-digital:
Awesome, thanks @trabucayre!

PR #2550, opened 2026-08-11
litex/gen: fix hierarchical Verilog wire classification for memory port slices

hansfbaier:
# litex/gen: fix hierarchical Verilog wire classification for memory port slices

## Problem

`--hierarchical-verilog` generation produces Verilog that Vivado rejects at
synthesis for SoCs whose caches use byte-granular write enables:

```
ERROR: [Synth 8-9315] concurrent assignment to a non-net
       'main_basesoc_data_port_dat_r' is not permitted [...]
```

The L2 cache data memory (128-bit, `we_granularity=8`) is split by
`FullMemoryWE` into 16 byte-grains. Each grain's read-data port drives a
slice of the aggregate read-data signal, e.g.:

```verilog
assign main_basesoc_data_port_dat_r[7:0]  = data_mem_grain0_dat0;
assign main_basesoc_data_port_dat_r[15:8] = data_mem_grain1_dat0;
...
```

The flat converter declares the aggregate as `wire` and everything works.
The hierarchical converter declares it as

```verilog
reg [127:0] main_basesoc_data_port_dat_r = 128'd0;
```

which is illegal: a `reg` driven only by continuous assigns (16 drivers).

## Root cause

Commit a7a9d1c7d ("fix hierarchical memory ownership") deliberately drops
`_MemoryPort` specials when building per-module fragments, so
`list_special_ios()` no longer sees memory port endpoints. That commit
covered read-data signals that become module ports (handled via
`force_wires`), but not the internal case where the aggregate read-data
signal is assembled from grain outputs inside the same module.

`_collect_memory_port_signals()` — the fallback used to classify
memory-driven signals — only accepted plain `Signal` port endpoints and
ignored `_Slice` targets, so the aggregate was never added to the wire set
and fell back to `reg`.

## Fix

Unwrap `_Slice` port endpoints to their base signal in
`_collect_memory_port_signals()`, so memory-driven read-data aggregates
are classified as wires. This is behavior-preserving for the flat path
(which already classified them as wires through the separate
`_MemoryPort` specials) and restores the same classification in the
hierarchical path.

## Test

New regression test `test_hierarchical_grain_memory_port_dat_r_is_wire`:
a 128-bit memory with byte-granular WE is split with `FullMemoryWE`, then
converted hierarchically. Fails before the fix (`reg [127:0]` present),
passes after (`wire [127:0]`).

Full verilog test suites pass: `test_verilog.py`,
`test_verilog_memory.py`, `test_hierarchical_verilog.py` (20 tests).

## Hardware verification

`digilent_arty_s7` (Spartan-7) with `--hierarchical-verilog`:

- 138 modules generated (up to depth 4, e.g. `bus → interconnect → arbiter → rr`)
  with the default automatic partitioning; also verified on the
  `full-hierarchy-refactor` branch with `--no-flatten` (full/deep hierarchy,
  248 modules): the same `dat_r` error is fixed there by this change
- Vivado 2024.1: synthesis 0 errors, 0 critical warnings, bitstream written
- Loaded to board: BIOS boots, CRC passes, VexRiscv @ 100 MHz responds on UART

### DIFF
--- a/litex/gen/fhdl/verilog.py
+++ b/litex/gen/fhdl/verilog.py
@@ -871,12 +871,18 @@ def _collect_memory_port_signals(specials, clock_domains=None):
                 sig = getattr(port, attr, None)
                 if isinstance(sig, ClockSignal) and (clock_domains is not None):
                     sig = clock_domains[sig.cd].clk
+                if isinstance(sig, _Slice):
+                    sig = sig.value
                 if isinstance(sig, Signal):
                     signals.add(sig)
-            # Memory read data is always driven by the memory.
+            # Memory read data is always driven by the memory.  The port may
+            # drive a slice of a wider aggregate signal (e.g. after a
+            # byte-granular FullMemoryWE split): classify the aggregate itself.
             sig = getattr(port, "dat_r", None)
             if isinstance(sig, ClockSignal) and (clock_domains is not None):
                 sig = clock_domains[sig.cd].clk
+            if isinstance(sig, _Slice):
+                sig = sig.value
             if isinstance(sig, Signal):
                 targets.add(sig)
                 wires.add(sig)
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @hansfbaier!

PR #2552, opened 2026-08-16
tools/json2dts: emit a node for every I2C bus, not just i2c0

rvodden:
generate_dts() only emitted a "litex,i2c" node when a CSR base named exactly "i2c0" was present. A SoC that instantiates more than one bit-banged I2C master -- for example one per physical bus on the board -- therefore lost every bus but the first from the device tree, silently: no warning, and no way to reach those buses from Linux.

Iterate over every i2c<N> CSR base instead, using the CSR name as the node label so board device nodes and overlays can reference &i2cN.

Output is unchanged for a SoC with a single i2c0 

Found on an ALINX AX7035B, which has an LM75 on one bus and a 24LC04 EEPROM on another.us.

### DIFF
--- a/litex/tools/litex_json2dts_linux.py
+++ b/litex/tools/litex_json2dts_linux.py
@@ -8,6 +8,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import os
+import re
 import sys
 import json
 import argparse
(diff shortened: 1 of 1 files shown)


enjoy-digital:
Thanks @rvodden!

PR #2553, opened 2026-08-17
Vhd2v converter fix missing params

trabucayre:
When `support_mixed_language` is True at platform level and conversion is not forced by user, the `self._normalize_instance_ports` is used to extract ports for `_instance`.
But parameters (`p_`) are skipped/ignored.
As a result, the instance is written with all parameters omitted and the synthesis tool uses default instead of valid value.
This situation imply some build failure hard to understand because the `p_` are correct at LiteXModule level but not seen as synthesis level.

This PR modify `extract_generics` to introduces an additional parameter called `target`. Values may be:
- `GHDL` to built a list of parameters as required by `ghdl`
- any others values to simply build a new dict with only `p_` entries in original dict

the `do_finalyze` method is also updated to update ip_params with generic since `normalize_instance_ports` skip these entries

### DIFF
--- a/litex/build/converter_common.py
+++ b/litex/build/converter_common.py
@@ -134,5 +134,7 @@ def normalize_instance_ports(params, *, top_entity, target="converter", generic_
     return ip_params
 
 
-def extract_prefixed_generics(params, *, prefix="p_"):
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @trabucayre!

PR #2559, opened 2026-08-27
litex_setup: Install packaging>=24.2 alongside setuptools

rvodden:
CI is failing on linux on litex, and probably a few other places, because setuptools has been upgraded here, but packaging has not. This brings packaging in line with setuptool so that the bulid can complete.

https://github.com/litex-hub/linux-on-litex-vexriscv/actions/runs/33004623903/job/98294896871


enjoy-digital:
Thanks @rvodden, that's merged.

Issue #2000, opened 2024-06-23
AXIDownConverter behaviour on sub-word accesses


FlyGoat:
Hi,

I discovered a problem on our AXIDownConverter.

When AXI Master initiate subword access, our AXIDownConverter would generate illegal request to downstream.

See following examples:
For a read request from a bus width of 256 and downstream width of 32, if master made a read to `ARADDR = 0x0` with `ARSIZE = 0b010`, our AXIDownConverter would convert it to 8 incr burst read to `ARDDR 0x0 ~ 0x20`, with `ARSIZE = 0b010` each, which is undesired.

If master made write to `AWADDR = 0x0` , `ARWIZE = 0b010` and `WSTRB= 0hffffffff`, it would be converted to 8 incr burst write to `AWDDR 0x0 ~ 0x20`, with first one have `WSTRB = 0hffffffff` and rest have `WSTRB = 0`, this is prohibited under AXI spec.

This is causing problems on using wider system bus, for similar problem on Wishbone system it is fixed by #1999, but for AXI system as AXI Lite bus does not have SIZE or SEL for selecting read masks, so those undesired bursts will be forward to CSR bus and draining UART FIFO.

Thanks

FlyGoat:
I tried to fix that problem on my own but failed. My migen and bus protocol skill is not sufficient for writing a fully functional converter.

My workaround for now is to use a third party verilog AXI IP.

https://github.com/alexforencich/verilog-axi

I tried to run litex’s bus components against designware’s AMBA VIP suite and there were too many failed cases popping up that I don’t know where to start with.

Given the challenge of implementing functional bus components I think it would be better if we could allow users to select between native implementation and third party implementation.

I wonder if it’s acceptable for litex upstream.

enjoy-digital:
Hi @FlyGoat,

first, thanks for the different fixes/PR! I'll try to look at your issue and I'm not against your suggestion to allow third party implementation of interconnect/components: In fact that's what has been used initially for AXI and could allow using/comparing our implementation against others.

We already have the wrappers around https://github.com/alexforencich/verilog-axi with https://github.com/enjoy-digital/litex_verilog_axi_test and in LiteX, or AXI, we could allow using native or verilog_axi implementations. I'll try to prepare a PoC for this in the next days.

FlyGoat:
Hi @enjoy-digital,

Any update?

Thanks

enjoy-digital:
[...]

PR #2560, opened 2026-09-01
soc/interconnect/axi: AXIDownConverter — hold narrow single-beat commands until the width converter drained

john-kearney:
## Summary

`AXIDownConverter` accepts a narrow single-beat AW/AR (the `is_aw_narrow`/`is_ar_narrow`
fast path added for sub-word accesses, issue #2000) **without checking that the previous
full-width transaction has fully drained through the W/R `StrideConverter`**. Because the
narrow path *detaches* the converter (`w_converter.sink.valid` is de-validated,
`w_converter.source.ready` is forced to 0, and the W/R mux is switched to the lane path),
accepting a narrow command while a wide beat is held mid-conversion corrupts the stream:

* **Write side** — the wide W beat currently held by the converter is swallowed: its
  remaining sub-beats are never emitted (silent write-data loss; the master is acked via
  `axi_from.w.ready`), one write's data is delivered under the other write's address
  (the pending wide beat is routed through the lane slice selected by the *narrow* AW's
  `cap_aw_lane`), and an extra W beat arrives after a `len=0` burst, desyncing
  beat-counting bridges (AXI2AXILite, Wishbone adapters).
* **Read side** — the sub-beat that completes the wide read is captured into `narrow_r_*`
  and presented to the master as the *narrow* read's response: the merged wide word is
  destroyed (wrong read data, zero-fill in its place) and the narrow response is never
  delivered (`axi_to.r.ready` stuck low — permanent read-channel deadlock).

The trigger is protocol-legal and routine: any sub-width single-beat access (e.g. a CPU
byte/word store) issued while a full-width access' W/R stream is still in flight on a
pipelined or slave-stalled fabric. The converter is auto-inserted by
`SoCBusHandler.add_adapter` on every master/slave whose data width differs from the bus
width, so any wide-bus SoC with a narrow region (or a wide DMA port on a narrower bus) is
exposed.

## Fix

Serialize the narrow path against the converter, mirroring what `AXIUpConverter` achieves
structurally with its per-beat lane FIFOs (no preemptible shared conversion state): reuse
the existing `_AXIRequestCounter` idiom (as `AXIArbiter` does for grant locking) to track
outstanding commands:

```python
self.aw_counter = _AXIRequestCounter(
    request  = axi_from.aw.valid & axi_from.aw.ready,
    response = axi_from.b.valid  & axi_from.b.ready,
)
aw_drained = self.aw_counter.empty
```

and gate the narrow AW/AR branch on `aw_drained`/`ar_drained` (read side: request = AR
handshake, response = final wide R beat). A B response is only produced after the slave
accepted all W beats of a write, and the last wide R beat is only produced after the
converter merged all narrow sub-beats of a read — so `empty` is an exact "converter
drained" indication:

* `axi_to.aw/ar.valid` is held low (the command is not presented to the slave) and
  `axi_from.aw/ar.ready` stays low until drained;
* the handshake condition uses the (now gated) master-side ready, so a slave-side
  `aw/ar.ready` asserted while we are holding the command can no longer latch
  `narrow_*_active` spuriously.

Full-width fast-path and FIXED slow-path commands are unaffected: they share the
converter consistently and never detach it, so no additional serialization is needed
there. The counter saturates at 256 outstanding commands (same default as the arbiter's
request counters).

## Tests

Two regression tests in `test/interconnect/test_axi.py` (TestAXI), both failing on the
unfixed tree and passing with this change:

* `test_axi_down_converter_narrow_interleave_write` — wide write stalled mid-conversion
  (0 of 2 sub-beats emitted) on the narrow bus; narrow write presented while held.
  Asserts the narrow AW is not accepted until the wide write fully drained, and that the
  narrow bus sees exactly `[W1_sub0, W1_sub1, W2_lane]` with correct strobes/last and two
  B responses.
* `test_axi_down_converter_narrow_interleave_read` — wide read mid-merge (first sub-beat
  delivered, second stalled); narrow read presented while held. Asserts the narrow AR is
  not accepted before the wide read's R beat is delivered, and that the master receives
  the correctly merged wide word followed by the lane-placed narrow word (no deadlock).

Full results with this change applied:

* `test/interconnect/test_axi.py`, `test_axi_lite.py`, `test_axi_stream.py`: 74/74 ok.
* entire `test/interconnect/` suite: 231/231 ok.
* `test/soc/`: 205 tests, 5 pre-existing environment errors (missing optional `litespi`,
  `litedram`, picolibc data module; identical on the unmodified tree).

## Notes for reviewers

* No new primitives are introduced — the change reuses `_AXIRequestCounter` and only
  gates existing handshakes, so the fast path keeps its zero-latency behaviour whenever
  the converter is idle (the common case for non-interleaved traffic).
* The existing narrow/slow-path serialization via `narrow_*_active` + B is unchanged;
  this change extends the same serialization to "any previous command still draining".

---

*Reproduction scripts (migen sim, no external dependencies) used to demonstrate the
failure before the fix are included alongside this PR body: `poc_axi_downconv_write.py`
and `poc_axi_downconv_read2.py`, with red (unfixed) and green (fixed) logs.*

### DIFF
--- a/litex/soc/interconnect/axi/axi_full.py
+++ b/litex/soc/interconnect/axi/axi_full.py
@@ -543,6 +543,19 @@ class AXIDownConverter(LiteXModule):
         is_aw_slow = ((axi_from.aw.burst == BURST_FIXED) & (axi_from.aw.len != 0)) | \
                      (axi_from.aw.len > max_fast_len)
 
+        # Outstanding-write counter: the narrow single-beat path detaches the W StrideConverter
+        # (its sink is de-validated and the W mux is switched to the lane path), so a narrow AW
+        # may only be accepted once every previously accepted write has fully drained: a B
+        # response is only generated after the converter emitted (and the slave accepted) all of
+        # the write's W beats. Accepting it earlier would swallow the wide beat currently held
+        # mid-conversion: silent write-data loss, one write's data delivered under another
+        # write's address and an extra W beat after a len=0 burst for beat-counting slaves.
+        self.aw_counter = _AXIRequestCounter(
+            request  = axi_from.aw.valid & axi_from.aw.ready,
+            response = axi_from.b.valid  & axi_from.b.ready,
+        )
+        aw_drained = self.aw_counter.empty
+
         aw_fsm = FSM(reset_state="IDLE")
         self.aw_fsm = aw_fsm
 
@@ -550,8 +563,10 @@ class AXIDownConverter(LiteXModule):
             If(narrow_aw_active,
                 axi_from.aw.ready.eq(0),
             ).Elif(is_aw_narrow,
-                aw_narrow_pending.eq(axi_from.aw.valid),
-                axi_to.aw.valid.eq(axi_from.aw.valid),
+                # Hold the narrow AW (and do not present it to the slave) until the W stream of
+                # all previously accepted writes has fully drained through the converter.
+                aw_narrow_pending.eq(axi_from.aw.valid & aw_drained),
+                axi_to.aw.valid.eq(axi_from.aw.valid & aw_drained),
                 axi_to.aw.addr.eq(axi_from.aw.addr),
                 axi_to.aw.len.eq(0),
                 axi_to.aw.size.eq(axi_from.aw.size),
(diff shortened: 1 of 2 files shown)


enjoy-digital:
Thanks @john-kearney for the detailed analysis, reproductions, and regression tests. The diagnosis was correct: the narrow single-beat path can take over the shared W/R conversion state while an earlier full-width transaction is still in flight. This PR was very useful for identifying and understanding the issue.

I opened https://github.com/enjoy-digital/litex/pull/2561 as a minimal replacement based on the same serialization principle. It waits for earlier B/final-R responses before entering the narrow path, while also backpressuring all AW/AR paths when `_AXIRequestCounter` is full. The additional capacity guard ensures every accepted command remains accounted for; otherwise commands accepted after counter saturation could allow the narrow path to be released too early.

#2561 includes focused write/read interleaving tests and a counter-capacity regression. Thanks again—the reasoning and feedback here directly shaped the final implementation.

PR #2563, opened 2026-09-02
libliteeth: Add RollBall protocol support

storborg:
Adds BIOS support for configuring and monitoring RollBall-capable SFP transceiver modules. Commands are provided for basic interaction, and config flags can be set in a SoC for boot-time configuration. This is primarily useful for configuring SFP+ transceivers to 5GBASE-R host modes for use with lower-speed transceivers, e.g. on the Acorn Baseboard Mini, the Xilinx AC701, etc.

Initial support includes register definitions for Marvell Alaska 10GBASE-T PHYs and SFP+ transceivers behind PCA954x I2C multiplexers.

See https://github.com/enjoy-digital/liteeth/pull/217 for the context behind this feature.


storborg:
I've simplified the API and narrowed the SoC-level configuration scope to just this:

```python
self.add_config("SFP_1_I2C",         "sfp1_i2c")
self.add_config("SFP_1_HOST_MODE",   "2500BASEX")

# Optional, for boards like AC701 with a PCA954x mux
self.add_config("SFP_1_MUX_ADDR",    0x74)
self.add_config("SFP_1_MUX_CHANNEL", 4)
```

I think this will make it easy to add support for many more modules/vendors in the future without breaking backwards compatibility.

PR #2564, opened 2026-09-03
soc/cores/cpu/zynq7000/core.py: added MIO mode and pin configuration for add_can, add_uart, add_sdio, add_spi, add_i2c and add_ethernet

trabucayre:
This PR has two goals
- for all add_xxx not already MIO/EMIO capable: methods are modified to accept a str like "MIO XX .. YY" meaning the controler is used with PS7 MIO instead of EMIO
- when a controler is used with PS7 MIO, corresponding pins must be configured at tcl level to produces the good/expect ps7_gpl_init.c (used by the bootloader). This commit also adds the correct sequence to set direction, io_type, slew pin per pin.


enjoy-digital:

PR #2593, opened 2026-09-12
soc/integration: Add an eth_mtu argument to add_etherbone()

storborg:
A very simple tweak to allow passing MTU into the `soc.add_etherbone()` call.


enjoy-digital:
Thanks @storborg!

PR #2617, opened 2026-09-17
Update CVA6 to upstream and re-work socs

shenki:
This updates CVA6 to c4c412a9f8c3.

The full variant is the cv64a6_imafdc_sv39. A 32-bit variant is preserved, but only tested in sim. A larger rv64gc with b extensions was added, suitable for running Linux on a A7-100T FPGA.

Depends on https://github.com/shenki/pythondata-cpu-cva6/tree/update-cva6-c4c412a9


enjoy-digital:
Thanks @shenki, that's merged and pythondata-cpu-cva6 has also been updated with from the work on your branch.

PR #2627, opened 2026-09-21
cpu/gowin_ae350: describe system clock separately from the CPU clock

soburi:
Zephyr's change https://github.com/zephyrproject-rtos/zephyr/pull/118968 updated AE350 SoC to take SYS_CLOCK_HW_CYCLES_PER_SEC from `/clocks/clk_sys`.

This change is intended to follow the change on the Zephyr. The clock output destination can now be specified using 'config_cpu_system_clock_node_ref'.
If a node is specified in this property, it will be set there; otherwise, it will be set to the CPU clock location as before.

Note: Zephyr's change already introduce to development line and release with version 4.5 that planned on 26 Oct.


enjoy-digital:
Thanks @soburi!

